PR #21, opened 2022-07-14
Support running selected SAS code

scnjdl:


client/src/commands/run.ts
```diff
@@ -1,28 +1,30 @@
...
+function getCode(outputHtml: boolean, selected = false): string {
+  const editor = window.activeTextEditor;
+  const doc = editor?.document;
+
+  if (!editor || !doc) {
+    return "";
+  }
+
+  const code = selected
+    ? doc.getText(new Range(editor.selection.start, editor.selection.end))
```
scnwwu, line 19: doc.getText(editor.selection) should be enough

vwjh37: yes, I have updated.

PR #32, opened 2022-11-07
fix: prevent the profile dialog pops up every time after submitting S…

bitaoz:


client/src/commands/run.ts
```diff
@@ -26,54 +28,62 @@ function getCode(outputHtml: boolean, selected = false): string {
 }
 
 async function _run(selected = false) {
-  const outputHtml = !!workspace
-    .getConfiguration("SAS")
-    .get("session.outputHtml");
-  const code = getCode(outputHtml, selected);
+  if (loading) return;
+  loading = true;
```
bitaoz, line 32: the reason for adding loading variable:
at the begining of opening the vscode file, commands.executeCommand will be executed slowly, so the run button will be visible for a while even if be clicked and the run function would be called multiple times.

scnwwu: Thanks for explaining.
How about name it "running" rather than "loading"?

client/src/commands/run.ts
```diff
@@ -26,54 +28,62 @@ function getCode(outputHtml: boolean, selected = false): string {
...
 
-  await window.withProgress(
-    {
-      location: ProgressLocation.Notification,
-      title: "Connecting to SAS session...",
-    },
-    setup
-  );
+  try {
```
scnwwu, line 34: I'm not a fan of huge try blocks. How about adding another function handling _run.catch().finally()?

bitaoz: all right, I'll revise this.

PR #34, opened 2022-11-07
prevent profile dialog popping up every time after submitting SAS code

bitaoz:


client/src/commands/run.ts
```diff
@@ -66,14 +68,25 @@ async function _run(selected = false) {
   );
 }
 
+async function _run(selected = false) {
```
scnwwu, line 71: Thank you for making this update.
Looks like this function doesn't need to be async as it doesn't await anything.
Either remove async or await runCode seems fine.

scnwwu (approved):

PR #35, opened 2022-11-22
feat: login with saslogon

scnwwu:
- Remove password grant flow as it's not secure enough.
- Remove token file approach, replace with authorization code grant flow as it's the standard approach and easier for user.
- Refresh access token with refresh_token.
- Built-in client ID for Viya4 2022.11 and later.
- Replace restAF library with OpenAPI generated library.


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## TBD
```
scottdover, line 3: Do we have a changelog format we want to follow? This deviates slightly from https://keepachangelog.com/en/1.0.0/ . Perhaps this should be `[Unreleased]`? Additionally, should we choose to adopt https://keepachangelog.com/en/1.0.0/, we should keep a link at the top of our changelog to point contributors to a guide of how to modify the changelog

client/src/commands/run.ts
```diff
@@ -75,7 +86,10 @@ function _run(selected = false) {
 
   runCode(selected)
     .catch((err) => {
-      window.showErrorMessage(JSON.stringify(err));
+      console.dir(err);
```
scottdover, line 89: Was this an intentional change?

client/src/session/rest/api/common.ts
```diff
@@ -0,0 +1,150 @@
...
+// Licensed under SAS Code Extension Terms, available at Code_Extension_Agreement.pdf
+
+/**
+ * Compute
+ * The Compute API defines the replacement for the application servers that were used for SAS 9. The Compute API enables clients to perform the following tasks: <UL> <LI>Submit and execute SAS code. <LI>Obtain the SAS log or listing information from executed SAS code. <LI>View output that is generated by a SAS code submission. <LI>Access SAS data sets that are created by a SAS code submission. <LI>Access SAS filerefs that are created by a SAS code submission. <LI>Access SAS variables that are defined and used in a SAS code submission. <LI>Retrieve information about engines that are available in a session. </UL>
+ *
+ * The version of the OpenAPI document: 11
+ * Contact: devex@sas.com
+ *
+ * NOTE: This class is auto generated by OpenAPI
```
scottdover, line 11: If these are auto-generated (and thus, shouldn't be reviewed), then we may want to suppress them w/ a `.gitattributes` file

scottdover: Potential `.gitattributes` file...
```
client/src/session/rest/api/common.ts linguist-generated=true
client/src/session/rest/api/compute.ts linguist-generated=true
client/src/session/rest/api/configuration.ts linguist-generated=true
```
Are there others that should be ignored, @scnwwu ?

Issue #36, opened 2022-11-28
Job running for longer than 60 seconds results in an error.


JosephDHenry:
Currently code that takes longer than 60 seconds results in an error in vscode, but the SAS code will continue to run. 
The extension should allow for executions of longer than 60 seconds.

This is code that will show the error.

`
data _null_;
n=0;
 do while (n<62);
   put n=;
   x=sleep(1,1);
   n+1;
 end;
run;
`

Zhirong2022:
@JosephDHenry 
I tried to use the code below to run in the formal released VAC SAS extension. I did not catch any error. I am sure if I missed something other steps? 
data _null_; n=0; do while (n<62); put n=; x=sleep(1,1); n+1; end; run;

2TomLi:
The fix has been included in release 0.1.2.

PR #40, opened 2022-11-28
updates for multiple issues.

JosephDHenry:
I renamed the session source folder to compute. This makes more sense since the entire compute API is in this source folder, not just session related stuff.

I refactored alot of the code out into individual classes Server, Session, Job, etc. This makes the related code easier to find and work on.

I changed the way results appear when SAS code runs. The Results window would always take the focus from the source file and the name of the results would always be the Same "Results". Now the window has the sas name ie. sashtml3.htm (could be better in the future). and the source window retains focus.

The logs for a source file now stream as the source code is being executed. This is a really nice quality of life update.

Previously any job that took longer than 60 seconds resulted in an error. I have removed this restriction since jobs can take much longer than that.


scnwwu:
Thank you very much for the contribution!
It will take some time for the team to review and test it. Will comment back.

client/src/compute/rest/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  //Now get the results
+  const results = await job.results();
+
+  const res: RunResult = {
+    html5: "",
+    title: "",
+  };
+  //We can return more than just html, but for right now we are only returning
+  //the first HTML file that we get
+  for (const result of results) {
```
scnwwu, line 106: In current main branch, it will get the "last" rather than "first" result. For example user closed the default ods and started a new one.

scnwwu (commented):
Let me explain the reason for current session directory. The intent is to provide an abstraction for different kinds of connection methods. So if in future we have other methods, it may look like
session
|- rest
|- ssh
|- iom
So I'm not sure if we want to rename it to compute. I understand it's confused with compute session. Perhaps we can rename it to `service` or `connection`?

So I would suggest to move the compute files into rest directory. So it will look like
```service or connection
|- rest
|    |- compute
|    |    |- job.ts, etc.
|    |- folders // in future
|    |- cas // in future
|    |- api
|    |    |- openapi generated files
|    |- auth.ts
|    |- index.ts
|- ssh // in future
|    |- index.ts
|- index.ts
```

client/src/commands/commands.ts
```diff
@@ -0,0 +1,137 @@
+import {
```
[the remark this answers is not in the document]
JosephDHenry: Yes I will remove it.

client/src/compute/rest/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  const apiConfig = getApiConfig();
+  if (!config.serverId) {
+    apiConfig.accessToken = await getAccessToken(config);
+  }
+
+  if (computeSession && computeSession.sessionId) {
+    const state = await computeSession
+      .getState()
+      .catch(() => (computeSession = undefined));
+    if (state === ComputeState.Error) {
```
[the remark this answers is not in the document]
JosephDHenry: Do you have a test case that enters syntax check mode? I fell like it would be better to be able to detect that mode and only cancel when needed instead of every time.

In the mean time I will change it back to cancel every time and open an issue to look into only sending cancel when necessary.

client/src/compute/rest/index.ts
```diff
@@ -0,0 +1,135 @@
...
+      await computeSession.cancel();
+    }
+  }
+
+  if (computeSession) return;
+
+  //Set the local in the base options so it appears on all api calls
```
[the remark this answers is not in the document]
JosephDHenry: Thanks.

client/src/compute/rest/index.ts
JosephDHenry: I figured as much. I was working on a change where ALL the ODS results were returned and multiple WebView's were opened, but I thought that change was too much for this PR so I left it out. I will change it back to using the last one.

client/src/compute/rest/index.ts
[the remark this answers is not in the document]
scnwwu: run
```
proc asd;
run;
```
and it will enter syntax check mode

scnwwu:
Overall good to me. Will look for design feedback about the Results title and testing results. Will report back.

JosephDHenry:
Is this change request going to be approved? I have multiple changes ready to put in PRs for, but all depending on this change.

scottdover:
Hey @JosephDHenry. I've added this to my todo list and will take a look this morning :)

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,132 @@
...
+    //The link must have an href
+    if (!link.href) throw new Error();
+
+    /*
+    const url = new URL(`http://${link.href}`);
+    const parms = {};
+    url.searchParams.forEach((value: string, key: string) => {
+      parms[key] = value;
+    });
+    */
```
scottdover, line 101: Is this needed?

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    //Take the optional options given and merge in the link options
+    Object.assign(options, {
+      headers: headers,
+      method: link.method,
+      url: link.href,
+    });
+
+    //TODO: We should not have to remove the /compute from the link
+    return { url: link.href.slice("/compute".length), options: options };
```
scottdover, line 111: Wondering if `link.replace('/compute', '')` would be a better fit here?

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,132 @@
...
+    /*
+    const url = new URL(`http://${link.href}`);
+    const parms = {};
+    url.searchParams.forEach((value: string, key: string) => {
+      parms[key] = value;
+    });
+    */
+
+    //Take the optional options given and merge in the link options
+    Object.assign(options, {
```
scottdover, line 104: Here (and up above), we're re-assigning values to a property that was passed in. This is typically discouraged. I'm wondering if we can do something like...

```
const headers = options ? { ...options?.headers } : {};
```
up above, and create a new object here, like...

```
const processedConfig: AxiosRequestConfig = {
  ...config,
  headers: headers,
  method: link.method,
  url: link.href,
};
```

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /*
+  Make a request via a link
+  */
+  async requestLink<T>(
+    link: Link,
+    options?: AxiosRequestConfig
+  ): Promise<AxiosResponse<T>> {
+    if (options === undefined) {
```
scottdover, line 121: Same comment as above

client/src/connection/rest/context.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const resp = await this.requestLink(link, options);
+
+    if (!(resp.status === 201 || resp.status === 200)) {
+      throw new Error(<string>resp?.data);
+    }
+
+    //Create the session from the http resposne
+    const session = ComputeSession.fromResponse(resp);
+    return session;
```
scottdover, line 88:
```suggestion
    return ComputeSession.fromResponse(resp);
```

JosephDHenry:
So I did that with the results window so that all the tabs dont all look the same. I just took the name that SAS names the html files that it creates in SASWORK.

client/src/connection/rest/job.ts
```diff
@@ -0,0 +1,170 @@
...
+    const resp = await this.api.getJobState(parms);
+    if (resp.status === 200) {
+      //Set the new etag
+      this.etag = resp.etag;
+
+      //return the state
+      return resp.data;
+    } else if (resp.status === 304) {
+      //Not modified
+      return await this.getState(); //This is bad. We need to cache the last state value
```
scottdover, line 75: Is more work needed, here?

client/src/connection/rest/server.ts
```diff
@@ -0,0 +1,123 @@
+import { BaseCompute, Compute, getApiConfig, stateOptions } from "./common";
+import { ServersApi, Server, Link } from "./api/compute";
+import { ComputeSession } from "./session";
+import { AxiosResponse } from "axios";
+//import { ComputeSession } from "./session";
```
scottdover, line 5: Clean up?

.gitignore
```diff
@@ -3,3 +3,4 @@ dist
 node_modules
 .vscode-test
 *.vsix
+.vscode-test-web
```
scottdover, line 6: Can we collapse this into the above and have a `.vscode-test*` gitignore rule?

scottdover:
> So I did that with the results window so that all the tabs dont all look the same. I just took the name that SAS names the html files that it creates in SASWORK.

Yeah, I get that. I was just wondering if there's a way to provide better context to the user.

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,132 @@
...
+export class Compute {
+  etag: string;
+
+  async self<T>(): Promise<T> {
+    throw new Error("Not implemented");
+  }
+
+  //Shortcut function to get a link by its name
+  getLink(links: Array<Link>, rel: string): Link | undefined {
+    return links.find((x) => x.rel === rel);
```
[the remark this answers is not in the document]
JosephDHenry: Done

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,132 @@
...
+  */
+  getLinkOptions(link: Link, options?: AxiosRequestConfig): RequestArgs {
+    if (options === undefined) options = {};
+    const headers = { ...options?.headers };
+
+    if (link.method === "POST") {
+      if (link.type !== undefined)
+        headers["Content-Type"] = computeMediaType(link.type);
+      if (link.responseType !== undefined)
+        headers["Accept"] = computeMediaType(link.responseType);
```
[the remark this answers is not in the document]
JosephDHenry: Identical to what section?

client/src/connection/rest/common.ts
JosephDHenry: No, I removed it.

JosephDHenry: There was a problem with a new version of Axios where query parms on the URL were getting dropped. This was my workaround, but the problem was fixed in Axios, so if we upgrade to the latest version it should work.

client/src/connection/rest/common.ts
JosephDHenry: Yes, probably, but this whole thing is kind of a kludge anyways. Right now we have to add /compute to the API instances but remove /compute from the links. Seems like there should be a way for it all to work together without having to add/remove this root.

client/src/connection/rest/common.ts
JosephDHenry: So basically I was trying to find a way to combine the 2 header objects together.

The baseOptions have header values and the input could also have header values. This was the only way I could find to conditionally add the baseOptions header map to the incoming header map.

client/src/connection/rest/context.ts
JosephDHenry: Done.

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,132 @@
...
+    return links.find((x) => x.rel === rel);
+  }
+
+  /*
+  Get the options for a link
+  This sets the URL, accept/content-type headers as well as the Method and 
+  etag headers if needed
+  */
+  getLinkOptions(link: Link, options?: AxiosRequestConfig): RequestArgs {
+    if (options === undefined) options = {};
```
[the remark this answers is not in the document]
scottdover: Yeah, I tend to agree with you there. It does look inconsistent, and I don't love that. I'm wondering if @smorrisj has any thoughts, here?

client/src/connection/rest/server.ts
JosephDHenry: Done.

.gitignore
JosephDHenry: Yes

JosephDHenry:
> 

The change I am proposing does provide a better context to the user than what it was, But you are right that it could be better.

client/src/connection/rest/common.ts
scottdover: Can you give me an example of what what `link.href` is?

client/src/connection/rest/common.ts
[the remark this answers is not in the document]
JosephDHenry: It does look similar, but we dont want to add the etag on POST, only on PUT.

client/src/connection/rest/common.ts
JosephDHenry: that looks promising. Let me try that out.

client/src/connection/rest/common.ts
JosephDHenry: So this is what the link looks like for Server state:

`        {
            "href": "/compute/sessions/0001-ses0001/state",
            "method": "GET",
            "rel": "state",
            "type": "text/plain",
            "uri": "/compute/sessions/0001-ses0001/state"
        },`

JosephDHenry:
I am pretty sure I have address all the comments, and the PR branch has been rebased to the lastest upstream version.

scottdover:
I'll take a second pass at this by EOD

.gitignore
```diff
@@ -1,5 +1,6 @@
 out
 dist
 node_modules
-.vscode-test
+.vscode-test*
 *.vsix
+.vscode-test-web
```
scottdover, line 6: We no longer need this line

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,143 @@
...
+    */
+    const processedConfig: AxiosRequestConfig = {
+      ...options,
+      headers: {
+        ...apiConfig.baseOptions?.headers,
+        ...options?.headers,
+      },
+    };
+
+    const o = this.getLinkOptions(link, processedConfig);
```
scottdover, line 139: Can we call this something like `options`?

client/src/connection/rest/job.ts
scottdover: Are there scenarios where this could lean to endless recursion? (i.e. getState calls itself over and over)

client/src/connection/rest/job.ts
```diff
@@ -0,0 +1,175 @@
...
+  due to an error or other signal.
+  */
+  async isDone(state?: string): Promise<boolean> {
+    const doneStates = ["done", "canceled", "error", "warning", "completed"];
+
+    if (state === undefined) {
+      state = await this.getState();
+    }
+
+    return doneStates.indexOf(state) === -1;
```
scottdover, line 161: Instead of reassigning a param, could we...
```js
return doneStates.indexOf(state || await this.getState()) === -1;
```

client/src/connection/rest/server.ts
```diff
@@ -0,0 +1,133 @@
...
+    const params = {};
+    const headers = {};
+
+    if (options !== undefined) {
+      if (options.onChange) {
+        headers["If-None-Match"] = this.etag;
+      }
+
+      if (options.wait) {
+        // eslint-disable-next-line @typescript-eslint/dot-notation
```
scottdover, line 112: I think this is probably fine to be `params.timeout`

client/src/connection/rest/common.ts
```diff
@@ -0,0 +1,143 @@
+import * as typer from "media-typer";
```
scottdover, line 1: Can we import what we need from `media-typer` instead of all of the things?

scottdover (commented):
I think this is the last of my feedback.

.gitignore
JosephDHenry: done

client/src/connection/rest/job.ts
JosephDHenry: > Instead of reassigning a param, could we...
> 
> ```js
> return doneStates.indexOf(state || await this.getState()) === -1;
> ```

yeah, I could do that. Still getting used to ts/js shortcuts.

The getState function can never lead to recursion. The only way a 304 is returned is if the if-none-match header is sent. calling getState with no options, will send a simple call to getState which simply returns the current state of the job.

client/src/connection/rest/server.ts
JosephDHenry: From the previous comment I made about dot notation, it looks odd to use dot notation in this if statement, but type notation in the one above it.

client/src/connection/rest/common.ts
JosephDHenry: Yes, I did that because the function names in media-typer we very generic and can easily collide with others. I will import the functions as a clear name though.

client/src/connection/rest/server.ts
scottdover: I'd advocate for following the eslint rule here instead of adding exceptions. It's somewhat inconsistent in look here, but it's consistent with the eslint rule and similar to your change in common.ts

client/src/connection/rest/job.ts
scottdover: Just copy/pasting from a different comment for visibility...
> The getState function can never lead to recursion. The only way a 304 is returned is if the if-none-match header is sent. calling getState with no options, will send a simple call to getState which simply returns the current state of the job.

mail2fei:
> > The Results window would always take the focus from the source file and the name of the results would always be the Same "Results". Now the window has the sas name ie. sashtml3.htm (could be better in the future)
> 
> I'm curious about this change. Having a distinct "Results" window related to the source code seems reasonable. That said, does `sashtml3.htm` mean anything to the consumer? It would be interesting if we could do something like `Results for <localFileName>`
> 
> > Overall good to me. Will look for design feedback about the Results title and testing results. Will report back.
> 
> Hey @scnwwu . Did you have a chance to gather feedback for this?

mail2fei:
Sorry, I touched the wrong button to close this.  I am asking UI to give comment on this PR.

mail2fei:
a17c4d2f-96da-4c02-88f7-8826b31f9e44
Adding the UI difference screenshot for UI to review.

skyeaussie:
@JosephDHenry   The SAS UX team reviewed the change to the Results tab. We agree with @scottdover  that using something like `sashtml.htm` wouldn't mean a lot to users. How do you feel about something like this? Using 'Results' in the tab name would be a familiar experience to SAS users.

Results - <name of sas file.sas>

JosephDHenry:
> 



> @JosephDHenry The SAS UX team reviewed the change to the Results tab. We agree with @scottdover that using something like `sashtml.htm` wouldn't mean a lot to users. How do you feel about something like this? Using 'Results' in the tab name would be a familiar experience to SAS users.
> 
> `Results - <name of sas file.sas>`

So 2 things, 

1.) There is not always a SAS file involved in the execution. You could have a file that is not saved OR running selected code. so Results - <file.sas> does not work.
2.) Results is kind of ambiguous in this situation. We are displaying the HTML results, but not any of the other results like PNG files, dataset, libraries, etc. Since we are using a webview, the name of the actual html file that is generated by sas seems to make sense.

skyeaussie:
> > 
> 
> > @JosephDHenry The SAS UX team reviewed the change to the Results tab. We agree with @scottdover that using something like `sashtml.htm` wouldn't mean a lot to users. How do you feel about something like this? Using 'Results' in the tab name would be a familiar experience to SAS users.
> > `Results - <name of sas file.sas>`
> 
> So 2 things,
> 
> 1.) There is not always a SAS file involved in the execution. You could have a file that is not saved OR running selected code. so Results - <file.sas> does not work. 2.) Results is kind of ambiguous in this situation. We are displaying the HTML results, but not any of the other results like PNG files, dataset, libraries, etc. Since we are using a webview, the name of the actual html file that is generated by sas seems to make sense.


How about this?
 
SAS Results 1.htm
SAS Results 2.htm
 
the .htm implies it's an HTML file so no need to add html  to the name
 
But there is still no associate to the originating program. A user could have 5 sas programs up and running.... how would they know which results belonged to which program?

scottdover:
@JosephDHenry @skyeaussie It seems there are still some things to figure out w/r/t tab display. Can we remove tab-related changes from this PR and open a new issue to discuss?

JosephDHenry:
> > > 
> > 
> > 
> > > @JosephDHenry The SAS UX team reviewed the change to the Results tab. We agree with @scottdover that using something like `sashtml.htm` wouldn't mean a lot to users. How do you feel about something like this? Using 'Results' in the tab name would be a familiar experience to SAS users.
> > > `Results - <name of sas file.sas>`
> > 
> > 
> > So 2 things,
> > 1.) There is not always a SAS file involved in the execution. You could have a file that is not saved OR running selected code. so Results - <file.sas> does not work. 2.) Results is kind of ambiguous in this situation. We are displaying the HTML results, but not any of the other results like PNG files, dataset, libraries, etc. Since we are using a webview, the name of the actual html file that is generated by sas seems to make sense.
> 
> How about this?   SAS Results 1.htm SAS Results 2.htm   the .htm implies it's an HTML file so no need to add html  to the name   But there is still no association to the originating program. A user could have 5 sas programs up and running.... how would they know which results belonged to which program?

There is no direct association between a program and results. There is only an association between a Job and the Results a job produces, and a Job can be the entire contents of a file, a selection in that file, or a file reference that has not yet been saved. 
The Jobs API returns the results in a certain way, and the "name" of the html results is what I was displaying.

As far as when you say " A user could have 5 sas programs up and running.... how would they know which results belonged to which program?" That is a bigger issue that would need a bit more discussion. 

For right now I am simply going to revert the title changes back to what it was. The main thing I was trying to fix was the results tab becoming active after running SAS code.

JosephDHenry:
> @JosephDHenry @skyeaussie It seems there are still some things to figure out w/r/t tab display. Can we remove tab-related changes from this PR and open a new issue to discuss? That way, we can figure out our approach before making too many code changes.

I change the title back to "Result" but kept the change for not switching focus.

liu-lihua:
@scnwwu and @scottdover , is this pull request ready for test? If so, @Zhirong2022 will work on test and provide test result. Thank you all very much.

scnwwu:
@liu-lihua, @Zhirong2022, yes, please test it. Thanks.

Zhirong2022:
@JosephDHenry , I got an issue when I ran the code below.
   proc print data=maps.africa;
When I ran the code successfully for the first time. I closed the Result dialog, Log dialog and just kept the code there for about 2 hours. I tried to rerun the code for the second time, I got an error. Please check the screen shot.  However, if I continued to run the code, it worked well.  Could you have a try? 
image

JosephDHenry:
> 

I had already found that problem and fixed it in a future commit that is waiting on this PR.

I went ahead and merged the fix back.

Zhirong2022:
@JosephDHenry Thank you so much for your help and confirmation. I have no concerns now. 
The issues raised in https://github.com/sassoftware/vscode-sas-extension/issues/36 and https://github.com/sassoftware/vscode-sas-extension/issues/63 was fixed in this pull request.  And no other regression issues found during the testing.  
Mark it as test-complete.

Zhirong2022 (approved):

package.json
```diff
@@ -331,11 +331,12 @@
...
     "mocha": "^10.0.0",
-    "prettier": "^2.7.1",
     "path-browserify": "^1.0.1",
+    "prettier": "^2.7.1",
     "ts-loader": "^9.2.3",
     "typescript": "^4.4.3",
     "webpack": "^5.44.0",
-    "webpack-cli": "^4.7.2"
+    "webpack-cli": "^4.7.2",
+    "media-typer": "^1.1.0"
```
scnwwu, line 340: it looks like being used in source code so better not be in devDependencies. Suggest to be in the dependencies entry in client/package.json. It won't cause real issues as it will be bundled and not publish npm modules. But just make it clearer. client and server can be separated when needed in future, so their dependencies are in their own package.json file.

PR #41, opened 2022-11-29
Add .gitattributes/adjust changelog/remove console

scottdover:
wip

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+client/src/session/rest/api/ linguist-generated=true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
-## v0.1.1 (24 Nov 2022)
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [v0.1.1] - 24 Nov 2022
 
 ### Breaking Change
 
@@ -12,7 +14,7 @@
 - Refresh your access token with refresh token
 - Built-in client ID for Viya4 2022.11 and later
 
-## v0.1.0 (8 Oct 2022)
+## [v0.1.0] - 8 Oct 2022
 
 ### Breaking Change
 
@@ -29,13 +31,13 @@
 - Percentage sign should escape quotes in %str
 - PROC SQL snippet syntax
 
-## v0.0.7 (26 Jul 2022)
+## [v0.0.7] - 26 Jul 2022
 
 ### Feature
 
 - Support run selected SAS code
 
-## v0.0.6 (8 Jul 2022)
+## [v0.0.6] - 8 Jul 2022
 
 ### Fix
 
@@ -51,7 +53,7 @@
 
 - Update README to show animated gifs
 
-## v0.0.5 (26 May 2022)
+## [v0.0.5] - 26 May 2022
 
 ### Fix
 
@@ -62,7 +64,7 @@
 - Rearrange settings fields
 - Update dependencies versions
 
-## v0.0.4 (19 May 2022)
+## [v0.0.4] - 19 May 2022
 
 ### Fix
 
@@ -72,6 +74,6 @@
 
 - Update README.md to link to wiki
 
-## v0.0.3 (17 May 2022)
+## [v0.0.3] - 17 May 2022
 
 Initial release
--- a/client/src/commands/run.ts
+++ b/client/src/commands/run.ts
@@ -86,7 +86,6 @@ function _run(selected = false) {
 
   runCode(selected)
     .catch((err) => {
-      console.dir(err);
       window.showErrorMessage(
         err.response?.data ? JSON.stringify(err.response.data) : err.message
       );
--- a/client/src/session/rest/api/compute.ts
+++ b/client/src/session/rest/api/compute.ts
@@ -14,7 +14,7 @@
 
 /* tslint:disable */
 /* eslint-disable */
-
+// I've added a change
 import { Configuration } from "./configuration";
 import { AxiosResponse, AxiosRequestConfig } from "axios";
 // URLSearchParams not necessarily used


client/src/commands/run.ts
```diff
@@ -86,7 +86,6 @@ function _run(selected = false) {
 
   runCode(selected)
     .catch((err) => {
-      console.dir(err);
```
scottdover, line 89: Looking through the code, it looks like this was probably intentional. Can you confirm, @scnwwu?

smorrisj:
A few questions:

1. Should we prefer use the conventional commit syntax for messages? I see a mix of both in the history.
2. Can you elaborate on the format change in the CHANGELOG.md? Just wondering if that was an intentional change or if it was generated via a tool.

CHANGELOG.md
```diff
@@ -12,7 +14,7 @@
 - Refresh your access token with refresh token
 - Built-in client ID for Viya4 2022.11 and later
 
-## v0.1.0 (8 Oct 2022)
+## [v0.1.0] - 8 Oct 2022
```
smorrisj, line 17: Was the formatting change intentional?

scottdover: Only better to match https://keepachangelog.com/en/1.0.0/ ...if that's something we want to do? 🤷

scottdover:
> Should we prefer use the conventional commit syntax for messages? I see a mix of both in the history.

I think we should be aiming for conventional commits, personally. And ideally, I think we should also be using "Squash & Merge" when we close out changes. That way, our history isn't littered with tiny changes that don't provide as much insight as to why a file was changed.

smorrisj:
> > Should we prefer use the conventional commit syntax for messages? I see a mix of both in the history.
> 
> I think we should be aiming for conventional commits, personally. And ideally, I think we should also be using "Squash & Merge" when we close out changes. That way, our history isn't littered with tiny changes that don't provide as much insight as to why a file was changed.

agreed

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
-## v0.1.1 (24 Nov 2022)
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
```
scnwwu, line 3: I'm good with following a guideline.
I guess current format was from the vscode python extension
https://github.com/microsoft/vscode-python/blob/main/CHANGELOG.md
Though we're not strictly follow it and there's no need to follow it.
I'm good with the "Keep a Changlog". But looks like this pr doesn't modify existing date format and types of changes. Do we want to change them all? Do we want to link the format before we fully follow it?

client/src/commands/run.ts
scnwwu: Yes, it would be helpful to see the entire error object for trouble shooting.

scnwwu:
> > > Should we prefer use the conventional commit syntax for messages? I see a mix of both in the history.
> > 
> > 
> > I think we should be aiming for conventional commits, personally. And ideally, I think we should also be using "Squash & Merge" when we close out changes. That way, our history isn't littered with tiny changes that don't provide as much insight as to why a file was changed.
> 
> agreed

Good to me.
BTW, current title of this pr has "feat:". To me it's more like "chore"?

CHANGELOG.md
smorrisj: Should we create an issue for this item to form discussion around?

.gitattributes
```diff
@@ -0,0 +1,3 @@
+client/src/session/rest/api/common.ts linguist-generated=true
+client/src/session/rest/api/compute.ts linguist-generated=true
```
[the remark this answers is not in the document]
smorrisj: For this directory is the intent that all of the files here would be generated typescript source? I'm just wondering about explicitly enumerating each file in the git attributes versus just denoting it for the entire directory.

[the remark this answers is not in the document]
scottdover: > For this directory is the intent that all of the files here would be generated typescript source? 

I'll defer to @scnwwu for the answer to this. The reason I chose to explicitly list each file is I didn't want to end up in a situation where a user put something in there that _isn't_ auto-generated, but that we were less apt to look at (since it's collapsed by default).

Also, this keeps the files from expanding in a pull request. However, they _still_ show up. This makes these files behave more like `yarn.lock` (and other well known generated files). So, while they aren't expanded by default, a reviewer can still expand to see _what_ was changed.

CHANGELOG.md
scottdover: I missed the dates. Perhaps we can:
 - Change the headers (versions & dates)
 - Get rid of the link to keep a changelog
 - create an issue to discuss the standard and whether we want to apply that to the version notes going forward, or retroactively apply it to things that've already been released? 

I also noticed I've not updated the Changelog as part of this PR 🤦

.gitattributes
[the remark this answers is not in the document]
scnwwu: I would expect this directory to contain only OpenAPI generated files.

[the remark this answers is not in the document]
scnwwu: BTW., `server/data/**` and `server/pubsdata/**` are also generated files.

CHANGELOG.md
```diff
@@ -1,6 +1,16 @@
 # Changelog
 
-## v0.1.1 (24 Nov 2022)
+## [Unreleased]
+
+### Added
```
scnwwu, line 5: To me, the changelog is more for customers. Especially it will also be displayed in the marketplace and the extension page in VS Code.
Reading the "Keep a Changelog", it says
- `Added` for new features.
- `Changed` for changes in existing functionality.

So my feeling is that this change doesn't need to be documented here

CHANGELOG.md
scnwwu: sounds good to me

CHANGELOG.md
scottdover: Makes sense to me :)

scnwwu (approved):
Thank you!

PR #42, opened 2022-12-01
chore: update README

scnwwu:
A user asked us to document the build-in client ID from Viya 2022.11 so that they can register the same name in older cadence then no change needed later.


jefurbee (approved):
Looks good to me.

Issue #43, opened 2022-12-01
Unable to submit code when there is a period in the profile name


clangsmith:
If you create a profile with a period in the name, you will not be able to submit code using that profile. Rather, when you attempt to run the code, you will get an error toast with the message:
"Cannot read properties of undefined (reading 'endpoint')"

You will still be able to submit code using other profiles that do not contain a period in the name. Also, as soon as you remove the period from the profile name, you will be able to submit code using that profile again.

clangsmith:
I confirmed this issue has been fixed. The fix is included in release 0.1.2.

PR #44, opened 2022-12-02
fix: copy vscode setting object (#43)

scnwwu:
It appears that altering value in the object got from vscode settings will result in error in this case, has to copy.


boyce-w (approved):
LGTM

PR #47, opened 2022-12-06
Enforce Dot Notation

smorrisj:
What should our stance on enforcing dot notation be? I found some spots where javascript key notation is used for getting and setting values. I'd like to propose that we take a stance of using idiomatic typescript whenever possible, including using the dot syntax. 

It gets a bit trickier for any types. I'd propose that we leave the enforcement of the rule at error and manually suppress these on a case by case basis. I would think that for this issue, having a lot of manual suppressions would indicate some kind of design smell in the types that we'd want to address.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+.eslintrc.js
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,6 @@ module.exports = {
   rules: {
     eqeqeq: "error",
     "prefer-const": "error",
-    "dot-notation": "error"
+    "dot-notation": "error",
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,5 +13,6 @@ module.exports = {
   rules: {
     eqeqeq: "error",
     "prefer-const": "error",
+    "dot-notation": "error"
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,11 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 6,
     sourceType: "module",
-    project: ["./tsconfig.json", "./client/tsconfig.json", "./server/tsconfig.json"]
+    project: [
+      "./tsconfig.json",
+      "./client/tsconfig.json",
+      "./server/tsconfig.json",
+    ],
   },
   env: {
     node: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,6 +4,7 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 6,
     sourceType: "module",
+    project: ["./tsconfig.json", "./client/tsconfig.json", "./server/tsconfig.json"]
   },
   env: {
     node: true,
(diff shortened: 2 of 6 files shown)


scottdover:
I was just writing up an issue to discuss consistency :). This is one that I was going to mention since we chatted about it. 

> I would think that for this issue, having a lot of manual suppressions would indicate some kind of design smell in the types that we'd want to address.

I agree. And, using the `any` type _should_ be discouraged unless absolutely necessary

smorrisj:
> I was just writing up an issue to discuss consistency :). This is one that I was going to mention since we chatted about it.
> 
> >

Great Idea. There are other things on that issue we'd want to address as well. Once its in, can link it here.

scottdover:
Here it is :) - https://github.com/sassoftware/vscode-sas-extension/issues/48

scottdover (approved):

.eslintrc.js
```diff
@@ -4,6 +4,11 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 6,
     sourceType: "module",
+    project: [
```
scottdover, line 7: There are more details here: https://eslint.org/docs/latest/user-guide/configuring/configuration-files .  Had to look this up because I've never seen it before. I _think_ the last two array entries may be redundant, since the root `.tsconfig.json` references the client/server specific tsconfig

server/src/sas/SyntaxDataProvider.ts
```diff
@@ -2,7 +2,7 @@
 // Licensed under SAS Code Extension Terms, available at Code_Extension_Agreement.pdf
 
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types,
-@typescript-eslint/no-unused-vars,@typescript-eslint/no-explicit-any */
+@typescript-eslint/no-unused-vars,@typescript-eslint/no-explicit-any,@typescript-eslint/dot-notation */
```
scottdover, line 5: @scnwwu Do you know if this file was copy/pasted from somewhere else? In other words, should it be treated as "generated" code? If not, I'm wondering if there's a plan to cleanup the rules we disable here?

.eslintrc.js
smorrisj: This was required to to adding a rule from the typescript-estlint plugin. Without adding the project setting, the following rule would be seen:

You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.

It the client and server tsconfigs are removed, there will be numerous errors to this effect:

```
0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: server/src/sas/SyntaxProvider.ts.
The file must be included in at least one of the projects provided
```

scottdover:
I'd like to get @scnwwu's feedback on this one as well, but it looks good to me

.eslintignore
```diff
@@ -0,0 +1 @@
+.eslintrc.js
```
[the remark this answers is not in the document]
scnwwu: I don't see this error on my machine. Do you have specific setting in your extension? I can't think of why this should be an error:) But it's not a big deal. Others good to me

scnwwu (approved):

Issue #48, opened 2022-12-06
Fix inconsistent code formatting


scottdover:
I've noticed some inconsistencies (or lack of guidelines) in the extension code that I'd like to discuss. These inconsistencies can make it a challenge for contributors to know which pattern to follow, which can result in more churn in code review. Here are a few examples:

### Code inconsistencies

**Inconsistent brackets**. See https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/profile.ts#L99-L101 and https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/session/rest/index.ts#L92-L93. We could make use of https://eslint.org/docs/latest/rules/curly to enforce this

**Inconsistent object access**. @smorrisj has a draft PR in place to address this inconsistency: https://github.com/sassoftware/vscode-sas-extension/pull/47

I'm hoping to use this issue to come up with the ideal lint configuration for this project. A draft PR has been created to address any remaining eslint configuration changes: https://github.com/sassoftware/vscode-sas-extension/pull/49

### No guideline for changelog

In a previous commit, our changelog was updated to better reflect the guidelines at https://keepachangelog.com/en/1.0.0/ . Should we follow these guidelines moving forward? And, if so, should we update previously released versions of the code to better match the guidelines?

scottdover:
I'm also wondering about how we create and merge PRs.

Should we have a template to provide more consistency around _what_ we're doing, and how we've tested it? If so, we can use the following: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

Also, I'm wondering if there's benefit in copy/pasting the description from a PR into the description of the pull request when squash merging? I believe the default is to list each commit from the branch, but it may make more sense to have a summary of what has changed and why (not a great way to enforce this, though)

scottdover:
Reopening this issue, and now realizing closing a PR attached to the issue automatically closes the issue

PR #49, opened 2022-12-06
Update eslint to address inconsistencies

scottdover:
This updates our eslint config to match what comes out of the discussion in https://github.com/sassoftware/vscode-sas-extension/issues/48

**This is currently a WIP**


.eslintrc.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
   rules: {
     eqeqeq: "error",
     "prefer-const": "error",
+    curly: "error",
```
scnwwu, line 16: I would vote for ["error", "multi-line", "consistent"]. Some simple short lines look straightforward to me.

scottdover: I tend to lean towards explicit brackets for conditionals, for a couple reasons:

It provides consistency everywhere, where you don't have to _think_ about how to write an if statement (an if statement will _always_ have opening and closing brackets)

It provides a somewhat cleaner git history. For example, if I write code like...
```js
if (someCheck) return value;
```
and you update the code to...
```js
if (someCheck) {
  const calculatedValue = value + 1;
  return calculatedValue;
}
```
then it _appears_ that _you_ were responsible for defining `someCheck` (because you edited that line of code). So if a future dev has a question about `someCheck`, they might be more likely to reach out to you, although I was the one responsible for the original code. It's a super nitpicky edge case, though.

Also, I just find something like this kind of gross (but valid with w/ `["error", "multi-line", "consistent"]`):
```js
let sum = 0;
while (item.next()) sum += item.value;
```

I'd love to get other people's feedback on this one. I could definitely be over thinking it 😁

scnwwu: Makes sense. I agree with you now:) Thank you

scnwwu (approved):

.eslintrc.js
smorrisj: Agreed with @scnwwu. Looks good here

smorrisj (approved):

PR #51, opened 2022-12-07
Change Run and Run Selected behavior (including default keybindings) to be consistent with other SAS IDEs

clangsmith:
Matching the defaults of SAS Studio and EG, and very similar to DMS.

Planning to submit additional commits to change the run behavior to be more consistent with SAS Studio, EG, and DMS by default. For example, Run Selected should run all the code if there is no selection. Also, the "Run" button on the toolbar should perform the "Run selected" command...  if there is a selection, run just the selection, if there is no selection run all the code.

### DIFF
--- a/client/src/commands/run.ts
+++ b/client/src/commands/run.ts
@@ -19,10 +19,17 @@ let running = false;
 function getCode(outputHtml: boolean, selected = false): string {
   const editor = window.activeTextEditor;
   const doc = editor?.document;
-  let code: string;
+  let code = "";
   if (selected) {
-    // run selected code if there is a selection, otherwise run all code
-    code = doc?.getText(editor?.selection);
+    // run selected code if there is one or more non-empty selections, otherwise run all code
+
+    // since you can have multiple selections, append the text for each selection in order of selection
+    // note: selection ranges can be empty (ex. just a carat)
+    for (const selection of editor.selections) {
+      const selectedText: string = doc.getText(selection);
+      code += selectedText;
+    }
+    // if no non-whitespace characters are selected, treat as no selection and run all code
     if (code.trim().length === 0) {
       code = doc?.getText();
     }
@@ -30,11 +37,7 @@ function getCode(outputHtml: boolean, selected = false): string {
     code = doc?.getText();
   }
 
-  return code
-    ? outputHtml
-      ? "ods html5;\n" + code + "\n;quit;ods html5 close;"
-      : code
-    : "";
+  return outputHtml ? "ods html5;\n" + code + "\n;quit;ods html5 close;" : code;
 }
 
 async function runCode(selected?: boolean) {
--- a/client/src/commands/run.ts
+++ b/client/src/commands/run.ts
@@ -19,7 +19,16 @@ let running = false;
 function getCode(outputHtml: boolean, selected = false): string {
   const editor = window.activeTextEditor;
   const doc = editor?.document;
-  const code = selected ? doc?.getText(editor?.selection) : doc?.getText();
+  let code: string;
+  if (selected) {
+    // run selected code if there is a selection, otherwise run all code
+    code = doc?.getText(editor?.selection);
+    if (code.trim().length === 0) {
+      code = doc?.getText();
+    }
+  } else {
+    code = doc?.getText();
+  }
 
   return code
     ? outputHtml
--- a/package.json
+++ b/package.json
@@ -216,23 +216,35 @@
         "category": "SAS"
       }
     ],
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -216,22 +216,34 @@
...
-          "command": "SAS.run"
+          "command": "SAS.runSelected"
         }
       ],
       "editor/context": [
         {
-          "when": "editorFocus && editorLangId == sas && !SAS.hideRunMenuItem",
-          "command": "SAS.run",
+          "when": "editorLangId == sas && !SAS.hideRunMenuItem",
+          "command": "SAS.runSelected",
```
clangsmith, line 241: I assumed the order here would define the order these commands appear in the context menu, however Run is still appearing before Run Selected when I run in the debugger. Anyone know how to change this order?

scnjdl: The context menu entries are ordered alphabetically. So 'Run SAS Code' is always in front of 'Run Selected SAS Code'

scnwwu: https://code.visualstudio.com/api/references/contribution-points#Sorting-inside-groups

clangsmith: Great, thanks @scnwwu! I confirmed adding the order attribute to the group did the trick. I'll submit another commit.

scnwwu: Another thought is that perhaps we don't need the run entry now, as runSelected(OrAll) should work in most cases. Context menu is just a short cut for frequent actions. Keeping context menu as shorter as possible might be a good idea. Command Palette is the common place for commands.

scottdover (approved):

crlewi:
@clangsmith reviewed the UI changes with me and they look good from a UX perspective.

scnwwu:
@clangsmith in case you're looking for my review, would you mind reply my open comments? Thanks.

client/src/commands/run.ts
```diff
@@ -19,7 +19,14 @@ let running = false;
 function getCode(outputHtml: boolean, selected = false): string {
   const editor = window.activeTextEditor;
   const doc = editor?.document;
-  const code = selected ? doc?.getText(editor?.selection) : doc?.getText();
+  let code: string;
+  if (selected) {
+    // run selected code if there is a selection, otherwise run all code
+    code = doc?.getText(editor?.selection);
```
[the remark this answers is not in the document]
clangsmith: Yes, I think we should combine all selected code and run. In fact, I could see that being a convenient feature, where you wanted to select a few parts of a program and just run those. I coded that up and will push shortly for review.

package.json
clangsmith: Good point. I'm going to leave the Run command on the context menu for now though, since @crlewi and @skyeaussie reviewed the UI/UX and were good with it. It potentially helps with discovery as well. Last but not least, there are times I have a selection, but my intent is really to run the whole program, so I could see some users preferring to frequently run all the code, ignoring any potential selection.

scnwwu:
Thank you for the reply. Look good to me. I'll let testing finally approve this. Thanks.

scottdover:
@Becky-Williamson Did you intentionally close this PR? Do we not want to merge these changes?

Becky-Williamson:
@scottdover, no, I didn't mean to close it.  I don't even know how I did it.  How do I fix it?  Sorry!

scottdover:
> @scottdover, no, I didn't mean to close it. I don't even know how I did it. How do I fix it? Sorry!

No worries. I went ahead and re-opened it. As soon as testing has confirmed things look good, @clangsmith should be able to merge this one

Becky-Williamson:
Testing complete.  No new issues.

Run Button:
- Verified that all code is executed if nothing is selected.
- Verified that only selected code is executed if something is selected.

Context meu:
- Verified that all code is executed if "Run Selected or All SAS Code" if no code is selected.
- Verified that only selected code is executed if "Run Selected or All SAS Code" with selected code.
- Verified that all code is executed, regardless of code selection, if "Run All SAS Code" is selected.

Keyboard Shortcuts:
- Verified that all code is executed if nothing is selected and F3 is pressed.
- Verified that only selected code is executed and F3 is pressed.
- Verified that all code is executed, regardless of code selection, if F8 is pressed.

clangsmith:
I also want to point out that VS Code supports multiple selections. For example, you can Alt+mouse click-and-drag (by default) to add an additional selection. With these changes, the SAS extension now handles multiple selections (when "Run Selected or All SAS Code" command is executed) by combining the text from the selections **in the order in which they were selected**, and then submits the combined code. This is a convenient way to run multiple parts of a program, in whatever order you desire, in a single submission.

PR #52, opened 2022-12-07
Make prettier the default formatter

smorrisj:
Since we already recommend prettier as the formatter of choice, and already source a config js for it, set it as the default formatter. #48

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.tabSize": 2,
   "editor.insertSpaces": true,
   "editor.formatOnSave": true,


clangsmith (approved):
Looks good to me. @smorrisj and I debated whether it should be defaultFormatter for everything or target specific types (ex. .json and .ts -- see user settings json snippet below). Decided to make it default for all for now, since it looks like it checks the type anyway before executing.

    "[json]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
    "[typescript]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },

mail2fei:
Maybe we should limit the file types, I just added 2 new md file for new bug and request template, it is pretty picky on the md files. Like I can't use "...." which is copied from github's template, must change to "..."

scnwwu:
> Maybe we should limit the file types, I just added 2 new md file for new bug and request template, it is pretty picky on the md files. Like I can't use "...." which is copied from github's template, must change to "..."

We ever had issues when displaying the README in the marketplace while look good on github. Different markdown engine may have difference. Prettier's format appears to have good compatibility to me till now. So we really need it at least for the files that will be displayed in other places (README, CHANGELOG), as we don't have other way to verify on marketplace before publish.
If we want to keep some github templates unchanged and as they won't display outside github, we can add them to the `.prettierignore` file as needed.

PR #57, opened 2022-12-12
feat: add SAS content pane

scottdover:
**Summary**
This introduces the first iteration of SAS content navigator. In this first iteration, the following functionality is introduced:
 - File and folder creation
 - File, folder, and favorites renaming
 - File and folder deletion

One thing to call out here is that File deletion is permanent in this iteration. In a future iteration, we'll attempt to bring over the recycle bin functionality currently present in SAS drive. 

**Testing**
- [x] Test creating a file*
- [x] Test creating a folder*
- [x] Test renaming a file*
- [x] Test deleting a file*
- [x] Test renaming a file in My Favorites* (renaming a file or folder renames the link, but not the actual underlying resource. However, renaming a file w/n a link folder does update the file in both places)
- [x] Test signing in (both with and without profile)
- [x] Test opening a variety of files (sas programs but also json files / etc)
- [x] Test refresh functionality
- [x] Test collapse all functionality
- [x] Check with @smorrisj about how this functionality is impacted by v9 (where SAS content is unsupported)
- [x] **Make sure messages are finalized with @skyeaussie and @crlewi before merging**

* For each test case, test in root folder and sub folder. Test that file actions with naming collisions (i.e. creating a file with a name that already exists) fails


package.json
```diff
@@ -298,6 +298,39 @@
...
+          "id": "sas-libraries-pane",
+          "name": "Libraries"
+        }
+      ]
+    },
+    "viewsWelcome": [
+      {
+        "view": "sas-content-pane",
+        "contents": "Please [login](command:SAS.authenticate) to view SAS content.",
+        "when": "!SAS.session.outputHtml"
```
scottdover, line 327: @scnwwu Do you know if there is a way to detect if a user has a current valid session? This doesn't quite work. The general idea here is that we want to encourage users to login to see SAS Content if they aren't logged in.

scnwwu: I think the extension should maintain a login status. (Implement a https://code.visualstudio.com/api/references/vscode-api#AuthenticationProvider). That will be shared across various functionalities.

client/src/components/ContentNavigator/viya/const.ts
```diff
@@ -0,0 +1,59 @@
...
+  NewFileTitle: "New file",
+  NewFolderCreationError: 'Unable to create folder "%(name)s"',
+  NewFolderPrompt: "Please enter a folder name",
+  NewFolderTitle: "New folder",
+  RenameError: 'Unable to rename "%(oldName)s" to "%(newName)s"',
+  RenameFileTitle: "Rename file",
+  RenameFolderTitle: "Rename folder",
+  RenamePrompt: "Please enter a new name",
+  RenameUnsavedFileError: "Please save your file first",
+};
```
scottdover, line 59: Hey @crlewi @skyeaussie. Here are a list of messages used as part of these changes. I tried to make the keys pretty self-explanatory, but let me know if you have any questions or need me to make any adjustments.

client/src/node/extension.ts
```diff
@@ -101,15 +106,24 @@ export function activate(context: ExtensionContext): void {
...
+}
+
+function triggerProfileUpdate(): void {
+  const profileList = profileConfig.getAllProfiles();
+  const activeProfileName = profileConfig.getActiveProfile();
+  if (activeProfileName in profileList || activeProfileName === "") {
+    updateStatusBarProfile(activeProfileStatusBarIcon);
+    // TODO Update with changes for SAS 9 integration
+    // commands.executeCommand("setContext", "SAS.connectionMethod", profileList[activeProfileName].connectionMethod || "rest");
+    commands.executeCommand("setContext", "SAS.connectionMethod", "rest");
```
scottdover, line 123: @smorrisj I put something in place here that we can hopefully tweak w/ SAS 9 integration.

package.json
```diff
@@ -310,6 +369,47 @@
...
+        "view": "sas-content-get-started",
+        "when": "!SAS.authorized && SAS.connectionMethod == rest",
+        "contents": "You have not yet signed in to SAS.\n[Sign in](command:SAS.authorize)",
+        "enablement": "!SAS.authorizing"
+      },
+      {
+        "view": "sas-content-invalid-connection",
+        "when": "SAS.connectionMethod != rest",
+        "contents": "Your connection does not support SAS content navigation."
+      }
```
scottdover, line 412: @skyeaussie @crlewi Just want to highlight these two bits of text. Will you confirm these are okay or provide updated content

scnwwu:
Hi @scottdover @boyce-w,
When working with the compute service, we decided to make service call via OpenAPI generated code. It ensures we follow the API spec and help developers with inline assist with TypeScript’s type system. Developers don't need to check API and manually write service calls.
I generated API files for `folders` and `files` services and attached to https://github.com/sassoftware/vscode-sas-extension/pull/112
Do you think it can benefit?

scottdover:
> Hi @scottdover @boyce-w, When working with the compute service, we decided to make service call via OpenAPI generated code. It ensures we follow the API spec and help developers with inline assist with TypeScript’s type system. Developers don't need to check API and manually write service calls. I generated API files for `folders` and `files` services and attached to #112 Do you think it can benefit?
 
@scnwwu I think incorporating this is likely a good idea. That said, I'm wondering if we can tackle that in https://github.com/sassoftware/vscode-sas-extension/issues/109, since this could be a larger change. Does that sound acceptable?

scnwwu:
> > Hi @scottdover @boyce-w, When working with the compute service, we decided to make service call via OpenAPI generated code. It ensures we follow the API spec and help developers with inline assist with TypeScript’s type system. Developers don't need to check API and manually write service calls. I generated API files for `folders` and `files` services and attached to #112 Do you think it can benefit?
> 
> @scnwwu I think incorporating this is likely a good idea. That said, I'm wondering if we can tackle that in #109, since this could be a larger change. Does that sound acceptable?

Good to me

package.json
scnwwu: BTW., let's put UI text into the nls file whenever we can

package.json
```diff
@@ -310,6 +369,47 @@
         "language": "sas",
         "path": "./snippets/function-snippets.json"
       }
+    ],
+    "viewsContainers": {
+      "activitybar": [
+        {
+          "id": "sas-view-container",
+          "title": "SAS",
+          "icon": "icons/sas-svg.svg"
```
scnwwu, line 378: we usually need light and dark icons respectively. Will this fit both light and dark themes?

client/src/components/ContentNavigator/viya/DataDescriptor.ts
```diff
@@ -0,0 +1,88 @@
+// Copyright © 2022, SAS Institute Inc., Cary, NC, USA. All Rights Reserved.
```
boyce-w, line 1: This was a sub class of an abstracted class which had another WIP descriptor. They're used to describes the item for different platforms. It had some internal attributes which are now useless or hard coded. If we're going to support this navigation pane only for Viya, it's actually unnecessary to declare this class or declared as a collection of utility functions.

client/src/components/ContentNavigator/viya/ContentModel.ts
```diff
@@ -0,0 +1,417 @@
...
+      item.uri || getLink(item.links || [], "GET", "self")?.uri || null;
+    if (!parentFolderUri) {
+      return;
+    }
+
+    try {
+      const createFolderResponse = await this.connection.post(
+        `/folders/folders?parentFolderUri=${parentFolderUri}`,
+        {
+          name,
```
boyce-w, line 209: "My Favorites" folder and other user created favorite folders only allow to create favorite folders.
{
    name,
    type:  this.dataDescriptor.getTypeName(parentItem) === "favoritesFolder" ?
                    "favoritesFolder" : "folder"
}

package.json
scottdover: Good point. I've a kind of related question. Is there anyway to tap into package.nls.json within our extensions code? I briefly looked into it, but didn't find what I was looking for. It'd be ideal if _all_ of our text came from package.nls.json (imo)

client/src/components/ContentNavigator/viya/DataDescriptor.ts
scottdover: Good point. A class isn't really needed. I moved all functionality to utils.js and got rid of data descriptor

package.json
scottdover: I originally thought this would be fine, because the activity bar is usually a dark color. That said, it looks like themes can change the activity bar color. With that in mind, I'm wondering if you can provide dark and light icons, @skyeaussie ? Or point me to where I can find these?

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+  public readDirectory():
+    | [string, FileType][]
+    | Thenable<[string, FileType][]> {
+    throw new Error("Method not implemented.");
+  }
+
+  public createDirectory(): void | Thenable<void> {
+    throw new Error("Method not implemented.");
+  }
+}
```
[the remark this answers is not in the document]
scottdover: These are the methods that aren't implemented:
 - `delete`
 - `rename`
 - `readDirectory`
 - `createDirectory`

Fwiw, `readDirectory` and `createDirectory` were never triggered in any testing. Thus, those were left unimplemented.

Both `delete` and `rename` were a bit hard to work with in the context of a tree data provider (instead of purely a `FileSystemProvider`). Thus,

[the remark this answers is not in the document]
scottdover: For example, we _could_ implement `delete`. It would involve:
 - Getting the uri of the `ContentItem` when calling delete from the command
 - Calling `getResourceByUri`. This will give us _back_ the ContentItem we already had
 - Finding the delete link and using that to delete the entry

For that one, I think sticking with our custom implementation may make the most sense, especially as we start implementing "recycle bin" functionality where "delete" is a more nuanced concept

package.json
scottdover: After further exploration, it looks like vscode natively implements some kind of bitmasking. As it stands, the icon is black. However, here's how it shows up in vscode. 
Screenshot 2023-02-06 at 1 32 27 PM

In this case, I _think_ what we have is sufficient

scottdover:
@scnwwu I briefly looked into using https://github.com/microsoft/vscode-extension-samples/tree/main/l10n-sample for localization purposes (and having a common place to store strings). That said, this API is very new and doesn't seem to be working as expected. I think it's potentially something to look into in the future, though. It'd be nice to have all of our messaging in one place for ux to review.

client/src/components/ContentNavigator/ContentDataProvider.ts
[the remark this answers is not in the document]
boyce-w: Got it. It makes sense to me.

scnwwu:
> @scnwwu I briefly looked into using https://github.com/microsoft/vscode-extension-samples/tree/main/l10n-sample for localization purposes (and having a common place to store strings). That said, this API is very new and doesn't seem to be working as expected. I think it's potentially something to look into in the future, though. It'd be nice to have all of our messaging in one place for ux to review.
> 
> UPDATE: Looks like having this issue resolved would put us in a better spot - [microsoft/vscode#168127](https://github.com/microsoft/vscode/issues/168127). I'll hold off on these changes until progress is made there

Yeah, l10n should definitely be a separate task. My comment only meant for the package.nls.json file so that we have a good foundation.

package.json
scnwwu: cool

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -0,0 +1,409 @@
...
+          lastModified: res.headers["last-modified"],
+        };
+      } catch (error) {
+        return;
+      }
+    }
+
+    try {
+      const patchResponse = await this.connection.patch(
+        uri,
```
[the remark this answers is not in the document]
scottdover: I'm not entirely sure how to verify this. Here's what I tested:

 - I renamed a file. It was renamed in studio and in sas drive
 - I renamed a "file" in my favorites. The favorite was renamed, but the underlying file was not renamed. This matches the behavior in SAS Studio / SAS Drive

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -0,0 +1,409 @@
...
+      return "programFile";
+    }
+
+    try {
+      const typeResponse = await this.connection.get(
+        `/types/types?filter=contains('extensions', '${ext}')`
+      );
+
+      if (!typeResponse.data.items || typeResponse.data.items.length === 0) {
+        return typeResponse.data.items[0].name;
```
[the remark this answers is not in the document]
scottdover: Ah, you're right. I'm betting I _meant_ to have a question mark/default in there and forgot. I'll adjust that shortly.

boyce-w (approved):

scottdover:
Hey @Becky-Williamson. Can you confirm the issue with deleting files/favorites is resolved?

client/src/commands/closeSession.ts
```diff
@@ -15,4 +14,5 @@ export async function closeSession(message?: string): Promise<void> {
   if (message) {
     window.showInformationMessage(message);
   }
+  commands.executeCommand("setContext", "SAS.authorized", false);
```
scnwwu, line 17: Now that the AuthenticationProvider has been merged to master. It would be great for the content navigator to NOT rely on the compute session. It doesn't really require setting up a compute session, and won't be impacted by compute session close.
It can get the access token from authentication provider. When needed the AuthenticationProviderr can set UI context as appropriate.

client/src/node/extension.ts
smorrisj: @scottdover still working out the connection type values that are accepable. In my working model I have "viya" and "ssh" atm.

client/src/commands/closeSession.ts
scnwwu: OK, sounds like you don't have access right to your user name... I didn't hear this kind of thing before. If that's a valid case, we probably should try catch this part.

scnwwu: Cool. Thank you for catching and fixing it!

scnwwu: Hey, but, this line now can be removed right? Here user doesn't sign out but just close a compute session.

client/src/node/extension.ts
```diff
@@ -80,13 +82,15 @@ export function activate(context: ExtensionContext): void {
   context.subscriptions.push(
     commands.registerCommand("SAS.run", run),
     commands.registerCommand("SAS.runSelected", runSelected),
-    commands.registerCommand("SAS.close", (silent) =>
-      closeSession(silent === true ? undefined : "The SAS session has closed.")
-    ),
+    commands.registerCommand("SAS.close", (silent) => {
+      closeSession(silent === true ? undefined : "The SAS session has closed.");
+      commands.executeCommand("setContext", "SAS.authorized", false);
```
scnwwu, line 87: I'm thinking we don't need this line here. AuthProvider set it to false when sign out that looks sufficient to me.

scottdover: Ugh...missed that one. Should be removed now :)

Becky-Williamson:
Added issue https://github.com/sassoftware/vscode-sas-extension/issues/130

scottdover:
Hey @smorrisj @scnwwu @boyce-w! Can I get one more quick look over this before merging?

client/src/commands/closeSession.ts
scnwwu: @scottdover, would you check this?

package.nls.json
```diff
@@ -1,21 +1,32 @@
...
   "commands.SAS.deleteProfile": "Delete Connection Profile",
-  "commands.SAS.updateProfile": "Update Connection Profile",
-  "commands.SAS.runSelected": "Run Selected or All SAS Code"
+  "commands.SAS.deleteResource": "Delete",
+  "commands.SAS.refreshResources": "Refresh",
+  "commands.SAS.renameResource": "Rename...",
+  "commands.SAS.run": "Run All SAS Code",
+  "commands.SAS.runSelected": "Run Selected or All SAS Code",
+  "commands.SAS.switchProfile": "Switch Current Connection Profile",
+  "commwands.SAS.updateProfile": "Update Connection Profile",
```
scnwwu, line 28: typo

package.nls.json
```diff
@@ -1,21 +1,32 @@
...
+  "configuration.SAS.connectionProfiles.profiles.username": "SAS Viya User ID",
+  "configuration.SAS.session.outputHtml": "Enable/disable ODS HTML5 output",
 
-  "commands.SAS.run": "Run All SAS Code",
-  "commands.SAS.close": "Close Current Session",
-  "commands.SAS.switchProfile": "Switch Current Connection Profile",
+  "commands.SAS.addFileResource": "New File...",
+  "commands.SAS.addFolderResource": "New Folder...",
   "commands.SAS.addProfile": "Add New Connection Profile",
+  "commands.SAS.authenticate": "Sign in",
```
scnwwu, line 17: looks like not used

client/src/commands/closeSession.ts
scottdover: Hey @scnwwu . It appears logging in is currently not working as expected, and I'm unable to verify this change. Unless there's any further things to resolve, I'd recommend getting this one merged in and adjusting this command as needed after login issues are resolved

client/src/commands/authorize.ts
```diff
@@ -0,0 +1,23 @@
+import { authentication, commands } from "vscode";
```
scnwwu, line 1: looks like no copyright header here

client/src/commands/authorize.ts
```diff
@@ -0,0 +1,23 @@
...
+    await switchProfile();
+  }
+
+  if (profileConfig.getActiveProfile() !== "") {
+    try {
+      await authentication.getSession(SASAuthProvider.id, [], {
+        createIfNone: true,
+      });
+    } catch (error) {
+      // Do nothing here. We just want to make sure we're resetting
```
scnwwu, line 17: I'm thinking we should pop any errors up to let user know. And to fail any further actions.
resetting context can be in a "finally" block.

client/src/commands/closeSession.ts
scnwwu: I know the server now has a CSP issue I'll have to communicate it with the team separately. It's not currently anything can be done in client code.
Here's a workaround for now. Click "Copy" for the URL instead of "Open", and paste it into Firefox. Firefox doesn't have such issue.
However, your last screenshot in #131 seems another problem. It seems you have to check the checkbox before Authorize Access.
If you still have issues to authorize, I'm OK to check it after merge. Thanks.

client/src/commands/authorize.ts
scottdover: I've added a call to `window.showErrorMessage`. Although, is pressing escape to prevent auth code entry _really_ an error (in that case, the user _knows_ what they did)?

client/src/commands/closeSession.ts
scottdover: One thing to call out is that you _can't_ copy if your endpoint is configured as a trusted domain. Once I removed it from the list and removed that line, things worked as expected.

client/src/commands/authorize.ts
scnwwu: But there can be other errors like network or server reported errors.
We did see a few errors during authorize when making run code, so made below for error reporting.
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/commands/run.ts#L105

scnwwu (approved):

PR #58, opened 2022-12-13
chore: update pull request template 

scottdover:
**Summary**
This updates our pull request template to have more information around writing a summary & testing changes.

**Testing**

- [ ] Test authenticating with sas server
- [ ] Test running source code against server
- [ ] Make sure results and output display as expected

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,10 +1,7 @@
-**Summary**
+**Summary** [(more info)](CONTRIBUTING.md#summary)
 Please provide a description of your changes.
 
-**Testing**
-The following should be tested with every new pull request. When
-adding new functionality, please include a list of things you
-tested as part of the change.
+**Testing** [(more info)](CONTRIBUTING.md#testing)
 
 - [ ] Test authenticating with sas server
 - [ ] Test running source code against server
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,5 +1,11 @@
 **Summary**
-Please provide a description of your changes
+Please provide a description of your changes.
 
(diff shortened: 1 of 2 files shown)


pull_request_template.md
```diff
@@ -2,4 +2,7 @@
 Please provide a description of your changes
 
 **Testing**
-How did you test this change?
+
+- [ ] Test authenticating with sas server
```
scottdover, line 6: @scnwwu @clangsmith Are there other basic things that should be tested when creating a new pull request?

scottdover:
Hey @scnwwu @smorrisj @clangsmith. Just noticed this one is still hanging out. What do we think about this change?

pull_request_template.md
scnwwu: I'm not sure what's the template should include. I guess the required tests depend on the change scope.
If a change is LSP related, one should probably test the LSP features such as syntax highlighting, auto-complete, etc.
If a change is environment related, one should probably test remote mode or web worker mode.
If a change is related to run code, one should probably test some specific code, like long runs, syntax check mode, ods, etc. Or different kinds of back-end.
Regarding baseline (regression) tests, I guess they'd better be covered by unit tests and automation.

smorrisj: > Regarding baseline (regression) tests, I guess they'd better be covered by unit tests and automation.

Strongly Agree here. Ideally the test automation would catch regressions. A condition of code review should also be that relevant covering unit tests are written.

scnwwu (approved):
Good to me. Thanks.

Issue #46, opened 2022-12-05
Results window opens even if the code being submitted does not contain any ODS output.


JosephDHenry:
running something simple like:

proc options; run;

produces no ODS output, but yet an empty result window appears.

PR #77, opened 2023-01-11
fix: Only show Results window if result is generated #46.

JosephDHenry:
The results window was being displayed for all code executions, regardless of whether or not an actual ODS result was created. This is because the code
is wrapped in ods html5; ... ods html5 close;
This generates an html file, even if it does not
have anything in it.

Signed-off-by: Joseph Henry <joseph.henry@sas.com>

**Summary**
I am now checking for a valid body in the HTML before it is displayed.

**Testing**
I ran simple tests like 

`proc options; run;`

and 

```
data _null_;
x=10;
run;
```

That do not produce any ODS output and confirmed that the Result window does not appear.

Other tests like 

`proc print data=sashelp.cars;run;`

does produce a Result.

### DIFF
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -124,7 +124,6 @@ async function run(code: string, onLog?: (logs: LogLine[]) => void) {
         res.title = result.name;
       }
 
-      res.title = result.name;
       break;
     }
   }
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -116,7 +116,14 @@ async function run(code: string, onLog?: (logs: LogLine[]) => void) {
   for (const result of results.reverse()) {
     const link = result.links[0];
     if (link.type === "text/html") {
-      res.html5 = (await job.requestLink<string>(link)).data;
+      const html5 = (await job.requestLink<string>(link)).data;
+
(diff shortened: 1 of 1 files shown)


client/src/connection/rest/index.ts
```diff
@@ -116,7 +116,14 @@ async function run(code: string, onLog?: (logs: LogLine[]) => void) {
...
-      res.html5 = (await job.requestLink<string>(link)).data;
+      const html5 = (await job.requestLink<string>(link)).data;
+
+      //Make sure that the html has a valid body
+      if (html5.search('<*id="IDX*.+">') !== -1) {
+        res.html5 = html5;
+        res.title = result.name;
+      }
+
       res.title = result.name;
```
scottdover, line 127: Does `res.title` do anything in this case (i.e. if there is no html)?

scottdover (commented):
This looks good to me. Do you want to test this @Zhirong2022?

client/src/connection/rest/index.ts
JosephDHenry: No, Not sure where that came from.

I removed it.

Zhirong2022:
The issue #46 has been fixed in this PR.  No other regression issues.  Mark it as testing-complete.

PR #104, opened 2023-01-28
prepare v0.1.2

scnwwu:
**Summary**
Prepare to publish a new version.

- Drafted changelog and readme changes
- Corrected a dependency. (`media-typer` is used in production code and should not be as devDependencies)
- Added vscodeignore entries for recent added files

**Testing**
No functional changes, no need to test

### DIFF
--- a/client/src/connection/index.ts
+++ b/client/src/connection/index.ts
@@ -1,3 +1,6 @@
+// Copyright © 2022-2023, SAS Institute Inc., Cary, NC, USA. All Rights Reserved.
+// Licensed under SAS Code Extension Terms, available at Code_Extension_Agreement.pdf
+
 import { LogLine } from "./rest/api/compute";
 import { getSession as getRestSession } from "./rest";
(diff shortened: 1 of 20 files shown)


jefurbee (approved):
LGTM

scottdover (approved):

client/src/connection/index.ts
```diff
@@ -1,3 +1,6 @@
+// Copyright © 2022, SAS Institute Inc., Cary, NC, USA. All Rights Reserved.
```
scottdover, line 1: Should this be Copyright 2023? Same comment for the others

scottdover: And, if we update these to 2023, should we go ahead and do it for all the others?

scnwwu: Thank you for pointing it out. I added 2022 because the file was originally authored in 2022.  I checked our rule. It should be updated to be a range when modified in a new year. But if a file is not modified, we don't need to update it. I've go through all files and updated them acoordingly.

Issue #94, opened 2023-01-19
Enhance SAS authentication


scottdover:
**Is your feature request related to a problem? Please describe.**
At the moment, if a user closes & opens a vscode window, the user has to log back in. This leads to more redirection than is necessary, and could be frustrating for the user.

**Describe the solution you'd like**
My hope is that we can use `AuthenticationProvider` and cache the results so that a new window can use the cached access / refresh token to establish a session

clangsmith:
May be worth looking into this [SAS Authentication Library for the Browser](https://github.com/sassoftware/sas-viya-sdk-js/tree/main/sdk/sas-auth-browser) and/or discussing with @brmorr to see if anything we could learn from or re-use. It looks like that is the approach they are recommending for consumers of the VA SDK.

brmorr:
Currently sas-auth-browser supports cookie based auth and will soon also support oAuth with PKCE.  But, as indicated by the package name, is really meant specifically for browser based auth.  But it has been discussed that it may also be desired to have a similar sas-auth-node package.  I am not familiar with how the auth currently works for the SAS extension, or more generally how other VS Code extensions handle auth.  But it seems like this would probably be a node thing, not a browser thing.

PR #110, opened 2023-01-31
feat: support authorization redirect

scnwwu:
**Summary**

- Register SAS as an [AuthenticationProvider](https://code.visualstudio.com/api/references/vscode-api#AuthenticationProvider) to take standard VS Code authentication behavior, which means
  - SAS account will sit in VS Code's standard Accounts menu
  - Authentication can be shared in a standard way across different functionalities (run, content navi, etc.) and even different extensions in future
  - Authentication can persist across VS Code sessions (#94)
- Support direct re-direct from SASLogon so that user don't need to copy/paste authorization code
  - It also depends on server update
  - It uses the standard https://vscode.dev/redirect as the single redirect uri

**Testing**
- Make sure no regressions. User can still paste authorization code
- Close and re-open VS Code, the authentication persist
- It also works on https://vscode.dev with remote extension host

### DIFF
--- a/client/src/connection/rest/auth.ts
+++ b/client/src/connection/rest/auth.ts
@@ -2,7 +2,7 @@
 // Licensed under SAS Code Extension Terms, available at Code_Extension_Agreement.pdf
 
 import axios from "axios";
-import { env, Uri, window } from "vscode";
+import { CancellationTokenSource, env, Uri, window } from "vscode";
 import { URLSearchParams } from "url";
 import { createHash } from "crypto";
 import { Config } from ".";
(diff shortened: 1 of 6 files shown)


scottdover:
> It also depends on server update

Does this mean we need to hold on merging until a server update supports this new behavior? Or does this work in a backwards-compatible way?

client/src/connection/rest/auth.ts
```diff
@@ -13,60 +13,95 @@ interface Tokens {
...
+            client_secret: clientSecret,
+            grant_type: "refresh_token",
+            refresh_token: tokens?.refresh_token,
+          }).toString()
+        )
+        .then(
+          (res) => {
+            tokens = res.data;
+          },
+          () => {
```
scottdover, line 44: Huh. TIL axios has a second argument for `.then`

scottdover (commented):
I think this generally looks okay. I do have a question about what happens when a user has two profiles, stores a session for one profile, then switches to another. Does this code accommodate for that?

scnwwu:
> > It also depends on server update
> 
> Does this mean we need to hold on merging until a server update supports this new behavior? Or does this work in a backwards-compatible way?

We should NOT hold, instead, client update has to be prior to server. It has no effect if merge now. User will still have to copy/paste authorization code as before because server doesn't support it yet. But if server update first, the client will break as the released client doesn't listen for server redirect. After merge, whenever server updated to redirect, it will simply work.

scnwwu:
> I think this generally looks okay. I do have a question about what happens when a user has two profiles, stores a session for one profile, then switches to another. Does this code accommodate for that?

After switching to another profile, any service call will fail and it will prompt user to login again.

scottdover (approved):

Zhirong2022:
Acceptance Criteria:
2.Verify that the user can paste authorization code to run SAS code if having no authorization redirect support on server side
3.Verify that the user does not need to paste authorization code if having authorization redirect support on server side
4.Verify that the user authentication information will persist if the user close VS code and re-open VS code
5.Verity that the user can sign out and sign in VS code successfully
6.Make sure the user name or id is displayed on VS code Acounts menu after he/she signed in and the name is gone if signed out
7.Verify that it keeps a single user authentication information on VS code client side 
8.Make sure the messages make sense during the authorization redirect process
9.Verify that it will prompt connection failure if having network broken issues
10.Make sure there are no NLS issues if switching to other languages in VS code or Browser

Zhirong2022:
The feature works as expected. No regression issues.

Zhirong2022 (approved):

PR #143, opened 2023-03-02
feat: add sas library support

scottdover:
**Summary**

This introduces a new `LibraryNavigator` to facilitate displaying libraries & tables. At the moment, this functionality is limited to:
 - Displaying connected libraries & tables
 - Deleting libraries
 - Using a sample data viewer to see the first 100 rows of a library

Notable things worth mentioning.
 - This introduces a `DragAndDropController` that was written in a somewhat generic way. This should be able to be extended to improve drag/drop functionality of other areas of the code base (content,etc)

**Testing**
NOTE: Acceptance criteria can be found at - https://rndjira.sas.com/browse/PMVSCODE-7

 - [x] Tested viewing libraries and making sure pagination worked as expected
 - [x] Tested viewing the list of tables and making sure pagination worked as expected
 - [x] Made sure opening a table displayed the correct table information
 - [x] Made sure clicking an already open table would present that table instead of opening a new tab
 - [x] Made sure tables were able to be deleted in a similar fashion to how they are on studio (i.e. some tables are not part of readonly libraries, but are also not able to be deleted. In those instances, we show an error to the user matching what one would see in studio)
 - [x] Made sure tables could be dragged/dropped into sas program files


client/src/components/LibraryNavigator/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      }
+    });
+  }
+
+  public async refresh(): Promise<void> {
+    return new Promise((resolve) => {
+      setTimeout(() => {
+        this.libraryDataProvider.refresh();
+        resolve();
+      }, 3 * 1000);
```
scottdover, line 61: This is not ideal, but it seems it takes a couple of seconds after a SAS program runs to load the tables.

client/src/components/LibraryNavigator/LibraryModel.ts
```diff
@@ -0,0 +1,197 @@
...
+              sessionId: this.sessionId,
+              libref: item.id,
+            })
+        );
+
+        return { ...item, readOnly: data.readOnly };
+      })
+    );
+
+    libraryItems.push({
```
JosephDHenry, line 130: This looks quite odd. Is there a reason for this?

scottdover: Sure. `this.dataAccessApi.getLibraries` doesn't return a work library. This just manually adds one

README.md
```diff
@@ -261,3 +277,7 @@ We welcome your contributions! Please read [CONTRIBUTING.md](/CONTRIBUTING.md) f
 ## License
 
 This project is subject to the SAS Code Extension Terms, a copy of which is included as [Code_Extension_Agreement.pdf](/Code_Extension_Agreement.pdf)
+
+```
+
+```
```
scottdover, line 283: I should remove this...not sure how it ended up there

client/src/components/LibraryNavigator/LibraryModel.ts
JosephDHenry: getLibraries returns all assigned libraries, including WORK. If you are not seeing WORK returned, we should figure out why

JosephDHenry: It turns out that on the test server, there were 11 libraries and the default page size is 10. Updating the code to handle paging will fix this.

scottdover:
@skyeaussie / @crlewi There are a couple things to call out here:
 - I added some language around these changes, here: https://github.com/sassoftware/vscode-sas-extension/pull/143/files#diff-0659b008186cbe17a0fcb32ed30a81420d7258e94c20b705ab410a0207c4c7d6R4-R8 . Would you confirm this looks good or suggest any necessary tweaks
 - The language around this needs to be adjusted: https://github.com/sassoftware/vscode-sas-extension/blob/main/package.nls.json#L38 . Basically, it needs to cover these cases:
     - If a user has an ssh connection, they cannot view libraries or sas content
     - If the user is connecting to viya, they can view libraries and sas content
     - if a user is using pup (please cmiiw @JosephDHenry ), they will not be able to view sas content, but they will be able to view libraries after running a file w/ that connection

crlewi:
@scottdover the changes in https://github.com/sassoftware/vscode-sas-extension/pull/143/files#diff-0659b008186cbe17a0fcb32ed30a81420d7258e94c20b705ab410a0207c4c7d6R4-R8 look good.

As far as the other message for connections - I don't think we need to try to explain every combination. It would probably be good to show the existing message if neither pane (library, content) is available. As long as one of the two panes is available then no message. Is that possible?

scottdover:
> @scottdover the changes in https://github.com/sassoftware/vscode-sas-extension/pull/143/files#diff-0659b008186cbe17a0fcb32ed30a81420d7258e94c20b705ab410a0207c4c7d6R4-R8 look good.
> 
> As far as the other message for connections - I don't think we need to try to explain every combination. It would probably be good to show the existing message if neither pane (library, content) is available. As long as one of the two panes is available then no message. Is that possible?

Yeah, I suppose the message is fine as-is. It's a little confusing if you don't _think_ things will show up and then they just _do_ at some point (when using a pup connection). Of course, take this with a grain of salt, because the _only_ way I know how to connect is using viya

client/src/components/LibraryNavigator/index.ts
```diff
@@ -0,0 +1,64 @@
...
+      context.extensionUri
+    );
+    this.treeView = window.createTreeView("sas-library-navigator", {
+      treeDataProvider: this.libraryDataProvider,
+      dragAndDropController,
+    });
+
+    context.subscriptions.push(this.treeView);
+
+    commands.registerCommand(
```
scnwwu, line 32: it is recommended to put registers into `context.subscriptions` to be properly disposed

scottdover: Good catch. I've updated `LibraryNavigator` and `ContentNavigator` to both implement a new `SubscriptionProvider` and tweaked how we setup these commands (see https://github.com/sassoftware/vscode-sas-extension/pull/143/commits/97877263840268c0408f02cdbf610a61691dc74b)

client/src/panels/DataTable/index.ts
```diff
@@ -0,0 +1,105 @@
+import { Disposable, Uri, ViewColumn, WebviewPanel, window } from "vscode";
```
scnwwu, line 1: copyright statement

scottdover: Fixed all the missing copyrights (https://github.com/sassoftware/vscode-sas-extension/pull/143/commits/eab3fdb24b9b1b0deaa27720484a879fbaa02807) and added `npm run copyright:check` to make sure this isn't happening moving forward (https://github.com/sassoftware/vscode-sas-extension/pull/143/commits/78b14e30dc925ca780db414b44da99bdb16b45c4). 

Last, I made this part of our github actions so that we theoretically won't have to worry about checking each file for copyright manually (https://github.com/sassoftware/vscode-sas-extension/pull/143/commits/33f0cffee044858cd1ca5dcd45afd7a596a4443b).

Would you take a close look at this script to make sure I'm looking for copyright information in the "correct" way - https://github.com/sassoftware/vscode-sas-extension/pull/143/files#diff-1ae1758015fed0fadbe09d992dd2d46988a9b6dde1c28ad6ef0bc32b37e73593R4? Right now, it looks like we...
 - expect copyright on code that is used as part of the extension
 - do not expect copyright on certain build files (eslintrc, build.mjs)
 - do not expect copyright on testing files

It's also worth mentioning that this check _only_ looks at files added in the PR

scnwwu: Looks cool! Thanks.

client/src/components/LibraryNavigator/LibraryModel.ts
```diff
@@ -0,0 +1,215 @@
...
+// Licensed under SAS Code Extension Terms, available at Code_Extension_Agreement.pdf
+
+import { ProgressLocation, window } from "vscode";
+import { getSession } from "../../connection";
+import { DataAccessApi } from "../../connection/rest/api/compute";
+import { getApiConfig } from "../../connection/rest/common";
+import { LibraryItemType, LibraryItem, TableData } from "./types";
+import { sprintf } from "sprintf-js";
+import { DefaultRecordLimit, Messages } from "./const";
+import { sortBy } from "lodash";
```
scnwwu, line 11: nit: looking at the usage, it looks like easy to be written with native Array.prototype.sort, which looks has better performance
https://www.measurethat.net/Benchmarks/Show/5571/0/lodash-sortby-vs-native-sortby

scottdover: Duh. Should've used that to begin with. This has been fixed in https://github.com/sassoftware/vscode-sas-extension/pull/143/commits/aaf9496ad7728791d42e5f175963dc08538b7166

Becky-Williamson:
Testing complete.  Verified the following:

- Show connected libraries only
- Allow expansion to see tables in library
- Show correct icons for libraries and tables
- Refresh tree content when SAS code is submitted
- Show new libraries on refresh (ie libname testvs base '/path/to/mydata';
- Show new tables created on refresh
- Delete table
- Note - delete a library is expected only via code. No UI to delete a library
- Show the WORK library
- Drag and drop tables into code (SASHELP.CARS)
- Connect to CAS, assign libraries and create CAS tables

.github/workflows/pr.yml
```diff
@@ -18,6 +18,7 @@ jobs:
           cache: "npm"
       - run: npm ci
       - run: npm run format:check
+      - run: npm run copyright:check
```
smorrisj, line 21: woo hoo!

tools/check-copyright.mjs
```diff
@@ -0,0 +1,34 @@
+import { exec } from "child_process";
```
smorrisj, line 1: This is outstanding. One idea I had (maybe not for this PR) is that we could actually run a theoretical injection of the copyrighted header on changed files, very similar to what we do now. Then, we could run a git command to see if there are modified files as a result of the injection. 

If modified files are detected we'd fail the build.  The added benefit is that now if a contributor runs this locally, they end up with the fixed state ready for staging via git. We do something like this for some of the internal tooling we have for the viya platform and it's worked well I think. We could also autofix, I suppose, but I like still giving the contributor discretion on whether or not the changes get committed / pushed. Thoughts?

client/src/connection/index.ts
```diff
@@ -19,6 +19,7 @@ export interface Session {
   setup(): Promise<void>;
   run(code: string, onLog?: (logs: LogLine[]) => void): Promise<RunResult>;
   close(): Promise<void> | void;
+  sessionId?(): string | undefined;
```
smorrisj, line 22: Trying to remember on this one. Did we promote the sessionId up so that it would make caching/reuse easier based on the discussions we had last week? @JosephDHenry

tools/check-copyright.mjs
scottdover: Yep, this is exactly what I was thinking. As a minimum, we could provide a similar functionality to what lint provides (i.e. we could provide a `--fix` flag to automatically fix the issues for us). I would suggest doing this in another PR, though

client/src/connection/index.ts
scottdover: Joseph's changes are here: https://github.com/sassoftware/vscode-sas-extension/pull/175 . My best guess is we'll have to circle back and make adjustments after his changes are merged

PR #146, opened 2023-03-06
feat: SAS Local Execution

smorrisj:
**Summary**
Adds support for executing SAS code against a local SAS installation.

**Testing**
How did you test this change?


smorrisj:
I'll be working with @jefurbee over the next few days to get the readme updated. In the meantime I'd like to get the code reviewed.

client/src/connection/com/index.ts
```diff
@@ -0,0 +1,260 @@
+import { LogLine, RunResult, Session } from "..";
+import * as fs from "fs";
+import path = require("path");
```
scottdover, line 3:
```suggestion
import { readFileSync } from "fs";
import { resolve } from "path";
```

scottdover: You should be able to use destructured type importing with these

client/src/connection/com/index.ts
```diff
@@ -0,0 +1,260 @@
...
+};
+
+/**
+ * Formats the SAS Options provided in the profile into a format
+ * that the shell process can understand.
+ * @param sasOptions SAS Options array from the connection profile.
+ * @returns a string  denoting powershell syntax for an array literal.
+ */
+const formatSASOptions = (sasOptions: string[]): string => {
+  const optionsVariable = `@("${sasOptions.join(`","`) + `"`})`;
```
scottdover, line 93:
```suggestion
  const optionsVariable = `@("${sasOptions.join(`","`)}")`;
```

client/src/connection/com/index.ts
```diff
@@ -0,0 +1,260 @@
...
+  );
+};
+
+/**
+ * Executes the given input code.
+ * @param code A string of SAS code to execute.
+ * @param onLog A callback handler responsible for marshalling log lines back to the higher level extension API.
+ * @returns A promise that eventually resolves to contain the given {@link RunResult} for the input code execution.
+ */
+const run = async (
```
scottdover, line 178: It looks like we never call `resolve`/`runResolve` here

client/src/connection/com/script.ts
```diff
@@ -0,0 +1,108 @@
+export const scriptContent = `
```
scottdover, line 1: I just skimmed this because I've no idea what's going on

client/src/connection/com/index.ts
smorrisj: @scottdover correct. We save it off and call it in the stdout handler when we see the end mnemonic come in via the stdout.

client/package-lock.json
scnwwu, line 1: Is it intended to only change lock file without changing package.json?

package-lock.json
scnwwu, line 1: Is it intended to only change lock file without changing package.json?

client/package-lock.json
smorrisj: fixed

package-lock.json
smorrisj: fixed

client/src/connection/com/script.ts
```diff
@@ -0,0 +1,108 @@
...
+        $objServerDef.MachineDNSName = $profileHost # SAS Workspace node
+        $objServerDef.Port = 0  # workspace server port
+        $objServerDef.Protocol = 0     # 0 = COM protocol
+        
+        # Class Identifier for SAS Workspace
+        $objServerDef.ClassIdentifier = "440196d4-90f0-11d0-9f41-00a024bb830c"
+
+        # create and connect to the SAS session 
+        $this.objSAS = $objFactory.CreateObjectByServer(
+            "SASApp", # server name
```
clangsmith, line 31: I know it doesn't technically matter, but using "SASApp" as the name makes me a little uncomfortable, since that is the default name for SAS Workspace Servers. Perhaps we could use "Local" instead?

client/src/connection/com/script.ts
```diff
@@ -0,0 +1,108 @@
...
+  [void]ResolveSystemVars(){
+  $code =
+@'
+   %let workDir = %sysfunc(pathname(work));
+   %put &=workDir;
+   %let rc = %sysfunc(dlgcdir("&workDir"));
+   run; 
+'@
+    $this.Run($code)
+    $varLogs = $this.FlushLog(4096)
```
clangsmith, line 14: If a log is over 4MB, are we not going to get it all back?

clangsmith: Oh, nevermind. It looks like that is just for the code immediately above it, which I assume you (and I agree) are confident will always result in a log less than 4MB.

clangsmith (approved):
I am far from an expert in Typescript, but I am generally able to follow the changes. They look good to me, thanks!

client/src/connection/com/script.ts
smorrisj: Exactly right. That log output size should never reach the chunksize.

client/src/connection/com/script.ts
smorrisj:
```suggestion
            "Local", # server name
```

PR #160, opened 2023-03-10
fix: reset ready timeout on successful connection

smorrisj:
**Summary**
Correct the connection ready timeout to clear when a successful connection is made. The previous behavior was prematurely closing the ssh session because the timeout was firing even after a successful connection ready event was encountered. In the timeout function there is code to gracefully end the connection.

This behavior resulted in subsequent run button presses resetting/re-establishing the sas -nodms interactive session. When this happened any intermediate work table content was lost.

**Testing (Developer)**
- [x] Test via steps to reproduce:
        1. Connect to a 9.4 environment with a SAS 9.4 Remote connection profile.
        2. Place the following sas code in the editor contents:
            `data test; set sashelp.class; run;` 
        3. Click run. Verify that no errors appear in the log.
        4. Replace editor contents with code that uses the work table:
            `proc print data=test; run;`
        5. Click run.        
        6. Verify that proc print step from 4 prints the results from the work table to the log.
        7. Verify that result tab shows with 19 observations.
        8. Click the close button.
        9. Replace the editor contents with code from step 4.
        10. Click run.
        11. Verify that the intermediate work table is no longer found with error:
        `File WORK.TEST.DATA does not exist.`

- [x] Test that a sas program that runs for longer than the ready timeout (default 10) seconds value does not prematurely timeout. Program code:
      `data _null_;
          slept=sleep(25,1);
      run;`
- [x] Add appropriate unit tests


smorrisj:
Main issues are fixed. Working on how we might leverage the extension test api to test different aspect of the duplex stream on the ssh connection provider to shore up some coverage gaps.

client/src/connection/ssh/index.ts
```diff
@@ -25,175 +18,199 @@ export interface Config {
...
-          runResult.title = "Result";
-        }
-      }
-      resolve?.(runResult);
+      this.setTimer();
+      this.conn.connect(cfg);
     });
-  });
-}
+  };
```
scottdover, line 143: Can we add new lines between methods? Also, I noticed you used a mix of "traditional" (`a() {}`) style functions and arrow function (`a: () => {}`). Would you mind making all methods here consistent?

client/src/connection/ssh/index.ts
```diff
@@ -25,175 +18,199 @@ export interface Config {
...
+
+            this.stream.on("close", () => {
+              this.onLog = undefined;
+              this.stream = undefined;
+              this.resolve = undefined;
+              this.reject = undefined;
+              this.logs = [];
+              this.html5FileName = "";
+              this.timer = undefined;
+              this.conn.end();
```
scottdover, line 83: What's happening here? Is this "resetting" or "clearing" our session? If so, I might suggest creating a private method to handle that. Then your code could be something like...
```
this.stream.on('close', this.reset)
```

client/src/connection/ssh/index.ts
```diff
@@ -25,175 +18,199 @@ export interface Config {
...
 
-  return new Promise((_resolve, _reject) => {
-    resolve = _resolve;
-    reject = _reject;
-
-    stream?.write(`${code}\n`);
-    stream?.write(`%put ${endCode};\n`);
-  });
-}
+  setTimer() {
```
scottdover, line 168: Can we add method `private`/`protected`/etc to these methods to get a better idea of what's publicly accessible?

client/src/connection/ssh/index.ts
```diff
@@ -25,175 +18,199 @@ export interface Config {
...
-    close,
-  };
+        s.on("data", (data) => {
+          fileContents += data.toString().trimEnd();
+        }).on("close", (code) => {
+          const rc: number = code;
+
+          if (rc === 0) {
+            //Make sure that the html has a valid body
+            //TODO: should this be refactored into a shared location?
```
scottdover, line 205: I feel like I've asked something like this before, but do we have an issue or know how we'll make a decision on this? If we _have_ to leave a todo in code, I'd really like for it to reference an issue so we don't forget it.

client/test/components/ContentNavigator/ContentDataProvider.test.ts
```diff
@@ -92,16 +44,41 @@ const mockContentItem = (
 describe("ContentDataProvider", async function () {
   let authStub;
   beforeEach(() => {
-    authStub = sinon
-      .stub(authentication, "getSession")
-      .resolves({ accessToken: "12345" });
+    authStub = sinon.stub(authentication, "getSession").resolves({
```
scottdover, line 47: I really appreciate you doing all this refactoring. Thank you :)

client/test/connection/ssh/index.test.ts
```diff
@@ -0,0 +1,328 @@
+import * as sinon from "sinon";
+import { Client, ClientChannel } from "ssh2";
+import * as fs from "fs";
+import { SSHSession } from "../../../src/connection/ssh";
+import { assertThrowsAsync, getTestFixtureContent } from "../../utils";
+import { assert, expect } from "chai";
+import { StubbedInstance, stubInterface } from "ts-sinon";
+
```
scottdover, line 8: There are some small inconsistencies with new lines here. In general, I'm wondering if we can...
 - separate describe sections w/ a new line
 - separate each `it`/`beforeEach`/`afterEach` with a newline

client/test/utils.ts
```diff
@@ -14,3 +16,24 @@ export async function openDoc(docUri: vscode.Uri): Promise<void> {
...
+      assert.include(
+        typedError.message,
+        expectedMsg,
+        "Expected Message not found in returned error message"
+      );
+    }
+    return;
+  }
+  assert.fail("function was expected to throw, but did not");
+};
```
scottdover, line 39: Oh. I think I might be able to use this in another pr

client/src/connection/ssh/index.ts
smorrisj: The arrow function was necessary when dealing with the promise return so that the "this" context is properly defined so that I can set state on the instance. Will update the rest to use arrow functions to be consistent.

client/src/connection/ssh/index.ts
smorrisj: So it turns out that getSession() is called for pretty much all 3 phases (setup, run close) of code execution when dealing with sessions. This is attempting to re-initialize the instance fields when the user closes the connection since the instance itself will get reused (due to the getSession item I mention above). Will move some of these heavier blocks of logic off into private methods to make it easier to reason about what's going on.

client/test/utils.ts
smorrisj: Cool. Generally I'm not a fan of having files named "utils". It seems like there is a better name we could come up with. To me "utils" doesnt tell a contributor anything about what the functions in there actually _do_. Definitely open to suggestions on better naming!

client/test/connection/ssh/index.test.ts
smorrisj: No arguments here. Good suggestions. Will make this change.

client/test/connection/ssh/index.test.ts
```diff
@@ -0,0 +1,328 @@
...
+      streamStub.on.callsFake((event, callback) => {
+        if (event === "data") {
+          onDataListener = callback;
+          //need to pass a "?" to the callback to resolve the promise here
+          onDataListener("?");
+          return this;
+        }
+      });
+    });
+    afterEach(() => {
```
scottdover, line 189:
```suggestion

    afterEach(() => {
```

client/test/connection/ssh/index.test.ts
scottdover: There were a couple more left that I marked down below. Otherwise, this looks good to me

scottdover (approved):

PR #162, opened 2023-03-13
fix: improve the new file/folder name validator (#156)

### LINKED ISSUE #156: It failed to add a new file if the file name having Chinese characters
**Describe the bug**
Install the SAS extension package generated base on the main branch.  Sign in successfully and try to add a new file under My Folder. It failed to add a file if the filename contains Chinese characters. However, it can add such kind of file in SASStudio.

**Steps to reproduce**

1. Sign in to show contents pane
2. Click My Folder
3. Open ContextMenu and select New File...
4. Type a file name contains Chinese characters

**Expected behavior**
It can add a file successfully if the file name having Chinese character

**Screenshots**
image

boyce-w:
**Summary**
Fix #156 
I noticed that the folder service does not check the special characters so I changed the folder name validator to only check the max length. If the name is more than 100 characters, there would be a new error message.
And change the validator for the new file name to allow Chinese characters (as well as other languages)

**Testing**
Test case in #156

### DIFF
--- a/client/src/components/ContentNavigator/const.ts
+++ b/client/src/components/ContentNavigator/const.ts
@@ -47,7 +47,8 @@ export const Messages = {
   FileOpenError: "The file type is unsupported.",
   FileValidationError: "Invalid file name.",
   FolderDeletionError: "Unable to delete folder.",
-  FolderValidationError: "Invalid folder name.",
+  FolderValidationError:
+    "The folder name cannot contain more than 100 characters.",
   NewFileCreationError: 'Unable to create file "%(name)s".',
   NewFilePrompt: "Enter a file name.",
   NewFileTitle: "New File",
--- a/client/src/components/ContentNavigator/index.ts
+++ b/client/src/components/ContentNavigator/index.ts
@@ -20,11 +20,13 @@ import { ContentItem } from "./types";
 import { getUri, isContainer as getIsContainer } from "./utils";
 
 const fileValidator = (value: string): string | null =>
-  /^([a-zA-Z0-9\s._-]+)\.\w+$/.test(value)
+  /^([^,/<>?;:\\{}|`=+*&^%#@!~\uff01-\uff5e\u3000-\u303f\u0080-\u00ff]+)\.\w+$/.test(
+    value
+  )
     ? null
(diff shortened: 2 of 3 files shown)


scnwwu (approved):

client/src/components/ContentNavigator/const.ts
```diff
@@ -47,7 +47,8 @@ export const Messages = {
   FileOpenError: "The file type is unsupported.",
   FileValidationError: "Invalid file name.",
   FolderDeletionError: "Unable to delete folder.",
-  FolderValidationError: "Invalid folder name.",
+  FolderValidationError:
```
scottdover, line 50: I think this language change is probably fine, but we should confirm with @skyeaussie / @Becky-Williamson

client/src/components/ContentNavigator/index.ts
```diff
@@ -20,11 +20,13 @@ import { ContentItem } from "./types";
 import { getUri, isContainer as getIsContainer } from "./utils";
 
 const fileValidator = (value: string): string | null =>
-  /^([a-zA-Z0-9\s._-]+)\.\w+$/.test(value)
+  /^([^,/<>?;:\\{}|`=+*&^%#@!~\uff01-\uff5e\u3000-\u303f\u0080-\u00ff]+)\.\w+$/.test(
```
scottdover, line 23: This is a little hard to read, but I think I understand. A valid file contains any character _except_ those in the list above, right? 

In any case, do you mind adding a comment in code about what this regex does?

client/src/components/ContentNavigator/const.ts
crlewi: Yes, this new message matches one of our standard messages and it is acceptable. Thanks

boyce-w: thanks!

client/src/components/ContentNavigator/index.ts
boyce-w: Done

scottdover (approved):

Zhirong2022 (approved):

Issue #163, opened 2023-03-13
The connection was not refreshed automatically if leaving VSC there for a little bit long and without any interactions 


Zhirong2022:
**Describe the bug**
Install the SAS extension package generated base on the main branch. Sign in successfully and try to make some interactions in SAS Pane like Open a file... New folder...  Then, leave VSC a little while,  one hour or more.  Go back to VSC and make some interactions on SAS pane, click an existed file or click My Favorites folder.  Error happened for such sort of actions because of the connection was not refreshed automatically.  

**Steps to reproduce**

1. Sign in to show contents pane
2. Make some interactions towards contents pane like Open a folder
3. Leave VSC a little while, one hour or more
4. Go back to contents pane and open a folder or file 

**Expected behavior**
The connections could be refreshed automatically to avoid errors.

**Screenshots**
![image](https://user-images.githubusercontent.com/121005702/224660346-d8d51c33-0a5a-49d3-8bbd-5afdf79746e6.png)

scottdover:
I think I can re-use some logic I'm currently using for SAS libraries to address this issue. Thanks for bringing this to my attention, @Zhirong2022

PR #171, opened 2023-03-14
fix: reconnect to sas content on 401 error

scottdover:
**Summary**
This fixes an issue where session invalidation caused issues with displaying files and folders (see https://github.com/sassoftware/vscode-sas-extension/issues/163). This addresses the issue by triggering a reconnect when a 401 (Unauthorized) error is encountered.

**Testing**
 - Tested using SAS content as one normally would
 - Left the extension unattended for several hours, and made sure SAS content was refreshed


smorrisj:
Unit tests?

Becky-Williamson:
Verified that leaving the extension for several hours maintains connection.  Tested SAS Explorer pane and found no regressions.

scottdover:
This should now be taken care of, @smorrisj . I'm curious if we're able to use this for things w/n compute 🤔

smorrisj:
> This should now be taken care of, @smorrisj . I'm curious if we're able to use this for things w/n compute 🤔
> 
> Also, this broke all the tests 😢

Seems it would be valid for connections where we have to make use of the viya apis that deal with auth tokens. Not sure about other connections (such as direct where we dont have viya in the mix).

smorrisj (approved):
LGTM

PR #172, opened 2023-03-15
Directconnection

JosephDHenry:
**Summary**
This is a change set made to be an addition onto adding Library support #143 

If a direct connection is being made (IE, using the compute REST API but not going through Viya services, authentication is not done viya SASLogon, so the authentication requirement is not valid.

This change allows the Libraries view when a direct connection is established but also disables the content browser, since the Files and Folders service is not available. 

**Testing**
I made sure that libraries show up with a direct connection.


client/src/node/extension.ts
```diff
@@ -151,11 +151,20 @@ function triggerProfileUpdate(): void {
...
-      "SAS.connectionType",
-      profileList[activeProfileName].connectionType || ConnectionType.Rest
-    );
+
+    const connectionType =
+      profileList[activeProfileName].connectionType || ConnectionType.Rest;
+
+    //Set the connection type
+    commands.executeCommand("setContext", "SAS.connectionType", connectionType);
+    if (connectionType === ConnectionType.Rest) {
```
scottdover, line 160: I think could probably be simplified to something like...
```js
commands.executeCommand(
  'setContext',
  'SAS.connection.direct',
  connectionType === ConnectionType.Rest && !!profileList[activeProfileName]?.serverId
)
```

but this seems like a simple enough change to incorporate. I was imagining something more complicated 😁. We will have to talk to Sam/Cristina/Becky to see how this changes messaging on the welcome screen, but that needs to be adjusted now anyway since it reads...

> Your connection does not support SAS content navigation within the SAS Extension for Visual Studio Code. You can access files only using the Explorer view.

PR #173, opened 2023-03-15
fix: Support SAS Library viewer with a direct connection.

JosephDHenry:
**Summary**
This is a change set made to be an addition onto adding Library support https://github.com/sassoftware/vscode-sas-extension/pull/143

If a direct connection is being made (IE, using the compute REST API but not going through Viya services, authentication is not done viya SASLogon, so the authentication requirement is not valid.

This change allows the Libraries view when a direct connection is established but also disables the content browser, since the Files and Folders service is not available.

**Testing**
I made sure that libraries show up with a direct connection.

### DIFF
--- a/client/src/node/extension.ts
+++ b/client/src/node/extension.ts
@@ -163,7 +163,7 @@ function triggerProfileUpdate(): void {
       "setContext",
       "SAS.connection.direct",
       connectionType === ConnectionType.Rest &&
-        !!profileList[activeProfileName]?.serverId
+        "serverId" in profileList[activeProfileName]
     );
   } else {
     profileConfig.updateActiveProfileSetting("");
--- a/client/src/node/extension.ts
+++ b/client/src/node/extension.ts
@@ -157,14 +157,14 @@ function triggerProfileUpdate(): void {
 
     //Set the connection type
     commands.executeCommand("setContext", "SAS.connectionType", connectionType);
-    if (connectionType === ConnectionType.Rest) {
-      //Check to see if we are directly connected (Ie. we have a serverId)
-      if ("serverId" in profileList[activeProfileName]) {
-        commands.executeCommand("setContext", "SAS.connection.direct", true);
-      }
-    } else {
-      commands.executeCommand("setContext", "SAS.connection.direct", false);
-    }
+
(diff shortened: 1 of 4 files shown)


client/src/node/extension.ts
```diff
@@ -151,11 +151,20 @@ function triggerProfileUpdate(): void {
...
+    //Set the connection type
+    commands.executeCommand("setContext", "SAS.connectionType", connectionType);
+    if (connectionType === ConnectionType.Rest) {
+      //Check to see if we are directly connected (Ie. we have a serverId)
+      if ("serverId" in profileList[activeProfileName]) {
+        commands.executeCommand("setContext", "SAS.connection.direct", true);
+      }
+    } else {
+      commands.executeCommand("setContext", "SAS.connection.direct", false);
+    }
```
scottdover, line 167: I think this could probably be simplified to something like...
```js
commands.executeCommand(
  'setContext',
  'SAS.connection.direct',
  connectionType === ConnectionType.Rest && !!profileList[activeProfileName]?.serverId
)
```

but this seems like a simple enough change to incorporate. I was imagining something more complicated 😁. We will have to talk to Sam/Cristina/Becky to see how this changes messaging on the welcome screen, but that needs to be adjusted now anyway since it reads...

> Your connection does not support SAS content navigation within the SAS Extension for Visual Studio Code. You can access files only using the Explorer view.

client/src/node/extension.ts
```diff
@@ -151,10 +151,19 @@ function triggerProfileUpdate(): void {
...
+    commands.executeCommand("setContext", "SAS.connectionType", connectionType);
+
+    //See if the connection is direct (ie. serverId)
     commands.executeCommand(
       "setContext",
-      "SAS.connectionType",
-      profileList[activeProfileName].connectionType || ConnectionType.Rest
+      "SAS.connection.direct",
+      connectionType === ConnectionType.Rest &&
+        !!profileList[activeProfileName]?.serverId
```
scottdover, line 166: I was under the impression `serverId` existed on profile, or I'm missing it somewhere. It doesn't _seem_ that it does?

scottdover (approved):

PR #175, opened 2023-03-17
feat: Reconnect to existing SAS Session if possible.

JosephDHenry:
A sas session connection is made as needed by the extension. This session is maintained throughout the vscode session. Many things can cause the extension to reload (browser refresh, vscode reload, etc) and this causes a new connection to be established, which creates a brand new SAS session.

If an existing session is available, we should reconnect to it.

This is gated behind a REST profile option "reconnect"

**Summary**
Added a way to store key value pairs that persist through extension reloads. This uses the extension context.

If a sessionId is found, that session will be attempted to be reconnected to. If this succeeds, all future interaction will be
done on that session instead of creating a new on.

This also fixes a bug in JobState which causes unwanted server spam.

**Testing**
I made sure that Sessions will be reconnected to and that new sessions will be created if needed.

### DIFF
--- a/client/src/components/ExtensionContext.ts
+++ b/client/src/components/ExtensionContext.ts
@@ -19,6 +19,8 @@ export async function setContextValue(
 /*
  * Get an extension context value.
  */
-export async function getContextValue(key: string): Promise<string> {
+export async function getContextValue(
+  key: string
+): Promise<string | undefined> {
   return context.workspaceState.get(key);
 }
--- a/client/src/components/ExtensionContext.ts
+++ b/client/src/components/ExtensionContext.ts
@@ -1,0 +1,24 @@
+import { ExtensionContext } from "vscode";
+
+let context: ExtensionContext;
+
+export function setContext(c: ExtensionContext) {
+  context = c;
+}
+
+/*
+ * Set an extension context value.
+ */
+export async function setContextValue(
+  key: string,
+  value: string
+): Promise<void> {
+  context.workspaceState.update(key, value);
+}
+
+/*
+ * Get an extension context value.
+ */
+export async function getContextValue(key: string): Promise<string> {
+  return context.workspaceState.get(key);
+}
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -1,9 +1,13 @@
 // Copyright © 2022-2023, SAS Institute Inc., Cary, NC, USA. All Rights Reserved.
 // Licensed under SAS Code Extension Terms, available at Code_Extension_Agreement.pdf
 
-import { authentication, commands } from "vscode";
+import { authentication } from "vscode";
 import { RunResult, Session } from "..";
 import { SASAuthProvider } from "../../components/AuthProvider";
+import {
+  getContextValue,
+  setContextValue,
+} from "../../components/ExtensionContext";
 import { ContextsApi, LogLine } from "./api/compute";
 import { ComputeState, getApiConfig } from "./common";
 import { ComputeJob } from "./job";
@@ -57,10 +61,7 @@ async function setup(): Promise<void> {
     const server1 = new ComputeServer(config.serverId);
 
     //if we have a session already, get it
-    let sessionId: string = await commands.executeCommand<string>(
-      "SAS.contextValue.get",
-      "SAS.sessionId"
-    );
+    let sessionId: string = await getContextValue("SAS.sessionId");
     if (!config.reconnect) {
       sessionId = undefined;
     }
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -397,6 +407,10 @@
           "when": "editorLangId == sas && !SAS.hideRunMenuItem",
           "command": "SAS.run"
         },
+        {
```
scnwwu, line 410: looks like duplicated?

package.json
```diff
@@ -330,6 +330,16 @@
         "title": "%commands.SAS.collapseAll%",
         "category": "SAS",
         "icon": "$(collapse-all)"
+      },
+      {
```
scnwwu, line 334: It looks like not make sense for end user to call, right?
If so, how about somehow just export the function for other modules to call but not as a command?

client/src/connection/rest/index.ts
```diff
@@ -54,10 +55,33 @@ async function setup() {
 
   if (config.serverId) {
     const server1 = new ComputeServer(config.serverId);
+
+    //if we have a session already, get it
```
scnwwu, line 59: It looks like only do if serverId exists. What about no serverId? I think it also makes sense to reconnect, right?

client/src/connection/rest/index.ts
```diff
@@ -16,6 +16,7 @@ export interface Config {
   clientSecret?: string;
   context?: string;
   serverId?: string;
+  reconnect?: boolean;
```
scnwwu, line 19: Do we want to add configuration entry (in Settings UI) for it?

package.json
JosephDHenry: Yeah, Not sure how that got in there. I removed it.

package.json
JosephDHenry: Let me explain why I did it this way.

In order to save state like this, we need the extension context, which is only available on extension creation. 

So either I save the extension context in some global access area, Or I create a command that gets and sets values like this.

It seemed to me that directly exposing the extension context was not a good thing to do.

Also, even though this is a command, it is not callable by the user, we are simply using vscodes command interface to enable actions like this.

client/src/connection/rest/index.ts
JosephDHenry: I had not tested this feature in a Viya environment, so I limited the reconnection to the serverId case for now. I planned on revisiting the Viya case after discussion on redesign of the Session interface.

package.json
scnwwu: I've put my idea to
https://github.com/sassoftware/vscode-sas-extension/pull/183

client/src/connection/rest/index.ts
```diff
@@ -54,10 +59,26 @@ async function setup() {
 
   if (config.serverId) {
     const server1 = new ComputeServer(config.serverId);
+
+    //if we have a session already, get it
+    let sessionId: string = await getContextValue("SAS.sessionId");
```
scottdover, line 64: I think there's still more to discuss here around session IDs and a session interface. Ultimately, I'd love to see a function on a `Session` class/interface/etc because it makes sense conceptually (i.e. "a session has an identity"). 

For now, I think this is a good addition as we'll always likely need context to persist our session id across reloads/etc. With this in place, I _could_ rely on `getContextValue('SAS.sessionId')` to give me the session id I need (assuming that could also change as sessions evolve)

client/src/components/ExtensionContext.ts
```diff
@@ -0,0 +1,24 @@
+import { ExtensionContext } from "vscode";
```
scottdover, line 1: This seems like a persistence layer for a workspace. I'm wondering if our function names should be more inline with the abstract concept or persistence rather than the specific implementation. Perhaps this should expose `get` and `persist` or `persistValue`/`getPersistedValue`. Thoughts?

client/src/node/extension.ts
scnwwu, line 1: I can't add comment to line 195 so let me comment here. I guess we should remove `closeSession` call from `deactivate` with this change.
Also I guess it would be good to clear stored sessionId when user close session.

client/src/connection/rest/index.ts
JosephDHenry: Yeah, you could definitely use this to get the sessionId and it could easily change as the session interface changes.

package.json
```diff
@@ -458,6 +458,10 @@
           "when": "editorLangId == sas && !SAS.hideRunMenuItem",
           "command": "SAS.runSelected"
         },
+        {
```
scnwwu, line 461: well, looks like duplicated

scnwwu (commented):
Looks good to me except a merge issue. Thanks.

package.json
JosephDHenry: fixed

scnwwu (approved):

PR #188, opened 2023-03-23
fix: remove foreach when iterating with promise

smorrisj:
**Summary**
- Removes code that was errantly not migrating all legacy profiles over. This was due to use of foreach inside of a promise, which will not run to completion.
- Additionally add a validator to profile validator to detect a missing connectionType property when `getSession()` is called during execution.

**Testing**
- [x] add unit test for calling function directly against a given set of legacy profiles
- [x] add unit test for  positive case where a legacy profile is missing connectionType property.
- [x] manual verification: launched the extension with multiple profiles that did not have a connectionType key. The expected result is that on extension activation, these profiles are migrated as expected to be rest profile.

### DIFF
--- a/client/src/components/profile.ts
+++ b/client/src/components/profile.ts
@@ -105,6 +105,23 @@ export interface ProfileValidation {
  */
 export class ProfileConfig {
   /**
+   * Helper function to migrate legacy profiles without a connection type.
+   */
+  async migrateLegacyProfiles() {
+    const profiles = this.getAllProfiles();
+
+    if (profiles) {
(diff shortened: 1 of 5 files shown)


scnwwu (approved):
looks good to me

Becky-Williamson:
Verified that "connectionType" is added to all profiles that do not have the setting on submitting code for the first time to any of those profiles.

Verified that when a profile has an invalid "connectionType" a message stating "Invalid connectionType. Check Profile settings" is displayed.

Issue #191, opened 2023-03-24
It has no error validation if trying to rename a file name having NLS characters to an existed one


Zhirong2022:
**Describe the bug**
Install the SAS extension package generated base on the main branch. It has no error validation if trying to rename a file name having NLS characters to an existed one.  

**Steps to reproduce**
1. New two files having NLS characters and they have the same file name and different file type
2. Rename one of them to another one 

**Expected behavior**
Error validation if trying to rename a file name to an existed one

**Screenshots**
Error validation for the filename with English characters for reference
![image](https://user-images.githubusercontent.com/121005702/227493226-a79e1407-58ba-4d8a-bef0-6e1cea23dc27.png)

PR #193, opened 2023-03-27
validate the new name before rename

boyce-w:
**Summary**
fix the issue #191 
validate the new name with folder service before requesting the rename action

**Testing**
As described in #191 
Add unit tests


Zhirong2022:
Error validation works as expected.

Zhirong2022 (approved):

PR #199, opened 2023-03-27
Feat/129/2

scottdover:
**Summary**

**Testing**
How did you test this change?


scnwwu:
Hi Scott, just bringing an idea that I saw Microsoft is using [Preact](https://preactjs.com/) in some of their webviews (for [example](https://github.com/microsoft/vscode-github-issue-notebooks/blob/main/src/renderer/renderer.tsx#L5)) rather than React. VS Code may render many webviews. Preact looks like having some advantage of lightweight and performance. Just FYI.

scottdover:
NOTE: Re-opening a file that was once open does not work

Becky-Williamson:
I created a table with 250 columns and 200,000 rows.  The table opened and I can page through it to a point.  Eventually, it appears the memory is maxing out and it hangs.  The program I ran to create this table is:

`%macro CreateDatasetWithDimensions(numCols, numRows);

data work.Cols&numCols.Rows&numRows (drop=rowNum);
  do rowNum = 1 to &numRows;
  	%local i;
    %do i = 1 %to &numCols;
  		col&i="Column &i, Row " || left(put(rowNum, 7.));
	%end;
	output;
  end;
run;

%mend CreateDatasetWithDimensions;

%CreateDatasetWithDimensions(250,200000)`

Also, we need to somehow format the data view a little better.  This is what the results of that code looks like in the data viewer.  

image

Becky-Williamson:
One other issue that I've seen is that there is a context menu displayed if you right click in a cell in the data view with menu options Cut, Copy and Paste.  None of these options appear to work so we should probably either remove the context menu or disable the menu options.  I don't think making them work is in the scope for this release.

scottdover:
@Becky-Williamson and I chatted about the things above. Here's a summary:
 - Large datasets do pose a problem. I know @clangsmith mentioned something about a "windowed" paging system (keeping only a few pages of results loaded at a time). This isn't included in this change set, but could be something that was worked on a bit more. My question is do we want to put the effort there, or wait until AG grid details are sorted?
 - There are two current issues in main: Table columns being too narrow and context menus not working. These are both present in main. 

So my question is, what do we want to include in the 0.1.5 release? This data viewer is an improvement over only viewing the first 100 rows, but it does have some current limitations

PR #202, opened 2023-03-30
feat: ssh agent auth

smorrisj:
**Summary**
- Adds support for encrypted private keys that are registered with ssh-agent
- Private Key file is removed from the connection profile and new connection prompts and is now instead defined in the global ssh config
- SSH Connection Provider is updated to read process.env.SSH_AUTH_SOCK (env var) and set Client agent property to this value at connection time

# Testing (Developer) #
### Windows (OpenSSH installed as a Windows Feature) ###
1. Enabled open ssh client optional feature
2. Created env variable SSH_AUTH_SOCK with value //./pipe/openssh-ssh-agent (windows uses a named pipe for the auth sock)
3. Ensure ssh-agent service is running, set startup type to automatic
4. Define an entry in $HOME/.ssh/config of the form:
```  
    Host host.machine.name 
        AddKeysToAgent yes
        IdentityFile /path/to/private/key/with/passphrase 
``` 
5. Add the private key to ssh-agent: `ssh-add /path/to/private/key/with/passphrase`
6. Define a connection profile in settings.json for a remote server:

```
    "ssh_test": {
        "connectionType": "ssh",  
        "host": "host.machine.name",  
        "saspath": "/path/to/sas/executable",  
        "username": "username",  
        "port": 22
    } 
```
7. Run test code using ssh_test connection profile. Should get 428 observations and results tab displayed.
```
    proc print data=sashelp.cars;  
    run;  
```

### macOS ###
1. Start ssh-agent in the background:
```
    eval "$(ssh-agent -s)"
```
2. Ensure that SSH_AUTH_SOCK has a value
```
   echo $SSH_AUTH_SOCK
```
4. Define an entry in $HOME/.ssh/config of the form:
```  
    Host host.machine.name 
        AddKeysToAgent yes
        UseKeychain yes
        IdentityFile /path/to/private/key/with/passphrase 
```
5. Add the private key to ssh-agent: `ssh-add /path/to/private/key/with/passphrase`
6. Define a connection profile in settings.json for a remote server:
```
    "ssh_test": {
        "connectionType": "ssh",  
        "host": "host.machine.name",  
        "saspath": "/path/to/sas/executable",  
        "username": "username",  
        "port": 22
    } 
```
7. Run test code using ssh_test connection profile. Should get 428 observations and results tab displayed.
```
    proc print data=sashelp.cars;  
    run;  
```

### DIFF
--- a/client/src/connection/ssh/index.ts
+++ b/client/src/connection/ssh/index.ts
@@ -24,14 +24,14 @@ export function getSession(c: Config): Session {
   if (!sessionInstance) {
     sessionInstance = new SSHSession();
   }
-  sessionInstance.config = c;
+  sessionInstance.setConfig(c);
   return sessionInstance;
 }
 
 export class SSHSession implements Session {
   private conn: Client;
   private stream: ClientChannel | undefined;
-  public config: Config;
+  private config: Config;
   private resolve: ((value?) => void) | undefined;
   private reject: ((reason?) => void) | undefined;
   private onLog: ((logs: LogLine[]) => void) | undefined;
@@ -43,6 +43,10 @@ export class SSHSession implements Session {
     this.config = c;
     this.conn = new Client();
   }
+
+  public setConfig = (config: Config) => {
+    this.config = config;
+  };
 
   public setup = (): Promise<void> => {
     return new Promise((pResolve, pReject) => {
@@ -60,6 +64,9 @@ export class SSHSession implements Session {
         username: this.config.username,
         readyTimeout: sasLaunchTimeout,
         agent: process.env.SSH_AUTH_SOCK || undefined,
+        debug: (message: string) => {
+          console.log(message);
+        },
       };
 
       this.conn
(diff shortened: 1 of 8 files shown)


smorrisj:
@jefurbee  wanted to get your thoughts on the doc changes?

client/src/connection/ssh/index.ts
```diff
@@ -15,28 +14,32 @@ export interface Config {
...
   }
+  sessionInstance.config = c;
   return sessionInstance;
 }
 
 export class SSHSession implements Session {
   private conn: Client;
   private stream: ClientChannel | undefined;
-  private config: Config;
+  public config: Config;
```
JosephDHenry, line 34: I feel like it would be better to leave this private and create a Getter. This also makes more sense now that the config is an optional parameter in the constructore.

smorrisj: @JosephDHenry  fixed, see updated.

JosephDHenry (approved):

ShuguangSun:
So server side should be linux-like? In my orgnization, SAS is intalled in a Windows server (2019). We started the openssh server in the window box. However, it seems the command submmited is for linux-like box.

```
_JAVA_OPTIONS="-Djava.awt.headless=true" C:\sas\SASHome\SASFoundation\9.4\sas.exe -nodms -terminal -nosyntaxcheck
endsas;
```

PR #208, opened 2023-04-05
chore: update changelog

smorrisj:
**Summary**
Changelog update for various issues fixed during milestone

**Testing**
Visual markdown inspection, tested links


2TomLi (approved):

scnwwu:
Hi @smorrisj,
This change looks a little weird to me. Normally bug fixes should go into the `### Fixed` section. However, the bugs are not from v0.1.2.
The changelog is not only for github, but also for marketplace. In my opinion, changes should be compared to released version (v0.1.2). Perhaps they can be merged to the remote 9.4 item.

PR #211, opened 2023-04-06
fix: fix miscellaneous bugs for 0.1.3 deployment

scottdover:
**Summary**
This fixes some inconsistencies and bugs around the 0.1.3 deployment. Here's the summary:
 - Fixes an issue in a readme
 - Fixes a missing refresh command for SAS content
 - Adds the ability to collapse libraries
 - Fixes a console error when there is no active profile
 - Fixes a bug where _no_ profile resulted in endlessly writing to user settings
 - Fixes a bug where _no_ profile didn't allow users to sign in using the SAS sign in view

**Testing**
 - [x] Made sure sign in was possible with no profile
 - [x] Made sure there were no console errors or issues when writing an empty profile
 - [x] Made sure sas content and libraries were both collapsible
 - [x] Made sure the readme looked as expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ To access SAS Content:
 
 **Notes**:
 
-- # SAS Content requires a profile with a connection to a Viya instance.
+- SAS Content requires a profile with a connection to a Viya instance.
 
 ### Accessing Libraries and Tables
 
--- a/client/src/components/AuthProvider.ts
+++ b/client/src/components/AuthProvider.ts
@@ -46,8 +46,8 @@ export class SASAuthProvider implements AuthenticationProvider, Disposable {
 
     const session: SASAuthSession = JSON.parse(stored);
     const activeProfile = profileConfig.getActiveProfileDetail();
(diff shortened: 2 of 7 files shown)


smorrisj (approved):

scottdover:
Hey @Becky-Williamson . Do you mind doing a quick once over to make sure this looks good to go? Most of the issues I found were exposed by starting with a missing config. I tested this by:
 - Removing `SAS.connectionProfiles` entirely from user settings JSON (you can find this by pressing ctrl + shift + p and searching for "user settings")
 - Installing the extension (or using launch client) and starting the extension

One other thing to mention is that this includes a new feature to collapse libraries, where it will be consistent w/ SAS content

scottdover:
Chatted w/ @Becky-Williamson about this one in Teams. It seems that this didn't introduce any new regressions. However, it uncovered some connectivity issues affecting Windows. These will be looked into further with this bug - https://github.com/sassoftware/vscode-sas-extension/issues/215

PR #221, opened 2023-04-12
feat: support user provided CA certificates

### LINKED ISSUE #220: NODE_EXTRA_CA_CERTS doesn't work
To workaround the `unable to verify the first certificate` error on Windows, [NODE_EXTRA_CA_CERTS](https://nodejs.org/api/cli.html#node_extra_ca_certsfile) was provided in our [FAQ](https://github.com/sassoftware/vscode-sas-extension/wiki/FAQ#i-got-unable-to-verify-the-first-certificate-error-when-run).
However it does NOT really work due to https://github.com/electron/electron/issues/10257 (It said fixed but not really, see the latest comments).
We need a way for users to add extra certs.

scnwwu:
**Summary**
Fix #220
Add a configuration field so that user can provide their certificates

This page describes some background. But the VS Code extension provided by this page does NOT really work.
https://github.com/ukoloff/win-ca

I'll update the FAQ page once this merged.

**Testing**
1. Export a crt file from browser against an endpoint with self-signed certificate.
2. In VS Code settings UI, click "Add Item" under "User Provided Certificate" entry, enter the full file path for the crt file.
3. Restart VS Code and Sign in, success

### DIFF
--- a/client/src/components/CAHelper.ts
+++ b/client/src/components/CAHelper.ts
@@ -14,18 +14,18 @@ export const installCAs = () => {
     return;
   }
 
-  const CAs = [].concat(tls.rootCertificates);
-  const originalLength = CAs.length;
+  const userCertificates = [];
   for (const filename of certFiles) {
     if (filename && filename.length) {
       try {
-        CAs.push(fs.readFileSync(filename));
+        userCertificates.push(fs.readFileSync(filename));
       } catch (e) {
         console.log(`Failed to read user provided certificate`, e);
       }
     }
   }
-  if (CAs.length !== originalLength) {
-    https.globalAgent.options.ca = CAs;
+  if (userCertificates.length > 0) {
+    https.globalAgent.options.ca =
+      tls.rootCertificates.concat(userCertificates);
   }
 };
--- a/client/src/components/CAHelper.ts
+++ b/client/src/components/CAHelper.ts
@@ -1,0 +1,31 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA. All Rights Reserved.
+// Licensed under SAS Code Extension Terms, available at Code_Extension_Agreement.pdf
+
+import * as fs from "fs";
+import * as https from "https";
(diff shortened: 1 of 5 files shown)


client/src/components/CAHelper.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const installCAs = () => {
+  const certFiles: string[] = workspace
+    .getConfiguration("SAS")
+    .get("userProvidedCertificates");
+  if (!certFiles || !certFiles.length) {
+    return;
+  }
+
+  const CAs = [].concat(tls.rootCertificates);
```
scottdover, line 17: Are we using this just to clone `rootCertificates`? If so, can we use something like this instead...
```suggestion
  const CAs = [ ...tls.rootCertificates];
```

scnwwu: I revised a little. It doesn't need to copy root certificates before assigning to agent.
Are you still thinking spread operator better than `concat` in the latest code? I'm thinking `concat` should have better performance than spread operator.

scottdover: In general, `concat` seems to be more performant (for large arrays). That said, the spread syntax to my eye more clearly highlights what ~is~ was being done. Personally, I think most of the time (I know we have cases where this isn't the case), code that is easier to understand is much better than code that is performant (where the performance savings aren't significant).

All that said, I think your revised approach looks much better :)

scottdover (approved):

Zhirong2022:
Acceptance Criteria:
2.Verify that it can add certificate file successfully
3.Make sure the error 'unable to verify the first certificate' is gone after setting a valid certificate file  
4.Verify that it does not work properly if setting an invalid certificate file 
4.1 Invalid certificate file location
4.2 Invalid certificate file name
4.3 Invalid certificate file contents
5.Verify 1~4 in OS: Windows and Mac
6.Verify 1~5 in NLS environment

Zhirong2022 (approved):

PR #232, opened 2023-04-13
fix: fix trailing slash on viya profile endpoints

scottdover:
**Summary**
This fixes an issue where trailing slashes can lead to 404/connection errors. This solves the issue in two ways:
 - It trims slashes when creating new profiles
 - It upgrades legacy profiles to add a trailing slash

**Testing**
 - [x] Test migrating legacy profile with trailing slash on endpoint
 - [x] Test creating new profile with trailing slash on endpoint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


scottdover:
Need to write a couple tests for this one

Becky-Williamson:
Verified that trailing / is removed from existing profile endpoints.
Verified that creating a new profile with an endpoint containing a trailing / is saved without the /.

smorrisj (approved):
LGTM

PR #243, opened 2023-04-19
fix: use vscode.open to open sas content

### LINKED ISSUE #224: Double-clicking a file in Explorer opens it in the preview editor instead of opening in a new tab
**Describe the bug**
By default, when you single click a file in the Explorer in VS Code, the file is opened into a preview editor. This is a very convenient way to preview the contents of files without having to explicitly open them in new tabs in your workspace. From a preview editor, you can explicitly choose to keep it open from the tab's context menu, by editing the contents of the file, or by double-clicking the file in Explorer. (Note: You can disable this feature by unchecking the "Workbench>Editor: Enable Preview" option in the VS Code settings.)

SAS' Explorer currently honors the preview editor, however it does not honor the double-click behavior, thus inconsistent with VS Code's Explorer. When you double-click separate files in SAS' Explorer each file is still opened in the same preview editor (same tab) rather than kept open in a new tab.

**Steps to reproduce**

1. In VS Code settings, make sure the "Workbench>Editor: Enable Preview" option is checked on (the default).
2. In SAS pane, Explorer section, save two .sas files anywhere in SAS Content (using VS Code or SAS Studio to create the files is fine).
3. Close the files so they are not open in VS Code.
4. Double-click one of the files to open.
5. Notice that the file is opened in a preview editor, which is reflected by the italic text in the tab header.
6. Double-click the other .sas file to open.
7. Notice that this file is opened in the same preview editor, replacing the first, rather than opened in a new tab.

**Expected behavior**
When double-clicking a file in Explorer, it should open in a new "kept-open" tab rather than in a preview editor. When single-clicking a file in Explorer, it should open in a single preview editor (unless the VS Code setting has been turned off). In other words, it should be consistent with how it behaves in VS Code's Explorer pane.

Client OS: Windows 10
Extension version: v0.1.3
SAS version: Viya 2023.04

**Additional context**
I noticed that tables in the Libraries view do not currently behave consistent with VS Code's Explorer either. Currently, single-click opens each table in a new tab -- it doesn't use the preview editor even when the option is on. I'm thinking we should be consistent and use preview on single-click and keep open on double-click there as well. One concern though is whether opening the data may be too expensive an operation to open in preview. If that's the case though, I would think we wouldn't open on single-click at all, and rather, require double-click (more explicit action) to open. For now, I think it's better to be consistent. Can open another issue if you want to track these (Explorer and Libraries pane) separately.

scottdover:
**Summary**
This updates our code to use vscode.open instead of a custom open command. This takes advantage of preview mode and either opens in a preview w/ single click or opens for editing w/ a double-click

**Testing**
 - Tested opening files with preview mode on and off, and made sure functionality worked as expected

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -10,6 +10,8 @@ import {
   FileSystemProvider,
   FileType,
   ProviderResult,
+  Tab,
+  TabInputText,
   TextDocumentContentProvider,
   ThemeIcon,
   TreeDataProvider,
@@ -17,20 +19,18 @@ import {
   TreeItemCollapsibleState,
   Uri,
   window,
-  Tab,
-  TabInputText,
 } from "vscode";
 import { ContentModel } from "./ContentModel";
 import { ContentItem } from "./types";
 import {
   getCreationDate,
   getId,
+  isContainer as getIsContainer,
   getLabel,
+  getLink,
   getModifyDate,
   getUri,
-  isContainer as getIsContainer,
   resourceType,
-  getLink,
 } from "./utils";
 
 class ContentDataProvider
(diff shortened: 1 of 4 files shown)


smorrisj (approved):
LGTM

Becky-Williamson:
Verified:

1. Single click on file 1 opens file in preview.
2. Single click on file 2 replaces file 1 in preview.
3. Double click opens file in "kept open" tab
4. Double clicking when file is open in preview turns tab to "kept open" tab

PR #251, opened 2023-04-21
Fixed inputTextBox closing with empty client ID

tsej:
For SAS Viya 2022.11 and later, you can leave Client ID empty, but in interactive mode, pressing Enter with an empty box will interrupt editing instead of saving.

Zrzut ekranu 2023-04-21 144944

### DIFF
--- a/client/src/components/profile.ts
+++ b/client/src/components/profile.ts
@@ -426,9 +426,6 @@ export class ProfileConfig {
         ProfilePromptType.ClientId,
         profileClone.clientId
       );
-      if (profileClone.clientId === "") {
-        delete profileClone.clientId;
-      }
 
       if (profileClone.clientId) {
         profileClone.clientSecret = await createInputTextBox(
@@ -602,7 +599,7 @@ const input: ProfilePromptInput = {
   [ProfilePromptType.ClientId]: {
     title: "Client ID",
     placeholder: "Enter a client ID",
-    description: "Enter the registered client ID. An example is myapp.client.",
+    description: "Enter the registered client ID. An example is myapp.client. For SAS Viya 2022.11 and later, you can leave Client ID empty and simply press Enter",
   },
   [ProfilePromptType.ClientSecret]: {
     title: "Client Secret",


client/src/components/profile.ts
```diff
@@ -602,7 +599,8 @@ const input: ProfilePromptInput = {
   [ProfilePromptType.ClientId]: {
     title: "Client ID",
     placeholder: "Enter a client ID",
-    description: "Enter the registered client ID. An example is myapp.client.",
+    description:
```
scottdover, line 602: Hey @tsej. I believe client id info has been covered in https://github.com/sassoftware/vscode-sas-extension#profile-anatomy-viya. 

That said, it isn't immediately apparent what you should enter for client id without looking at the readme.

client/src/components/profile.ts
```diff
@@ -602,7 +599,8 @@ const input: ProfilePromptInput = {
...
+      "Enter the registered client ID. An example is myapp.client. For SAS Viya 2022.11 and later, you can leave Client ID empty and simply press Enter",
```
scottdover, line 603: If we were to make a change, having something like this would be preferable...
```suggestion
      "Enter the registered client ID. An example is myapp.client. For SAS Viya 2022.11 and later, leave client ID empty.",
```
(since `Press 'Enter'...` gets appended to the selection anyway).

@skyeaussie Do we think this additional context is worth having in the profile prompt?

client/src/components/profile.ts
```diff
@@ -426,9 +426,6 @@ export class ProfileConfig {
         ProfilePromptType.ClientId,
         profileClone.clientId
       );
-      if (profileClone.clientId === "") {
```
scottdover, line 429: At the moment, we prevent writing config details that are the "default" case (in this case, an empty string). We do something similar for compute context (we only write it to a profile _if_ it's different than the default). 

For now, I wouldn't include this change. Some additional discussion is needed to know whether we should:
 - only write non-default values to a config
 - write all values to a config

In any case, what we do should be consistent for all properties with a default value.

client/src/components/profile.ts
tsej: @scottdover here https://github.com/sassoftware/vscode-sas-extension/blob/main/README.md#add-new-sas-profile means that you can leave the prompt empty, but if you do, editing will be aborted (_delete profileClone.clientId;_). 

https://github.com/sassoftware/vscode-sas-extension/blob/a4f53667babf06666fabc708646223e7a241a1a8/client/src/components/profile.ts#L425-L432

scottdover: Interesting. My understanding is that `delete profileClone.clientId` only suppresses clientId from being written to user settings

tsej: @scottdover I don't know how it works. I've tried in several VS and each time it breaks the edit/add process if I try to proceed without specifying clientId.

scnwwu: @tsej, what did you mean by interrupted? If you leave clientid empty and press Enter, the profile creation process will just complete. Default client id doesn't need to be written to your profile. Do you have any error connecting with the created profile?

scottdover:
Closing this pull request as there hasn't been any activity in several months. Feel free to [report a bug](https://github.com/sassoftware/vscode-sas-extension/issues/new?assignees=clangsmith%2C+2TomLi&labels=bug&projects=&template=bug_report.md&title=) if there are any further outstanding issues that need to be resolved.

PR #296, opened 2023-05-16
fix: error when run code

### LINKED ISSUE #295: Error when run code
**Describe the bug**
On current main branch, get below error when trying to run SAS code.
`Cannot read properties of undefined (reading 'serverId')`

**Steps to reproduce**

1. Install the extension built from `main` branch.
2. Open a sas program
3. Click the running man icon

**Expected behavior**
Run as usual

scnwwu:
**Summary**
Fix #295

**Testing**
Test case in #295

### DIFF
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -221,7 +221,8 @@ async function close() {
   }
 }
 
-export function getSession(config: Config): Session {
+export function getSession(c: Config): Session {
+  config = c;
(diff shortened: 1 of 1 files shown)


scnwwu:
> Previously we were cloning it via `config = { ...c }`. Do we _need_ to do that?

not needed now. The cloning is done when getting profile detail.

Zhirong2022 (approved):

Issue #187, opened 2023-03-22
Need a way to cancel a program submission


Becky-Williamson:
If a user submits a program and for some reason wants to cancel the submission, there is no way for them to do that.  It would be nice to have a Cancel Submission button somewhere.

PR #321, opened 2023-06-05
feat: support cancelling running job

scnwwu:
**Summary**
Support #187
Currently only for Rest. (I guess SSH can't really cancel like Rest. It can only close session.)

**Testing**
Click the `Cancel` button on the `SAS code running...` progress notification.

### DIFF
--- a/client/src/commands/run.ts
+++ b/client/src/commands/run.ts
@@ -128,9 +128,14 @@ async function runCode(selected?: boolean) {
     {
       location: ProgressLocation.Notification,
       title: "SAS code running...",
+      cancellable: typeof session.cancel === "function",
     },
-    () =>
-      session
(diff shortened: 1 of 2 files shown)


client/src/commands/run.ts
```diff
@@ -128,9 +128,14 @@ async function runCode(selected?: boolean) {
     {
       location: ProgressLocation.Notification,
       title: "SAS code running...",
+      cancellable: typeof session.cancel === "function",
     },
-    () =>
-      session
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
scottdover, line 133: Do we _need_ this line? I removed it from your commit and things seem to build without complaint

scnwwu: Good catch. Thank you!

Zhirong2022 (approved):

PR #339, opened 2023-06-13
feat: SAS Options support for rest connection profiles

smorrisj:
**Summary**
Promotes sasOptions to a new BaseProfile interface so that all Connection Providers have access to it. Rest Connection Provider inspects new sasOptions profile to set them on the compute session api request if present.

**Testing**

1. Create a new rest connection profile with the following sasOptions: ```["-PAGESIZE=MAX"]```
Example:
```
"rest-conn": {
        "connectionType": "rest",
        "endpoint": "[...]",
        "reconnect": "true",
        "sasOptions": [
          "-PAGESIZE MAX"
        ]
      },
```
2. Run the following test code:
```proc print data=sashelp.cars; run;```
Should get 428 observations. Should see the Result tab with ODS results.
3. Run the following test code:
```proc options option=PAGESIZE value; run;```

Should get the following log output:
```
1    ods html5;
NOTE: Writing HTML5 Body file: sashtml.htm
2    proc options option=PAGESIZE value; run;

    SAS (r) Proprietary Software Release V.04.00M0D06132023

Option Value Information For SAS Option PAGESIZE
    Value: 32767
    Scope: SAS Session
    How option value set: SAS Session Startup Command Line

NOTE: PROCEDURE OPTIONS used (Total process time):
      real time           0.00 seconds
      cpu time            0.00 seconds
      

3    
4    ;run;quit;ods html5 close;
```
4. Close the connection. 
5. Change the sasOptions connection profile property to ```["-PAGESIZE MAX"]```.
6. Repeat steps 1-3 and verify that the same log output is observed.

### DIFF
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -21,6 +21,7 @@ export interface Config {
   context?: string;
   serverId?: string;
   reconnect?: boolean;
+  sasOptions?: string[];
 }
 
 let config: Config;
@@ -78,6 +79,12 @@ async function reconnectComputeSession(): Promise<ComputeSession> {
 
 async function setup(): Promise<void> {
   const apiConfig = getApiConfig();
+  let formattedOpts: string[] = [];
+
+  if (config.sasOptions) {
+    formattedOpts = formatSASOptions();
+  }
+
   if (!config.serverId) {
     const session = await authentication.getSession(SASAuthProvider.id, [], {
       createIfNone: true,
@@ -117,6 +124,7 @@ async function setup(): Promise<void> {
   //Start a new session
   if (config.serverId) {
     const server1 = new ComputeServer(config.serverId);
+    server1._options = formattedOpts;
     computeSession = await server1.getSession();
 
     //Maybe wait for session to be initialized?
@@ -135,7 +143,10 @@ async function setup(): Promise<void> {
 
     const sess = (
       await contextsApi.createSession(
-        { contextId: context.id },
+        {
+          contextId: context.id,
+          sessionRequest: { environment: { options: [...formattedOpts] } },
+        },
         { headers: { "accept-language": locale } }
       )
     ).data;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by scottdover — APPROVED** [review]

**Comment by scottdover** [comment]
Oops, just noticed this was a draft. Looks good so far :)

**Comment by smorrisj** [comment]
> Oops, just noticed this was a draft. Looks good so far :)

No worries. I had set it to non-draft by mistake. It's ready now. Would also like @JosephDHenry to take a look if that's alright since this will dovetail into some other items down the road.

File: client/src/connection/rest/index.ts [hunk 1]
```diff
@@ -146,6 +157,28 @@ async function setup(): Promise<void> {
...
+ * ["-PAGESIZE=MAX"] -> ["-PAGESIZE MAX"]
+ * ["-NOTERMINAL"] -> ["-NOTERMINAL"]
+ * ```
+ *
+ * @returns formatted SAS Options
+ */
+function formatSASOptions() {
+  const formattedOpts = config.sasOptions.map((opt) => {
+    let formatted = opt;
+    formatted = formatted.replace(/-/gi, "");
```
scnwwu, line 176: Hi @smorrisj, this looks confusing to me. Shouldn't it be
```
formatted = formatted.replace(/=/gi, " ");
```

smorrisj: > Hi @smorrisj, this looks confusing to me. Shouldn't it be
> 
> ```
> formatted = formatted.replace(/=/gi, " ");
> ```

We want to turn:
-PAGESIZE max
into:
PAGESIZE max

Based on this, its appropriate to replace the dash with empty?

smorrisj: > Hi @smorrisj, this looks confusing to me. Shouldn't it be
> 
> ```
> formatted = formatted.replace(/=/gi, " ");
> ```

This sounds correct. @Becky-Williamson is it possible to log a bug for this? We'll need to fix it on a separate PR since this has been closed.

PR #355, opened 2023-06-19
feat: autoexec support for viya connection

smorrisj:
**Summary**
Adds autoexec support to the rest connection profle. AutoExec can be setup in multiple ways, by referencing one or more external files that are readable by vscode:

``` json
     "rest-profile-1": {
        "connectionType": "rest",
        "endpoint": <...omitted...>,
        "autoExec": [{
          "type": "file", 
          "filePath":"/file/path/one.sas"
        },{
          "type":"file",
          "filePath": "/file/path/two.sas"}
         ]
      }
```

By directly embedding SAS lines into the connection profile:
``` json
     "rest-profile": {
        "connectionType": "rest",
        "endpoint": <...omitted...>,
        "autoExec": [{
          "type": "line", 
          "line":"libname _all_ list;"
        },{
          "type":"line",
          "line": "%put _all_;"}
         ]
      }
```
Or by using a mix of the two approaches:

``` json
     "rest-profile": {
        "connectionType": "rest",
        "endpoint": <...omitted...>,
        "autoExec": [{
          "type": "line", 
          "line":"libname _all_ list;"
        },{
          "type":"file",
          "filePath": "/file/path/two.sas"}
         ]
      }
```

**Testing**
1.) For each example profile above:
2.) Create a connection profile with the specific autoexec form.
3.) Run the following test code:
``` sas
proc print data=sashelp.cars;
run;
```
4.) Check the SAS Log Window in VSCode. Ensure that the session log lines are output including relevant autoexec output.

TODO:
- [x] Show log earlier in execution cycle so that the user can see startup log and autoexec log

### DIFF
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -13,7 +13,7 @@ import { ComputeState, getApiConfig } from "./common";
 import { ComputeJob } from "./job";
 import { ComputeServer } from "./server";
 import { ComputeSession } from "./session";
-import { BaseSession, Session } from "../session";
+import { Session } from "../session";
 
 let sessionInstance: RestSession;
 
@@ -26,7 +26,7 @@ export interface Config extends BaseConfig {
   reconnect?: boolean;
 }
 
-class RestSession extends BaseSession implements Session {
+class RestSession extends Session {
   private _config: Config;
   private _computeSession: ComputeSession | undefined;
 
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { authentication } from "vscode";
-import { BaseConfig, RunResult, Session } from "..";
+import { BaseConfig, RunResult } from "..";
 import { SASAuthProvider } from "../../components/AuthProvider";
 import {
   getContextValue,
@@ -13,6 +13,7 @@ import { ComputeState, getApiConfig } from "./common";
 import { ComputeJob } from "./job";
 import { ComputeServer } from "./server";
 import { ComputeSession } from "./session";
+import { BaseSession, Session } from "../session";
 
 export interface Config extends BaseConfig {
   endpoint: string;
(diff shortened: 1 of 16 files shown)


client/src/connection/rest/server.ts
```diff
@@ -86,7 +91,7 @@ export class ComputeServer extends Compute {
       attributes: {},
       environment: {
         options: [...DEFAULT_COMPUTE_OPTS, ...this._options],
-        autoExecLines: [],
+        autoExecLines: [...this._autoExecLines],
```
scottdover, line 94: Do you need to create a copy, here?

package.nls.json
```diff
@@ -1,4 +1,9 @@
 {
+  "configuration.SAS.connectionProfiles.profiles.autoExec": "SAS Connection AutoExec",
+  "configuration.SAS.connectionProfiles.profiles.autoExec.lines": "SAS Connection AutoExec Lines",
+  "configuration.SAS.connectionProfiles.profiles.autoExec.filePaths": "SAS Connection AutoExec File paths",
+  "configuration.SAS.connectionProfiles.profiles.autoExec.type": "SAS Connection AutoExec type",
```
scottdover, line 5: Nitpick. Will you alphabetize these?

smorrisj:
> Hey @smorrisj . I noticed in your config you have `"endpoint": ["..."]` . Are there plans to make endpoint an array, or this a typo?

It's not a typo. Thats just to denote that I'm not listing out an endpoint to avoid divulging sensitive information.

smorrisj:
> A few initial comments. I think we need to support a combination of raw files and static lines, not just 1 or the other. it might also be nice to support multiple files? IDK though, that might be overkill, but might be nice.
> 
> Also should we save the log from the autoexec and show it somewhere?

With what's currently implemented, the files option can support one or many files.  In the combination approach, what is the expected order that the sas code would get applied, files first or lines?

smorrisj:
> Also should we save the log from the autoexec and show it somewhere?

I was thinking about this. An issue in the current design is that it only wires up the logfn handler when running user code. We could perhaps set ECHOAUTO if autoexec profile options exist to see if that would get us what we need?

smorrisj:
> > Hey @smorrisj . I noticed in your config you have `"endpoint": ["..."]` . Are there plans to make endpoint an array, or this a typo?
> 
> It's not a typo. Thats just to denote that I'm not listing out an endpoint to avoid divulging sensitive information.

Updated to better show intent.

scottdover:
> > Hey @smorrisj . I noticed in your config you have `"endpoint": ["..."]` . Are there plans to make endpoint an array, or this a typo?
> 
> It's not a typo. Thats just to denote that I'm not listing out an endpoint to avoid divulging sensitive information.

Ah, got it. The square brackets just threw me a little

scottdover (approved):

JosephDHenry:
> 

I would think it would just be the order it showed up in the config.

client/src/connection/session.ts
```diff
@@ -0,0 +1,20 @@
...
+  _onLogFn: OnLogFn;
+  public set onLogFn(value: (logs: LogLine[]) => void) {
+    this._onLogFn = value;
+  }
+}
+
+export interface Session {
+  onLogFn: (logs: LogLine[]) => void;
+  setup(): Promise<void>;
+  run(code: string, onLog?: (logs: LogLine[]) => void): Promise<RunResult>;
```
scnwwu, line 16: is `onLog` parameter still needed here?

client/src/connection/session.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright © 2022-2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { LogLine, OnLogFn, RunResult } from ".";
+
+export class BaseSession {
+  _onLogFn: OnLogFn;
+  public set onLogFn(value: (logs: LogLine[]) => void) {
```
scnwwu, line 8: nit: as `OnLogFn` type already defined, it would be better to use it for all type annotations everywhere.

client/src/connection/session.ts
```diff
@@ -0,0 +1,20 @@
...
+    this._onLogFn = value;
+  }
+}
+
+export interface Session {
+  onLogFn: (logs: LogLine[]) => void;
```
scnwwu, line 14: We can define Session as a class with needed base implementation and abstract methods. A class can be used as an interface. It can avoid duplicated `onLogFn` declaration.

client/src/connection/rest/index.ts
```diff
@@ -13,263 +13,271 @@ import { ComputeState, getApiConfig } from "./common";
...
-  return {
-    setup,
-    run,
-    cancel,
-    close,
-    sessionId,
-  };
+  getApiConfig().basePath = c.endpoint + "/compute";
+
+  return new RestSession(c);
```
scnwwu, line 282: `getSession` can be called many times from anywhere. It would be better to make it singleton. Otherwise different components may operate on different instances and may increase complexity.

client/src/connection/session.ts
smorrisj: Removed. See Updated.

client/src/connection/session.ts
smorrisj: Implemented. See Updated.

client/src/connection/session.ts
smorrisj: A cleaner approach for sure. See Updated.

client/src/connection/rest/index.ts
smorrisj: Good catch, I'd done this on the ssh support but forgot to implement it here. See updated.

client/src/connection/rest/server.ts
smorrisj: Fixed, see updated.

package.nls.json
smorrisj: Fixed, see updated.

scnwwu (approved):
Looks good to me. Thanks.

Issue #189, opened 2023-03-24
VSC SAS Extension enhancement to support pseudo locale to cover I18N testing 


Zhirong2022:
So far, it has no appropriate way to check I18N support of SAS Extension.  We need some enhancement for it if possible.

PR #370, opened 2023-07-03
feat: l10n infrastructure

scnwwu:
**Summary**
Support #189 with [vscode-l10n](https://github.com/microsoft/vscode-l10n).
Added a `pseudoLocalize` script.
Note that it only applies to `client` directory. `server` is already localized with SAS tooling.

**Testing**

1. Checkout branch code
2. `npm install`, `npm run pseudoLocalize`, `vsce package`
3. Install the vsix
4. Install [Pseudo Language Language Pack](https://marketplace.visualstudio.com/items?itemName=MS-CEINTL.vscode-language-pack-qps-ploc) extension
5. Run `Configure Display Language` command and switch to the Pseudo Language. (It requires restart)
6. See UI text provided by the extension are pseudo translated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ node_modules
 .vscode-test*
 *.vsix
 *.node
+*.qps-ploc.json
+l10n/bundle.l10n.json
--- a/client/src/components/ContentNavigator/const.ts
+++ b/client/src/components/ContentNavigator/const.ts
@@ -1,6 +1,8 @@
 // Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { l10n } from "vscode";
+
 const CONTENT_FOLDER_ID = "CONTENT_FOLDER_ID";
 export const ROOT_FOLDER_TYPE = "RootFolder";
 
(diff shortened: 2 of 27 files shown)


client/src/components/ContentNavigator/const.ts
```diff
@@ -43,36 +45,43 @@ export const FOLDER_TYPES = [
...
+  RemoveFromFavoritesError: l10n.t(
+    "The item could not be removed from My Favorites."
+  ),
+  RenameError: l10n.t('Unable to rename "{oldName}" to "{newName}".'),
+  RenameFileTitle: l10n.t("Rename File"),
+  RenameFolderTitle: l10n.t("Rename Folder"),
+  RenamePrompt: l10n.t("Enter a new name."),
+  RenameUnsavedFileError: l10n.t(
+    "You must save your file before you can rename it."
+  ),
```
scottdover, line 86: It looks like these should probably just point to strings in `package.nls.json` since they are called with `l10n.t` in other locations

client/src/components/LibraryNavigator/LibraryModel.ts
```diff
@@ -27,7 +26,7 @@ class LibraryModel {
     await window.withProgress(
       {
         location: ProgressLocation.Notification,
-        title: "Connecting to SAS session...",
+        title: l10n.t("Connecting to SAS session..."),
```
scottdover, line 29: These don't follow our typical format in `package.nls.json`. Do we need to update all of these strings to conform to that format?

scnwwu:
> * Things called with `l10n.t` should conform to the format of similar items specified in package.nls.json
> * `const Messages = ...` should probably only reference keys in `package.nls.json`. Otherwise, we end up doing something like `l10n.t(l10n.t('Message'))`

Hi @scottdover,
The `package.nls.json` file is only for strings in `package.json`, not source code.
`l10n.t` directly wrap English message, not "key". `@vscode/l10n-dev` script scans the source code for strings that wrapped by `l10n.t` and generate pseudo translations into `l10n/bundle.l10n.qps-ploc.json` file. Please refer to [vscode-l10n](https://github.com/microsoft/vscode-l10n) documentation for details.
So wrapping `l10n.t` around `Messages` is needed for static scanning. Then another `l10n.t` is used for argument replacement. So it has to be nested `l10n.t` call. It's not a problem.

.gitignore
```diff
@@ -3,4 +3,6 @@ dist
 node_modules
 .vscode-test*
 *.vsix
-*.node
\ No newline at end of file
+*.node
+*.qps-ploc.json
+l10n/bundle.l10n.json
```
scottdover, line 8: Couple things. 
- Why do we ignore l10/bundle.l10n.json? 
- Will `.qps-ploc.json` _need_ to be ignored (i.e. will it be generated on the main branch at all, or is it just something you had setup for testing?)

scnwwu: Those files are auto-generated by the `pseudoLocalize` script by static scanning the source code. So they don't need to be committed. Just generate them with the source code you need.
The pseudo file will be generated when you test pseudo locale each time so *need* to be ignored.

scottdover: So I guess what I don't understand is how will we commit translations to the repository. Is there a point where we'll have things in `l10n/` like, `bundle.l10n.zh-cn.json` (or something similar)?

scnwwu: Yes, real translations need to be committed, like `l10n/bundle.l10n.zh-cn.json`. They won't be ignored.

scnwwu:
Hi @scottdover, any additional questions or concerns or do you think it can be approved? Thanks.

scottdover (approved):

liu-lihua:
@Zhirong2022 is out of office this week, so I work this PR testing for her.

liu-lihua:
AC test criteria:
Make sure UI text provided by the extension are pseudo translated in below elements:
1. Contextmenu 
2. Popup message 
3. Tooltip 
4. SAS Setting syntax help 
5. Command Palette
6. UI string on SAS content pane 
7. Result tab page title
8. Verify 1 to 7 in OS: Windows, Mac and Linux

liu-lihua:
I tested SAS Extension with the the branch code build. All the AC test criteria has been tested and accepted. No regression issue was found. So testing has been completed.

liu-lihua:
I could see the pseudo langauge support in the latest main branch build. So I verify it accordingly.

PR #378, opened 2023-07-03
fix: check that _onLogFn exists before calling in printSessionLog

smorrisj:
**Summary**
Checks that onLogFn has been set before printing out the session log for viya connections to prevent null reference errors.

**Testing**
Repro steps from the issue.

### DIFF
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -121,8 +121,7 @@ class RestSession extends Session {
         )
       ).data;
       this._computeSession = ComputeSession.fromInterface(sess);
-      const sessionLog = this.printSessionLog(this._computeSession);
-      await sessionLog;
+      await this.printSessionLog(this._computeSession);
     }
 
     //Save the current sessionId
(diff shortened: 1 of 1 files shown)


scnwwu (approved):
As I said in the issue, if user sees the library view first (which is a normal case), they can't see the session log when run code later. Because we only get session log on session creation.
I'm good with it if you guys are good. So just give my vote.

smorrisj:
> As I said in the issue, if user sees the library view first (which is a normal case), they can't see the session log when run code later. Because we only get session log on session creation. I'm good with it if you guys are good. So just give my vote.

Thanks @scnwwu. I have that issue fixed in a separate remote that I'm working in right now. Will issue a PR for that later on today.

Zhirong2022:
There is no error with library data loading.

Zhirong2022 (approved):

liu-lihua:
I tried today's main branch build and this issue is fixed. So I verify this PR accordingly.

Issue #369, opened 2023-07-03
Error 'this._onLogFn is not a function' is thrown with library data loading 


Zhirong2022:
**Describe the bug**
Install the SAS extension base on the latest Main branch. Sign in and try to load SAS contents, it failed to load data under Libraries with error 'this._onLogFn is not a function'.

**Steps to reproduce**

1.Sign in with a valid profile

**Expected behavior**
It can load Libraries data successfully without error

**Screenshots**
![image](https://github.com/sassoftware/vscode-sas-extension/assets/121005702/34a3a097-29d3-458a-90b7-382bf358b9b4)

**Environment (please complete the following information):**
Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.0.0]

scnwwu:
Regression introduced by #355
Library view requires a session. But it does not need to view any log. We can check if no _LogFn set, don't retrieve session log.
But if user then run code, as the session already exists, no session log will be shown.
Defer to @smorrisj

PR #380, opened 2023-07-05
refactor: Session Log Streaming

smorrisj:
**Summary**
Check the session job status before streaming the log to avoid various timing problems. Additionally, LibraryModel is currently creating a session. Modify LibraryModel to attach the log handler when LibraryModel creates a session in order to see the session log when LibraryModel invokes session setup. Fixes current issues where if the user opens the Library Panel first, before executing code, then the session log will not be output to the channel. Additionally, when running the actual sas file, it would also not be seen there, which is not desired behavior.

**Testing**
Test case in #369

### DIFF
--- a/client/src/components/LibraryNavigator/LibraryModel.ts
+++ b/client/src/components/LibraryNavigator/LibraryModel.ts
@@ -7,6 +7,7 @@ import { ProgressLocation, window } from "vscode";
 import { getSession } from "../../connection";
 import { DataAccessApi } from "../../connection/rest/api/compute";
 import { getApiConfig } from "../../connection/rest/common";
+import { LogFn as LogChannelFn } from "../LogChannel";
 import PaginatedResultSet from "./PaginatedResultSet";
 import { DefaultRecordLimit, Messages } from "./const";
 import { LibraryItem, LibraryItemType, TableData } from "./types";
@@ -23,6 +24,7 @@ class LibraryModel {
 
   public async connect(): Promise<void> {
     const session = getSession();
+    session.onLogFn = LogChannelFn;
 
     await window.withProgress(
       {
--- a/client/src/components/LogChannel.ts
+++ b/client/src/components/LogChannel.ts
@@ -1,0 +1,14 @@
+import { appendLog } from "./LogViewer";
+import { OutputChannel, window } from "vscode";
+
+let outputChannel: OutputChannel;
+export const LogFn = (logs) => {
+  if (!outputChannel) {
+    outputChannel = window.createOutputChannel("SAS Log", "sas-log");
+  }
+  outputChannel.show();
+  for (const line of logs) {
+    appendLog(line.type);
+    outputChannel.appendLine(line.line);
+  }
+};
--- a/client/src/components/LogChannel.ts
+++ b/client/src/components/LogChannel.ts
@@ -1,5 +1,6 @@
 // Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
 import { appendLog } from "./LogViewer";
 import { OutputChannel, window } from "vscode";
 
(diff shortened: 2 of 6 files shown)


client/src/connection/rest/session.ts
```diff
@@ -177,8 +177,45 @@ export class ComputeSession extends Compute {
...
+    const states = [
+      "done",
+      "canceled",
+      "error",
+      "warning",
+      "completed",
+      "idle",
+    ];
+
+    while (states.indexOf(state) === -1) {
```
scottdover, line 193: Is there a chance this ends up an infinite loop?

smorrisj: if "idle" were to be removed from the list of states then absolutely. Once the session goes idle, it would cycle there infinitely. Basically the logic we're trying to convey, is that if the session status is pending or not yet "complete", go ahead and grab what log we have and yield it out. In the case where the session spins up quickly on faster connections and becomes "idle" then the logic simply falls down into the block below where we yield out whatever is there and terminate when there are no more pages to yield.

client/src/components/LogChannel.ts
```diff
@@ -1,3 +1,5 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
smorrisj: fixed

scottdover (approved):

client/src/components/LibraryNavigator/LibraryModel.ts
```diff
@@ -23,6 +24,7 @@ class LibraryModel {
 
   public async connect(): Promise<void> {
     const session = getSession();
+    session.onLogFn = LogChannelFn;
```
scnwwu, line 27: The `LogChannelFn` will always **show** the output channel.
Do user like the output channel popup if they're not running code?

client/src/connection/rest/session.ts
```diff
@@ -190,7 +227,7 @@ export class ComputeSession extends Compute {
     //To clear out the log, we yeild all lines until there is not "next" link
     do {
       if (resp.status === 200) {
-        nextLink = resp.data.links?.find((link) => link.rel === "next");
+        nextLink = resp.links?.find((link) => link.rel === "next");
```
scnwwu, line 230: is it intended? Looks like `data` makes sense?

client/src/connection/rest/index.ts
```diff
@@ -289,9 +289,11 @@ class RestSession extends Session {
    * @param session a session id to print logs for.
    */
   private printSessionLog = async (session: ComputeSession) => {
-    const logs = await session.getLogStream();
-    for await (const log of logs) {
-      this._onLogFn(log);
+    if (this._onLogFn) {
```
scnwwu, line 292: I'm thinking it would be good if we can make this kind of situation more safe.
If we would like to always show log, how about make the `LogChannelFn` the default one for base Session?
If we don't want a default implementation, how about declare the type of the `_onLogFn` to be `OnLogFn | undefined` so that every usage knows that it could be undefined.

smorrisj: Good point, once we know what a desired UX is I can either update the type or move it to be base behavior.

smorrisj: @scnwwu if its agreeable I'd like to implement option#2. If we have further feature areas in the future that would require it then we could refactor to move it to the default implementation?

client/src/components/LibraryNavigator/LibraryModel.ts
smorrisj: Confirmed that the UX of showing the log is acceptable.

scnwwu (approved):
Good to me. Thanks.

liu-lihua:
@Zhirong2022 is out of office this week, so I work on the testing in this PR for her now.

liu-lihua:
I tested the code in branch "bug/session-log-timing" bug/session-log-timing. I could see log in output window if I open libraries panel before running a real SAS file in VSC. Please see the attached "fixed.png".
fixed

scottdover (approved):

liu-lihua:
I verified today's main branch build and this issue is gone. So I verify this PR accordingly.

Issue #346, opened 2023-06-16
Can't close session or end SAS using local COM


ShuguangSun:
**Describe the bug**
SAS 9.4M7 from SAS OA. I just tried the new feature of local COM. VScode show "SAS code running" and I can't run another SAS code (no command can be found "SAS: run all code").  The SAS LOG is shown:
```
ods html5 path="C:\Users\xxxx\AppData\Local\Temp\2\SAS Temporary Files\_TD10028_WIN-EAN3VMK680E_\Prc2";
NOTE: HTML5 Body: sashtml.htm
proc print data=sashelp.cars;
run;
```
It can be found in the console:
```
[SAS.sas-lsp]ENOENT: no such file or directory, open 'C:\Users\xxxxx\AppData\Local\Temp\2\SAS Temporary Files\_TD42672_WIN-EAN3VMK680E_\Prc2\undefined.htm'
```

**Environment (please complete the following information):**
Client OS: Windows 2019
Extension version: v1.0.0
SAS version: SAS 9.4M7 from OA

**Additional context**
SASpy works fine.

ShuguangSun:
I found it is because there is no space after 'HTML5 Body' in the NOTE if I use a zh_CN local. So that the original regex can't find the file name and make 'undefined'.

For Example
```
NOTE: ����д�� HTML5 Body�����壩�ļ�: sashtml.htm
```

ShuguangSun:
But it still an issue. The pr https://github.com/sassoftware/vscode-sas-extension/pull/348 fix the local file name.

However, if I use a remote COM, e.g., I have server machine with SAS OA with ITC and I use VScode in a local PC. I can add the host ip in the profile and it can connect to the server and submit the code. It can't find the html5 file because the file is on the server machine which can't be access using the local file name, e.g., `C:\Windows\TEMP\SAS Temporary Files\_xxx_xxx_\Prc2\sashtml.htm`. It still block vscode to close the SAS session.

scnwwu:
The additional issue is tracked in #349
Closing this one.

PR #381, opened 2023-07-07
 fix: regex to match html5 file name in local COM

ShuguangSun:
**Summary**
As issue https://github.com/sassoftware/vscode-sas-extension/issues/346, if I use a local of zh_CN, then there is no space after 'HTML5 Body', the regex can't match this line and can't find the output html filename.

For example,
```
NOTE: ����д�� HTML5 Body�����壩�ļ�: sashtml.htm
```

Sorry I haven't learned to how to test it.

**Testing**
I test it in 
- SAS 9.4M7 from OA
- vscode 1.79.2
- using local COM

### DIFF
--- a/client/src/connection/com/index.ts
+++ b/client/src/connection/com/index.ts
@@ -226,7 +226,7 @@ do {
           this.fetchResults();
         } else {
           this._html5FileName =
-            line.match(/NOTE: .+ HTML5.* Body .+: (.+)\.htm/)?.[1] ??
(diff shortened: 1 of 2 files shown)


scnwwu (approved):

smorrisj (approved):
LGTM. We will track tests as a separate issue since it may require a moderate refactor.

Zhirong2022:
It can run code with zh session.

scnwwu:
@ShuguangSun, thanks for your contribution!

Issue #363, opened 2023-06-26
It is not necessarily to indent in certain case with code typing


Zhirong2022:
**Describe the bug**
Install the SAS extension base on the official release. Typing some SAS code, it can make some improvements to avoid indenting other than disable it. 

**Steps to reproduce**

1.Create a new SAS file
2.Type some code below

proc contents data=x; run;

3.Type Enter

**Expected behavior**
No code indentation in such scenario

**Screenshots**
![image](https://github.com/sassoftware/vscode-sas-extension/assets/121005702/3cd2e535-b6b7-4584-ad78-8f030b7dccfb)

**Environment (please complete the following information):**
Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.0.0]

jingjiajie:
I added a pull request to resolve this issue. https://github.com/sassoftware/vscode-sas-extension/pull/396

PR #396, opened 2023-07-14
fix: indent bugfix (#363)

jingjiajie:
**Summary**
Issue #363 
&emsp;Refined the Regex to make the indent more accurate. 
&emsp;The Regex is a powerful tool, but still has many restrictions. I can't make it meet 100% senarios, but I think it's definitely enough for everyday use. For absolute accurate, I suggest to do it programmatically rather than using Regex.
Below are some correct and wrong cases.

**Testing**
**CORRECT:**
data a.b;run;
stmt;

data a.b;run; proc/\*123\*/ c;
&emsp;stmt;

proc contents data=x; run;
stmt;

data a.b;
&emsp;stmt;
run;

/\*comment\*/ data a.b;
&emsp;stmt;
run;

*comment; data a.b; /\*comment\*/
&emsp;stmt;
run;

*comment; data a.b; *comment;
&emsp;stmt;
run;

proc c; stmt1; run; data c.d; /\*comment \*/
&emsp;stmt2;
run;

data a.b; /\*comment\*/ \*/
&emsp;stmt;

**INCORRECT:**
/\* Should indent
\* This case can be resolved by copying the comment Regex part into the data/proc step Regex, but could make 
the Regex too complecated
\*/
data /\*;abc\*/ a;
stmt;

/\* Should indent\*/
data a.b; set c;
stmt;

/\* Should indent\*/
data a.b;;
stmt;

### DIFF
--- a/language-configuration.json
+++ b/language-configuration.json
@@ -24,7 +24,7 @@
     ["\"", "\""]
   ],
   "indentationRules": {
-    "increaseIndentPattern": "^\\s*(data|proc|%macro)\\b.*;$",
+    "increaseIndentPattern": "(data|proc|%macro)\\b[^;]*;(\\s|/\\*.*\\*/|\\*[^;]*;)*$",
     "decreaseIndentPattern": "^\\s*(run;|quit;|%mend;)$"
   },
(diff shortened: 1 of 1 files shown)


jingjiajie:
Use https://github.com/sassoftware/vscode-sas-extension/pull/451 to instead of this one.

Issue #393, opened 2023-07-13
Implement Cancel Function for SAS9.4(local-COM)


Zhirong2022:
**Describe the bug**
Install the SAS extension base on the latest Main branch. Switch to a profile with connection type 'SAS9.4(local-COM)',try to run some code, it will show Cancel button on 'SAS code running...'. However it does not support Cancel action with SAS9.4(local-COM) connection. 

**Steps to reproduce**

1. Create a new SAS profile with connection type SAS9.4(local-COM) or switch to a profile with connection type SAS9.4(local-COM)
2. Run some code below
	proc print data=sashelp.class; 
	run;

**Expected behavior**
No Cancel button on 'SAS code running...', since the feature is not available now. 

**Screenshots**
![image](https://github.com/sassoftware/vscode-sas-extension/assets/121005702/14d559d3-ab87-4333-9361-f5468f9de168)

**Environment (please complete the following information):**
Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.0.0]

scnwwu:
It's not only against COM but also SSH.
#355 added empty `cancel` implementation to SSH and COM.
It should not.
The idea is to not show cancel button if a session does not support cancel (no such function).
See https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/commands/run.ts#L141
I heard that @smorrisj will be implementing cancel for COM so will defer to him. If we don't support cancel on SSH, the function should be removed.

scnwwu:
Cancel supported for local 94 with #592

PR #406, opened 2023-07-21
fix: should not show cancel button if not supported

scnwwu:
**Summary**
As my https://github.com/sassoftware/vscode-sas-extension/issues/393#issuecomment-1633781315
Let me go ahead remove them so that we're in a good state to release.
Anytime it can be implemented and will just work fine.

**Testing**
See no cancel button while running against COM and SSH because we don't support it yet.

### DIFF
--- a/client/src/connection/com/index.ts
+++ b/client/src/connection/com/index.ts
@@ -155,13 +155,6 @@ export class COMSession extends Session {
   };
 
   /**
-   * Cancel a running code execution.
-   */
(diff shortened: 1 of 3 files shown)


smorrisj (approved):
LGTM

Zhirong2022:
There is no cancel button if trying to run code against COM and SSH.

Zhirong2022 (approved):

PR #407, opened 2023-07-21
build: update NodeJS to current LTS

smorrisj:
**Summary**
- Updates github action to build on Node 18.x (Current LTS)
- Refactor PR check to use a nvmrc file that gives us the ability to point to a logical "lts" mnemonic instead of having to pin to explicit version, as an aid to keeping this up to date. This means that the pr check will always use the nodejs version associated with the "lts" mnemonic on the github action. 

**Testing**
Local build with Node LTS, verify github action runs successfully


smorrisj:
Although this only updates the build env. It would be good to have test verify that basic functionality still works as a precaution.

smorrisj:
The obvious tradeoff here, is that during work on a PR, we could run into issues where the LTS version updates in the midst of some other feature work, which halts progress in order to address whatever compatibility issue there might be to satisfy the PR check.  

I'd propose that these scenarios are more the exception than the rule. The benefits of having an up to date and secure LTS  build environment that's checked with automation would be far more useful IMO. There are also several ways that we can temporarily mitigate the above issue such that progress can continue forward. Given this, I think this is an acceptable way to move forward.

scottdover (approved):

scnwwu (approved):
Good to me. Thanks.

Zhirong2022:
Everything works well with NodeJS LTS.

Zhirong2022 (approved):

Zhirong2022 (approved):

PR #416, opened 2023-07-27
fix: start should not exceed count in request parameter

scnwwu:
**Summary**
Viya 3.5 appears to not allow start parameter greater than count.

**Testing**
See table showing correctly on Viya 3.5 and Viya 4


scottdover (approved):
Good catch. I fixed this in one spot, but forgot to update it in the rest

PR #430, opened 2023-08-01
feat: honor Caps Lock status when autocompletion.

scnjdl:
**Summary**
Input lowercase or uppercase keyword per user input

**Testing**
manual testing

### DIFF
--- a/server/src/sas/CompletionProvider.ts
+++ b/server/src/sas/CompletionProvider.ts
@@ -5,6 +5,7 @@
 import {
   CompletionItem,
   CompletionItemKind,
+  CompletionList,
   Hover,
   MarkupKind,
 } from "vscode-languageserver";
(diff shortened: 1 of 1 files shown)


scnjdl:
The code is updated to make the completion list refresh for each typing.

server/src/sas/CompletionProvider.ts
```diff
@@ -265,26 +280,39 @@ export class CompletionProvider {
     }
   }
 
-  getCompleteItems(position: Position): Promise<CompletionItem[] | undefined> {
+  getCompleteItems(
+    position: Position,
+  ): Promise<CompletionItem[] | CompletionList | undefined> {
     return new Promise((resolve) => {
       this._getZone(position);
+      const prefix = this._getPrefix(position);
```
scnwwu, line 288: nit: this.popupContext.prefix is already there, no need to calculate again

scnwwu (commented):
How about changing the title to be something like
follow user's typing case
as we're not really checking Caps Lock status

server/src/sas/CompletionProvider.ts
scnjdl: Done

scnwwu (approved):

liu-lihua:
Assign it to @Zhirong2022 for interactive testing since it is ready for testing now.

Zhirong2022 (approved):

PR #435, opened 2023-08-07
fix: always show running man icon

### LINKED ISSUE #433: Enable and disable "running man" icon(s) rather than showing and hiding it
The "running man" (run) icon currently only appears when a SAS program has focus. If you move focus elsewhere (ex. the OUTPUT window), the run icon is hidden from all SAS programs in view (could have multiple active in multiple/split tabgroups). Therefore, you cannot simply click the run icon again to run the program again. You have to first put focus into the program, which will cause the run icon to re-appear, and then you can click it to run.

It would be nicer if the run icon was always visible for any active SAS programs (regardless of whether they had focus), and simply disabled while any program is running. (In the future, if we can queue, then would only need to disable for actively running programs.) For example, more like this:

image

(It probably doesn't make sense to have the disconnect button there on every toolbar though. A separate issue.)

In the future, it would be nice to try to integrate into VS Code's built-in Run/Debug capability, so that it would be consistent with working with other languages in VS Code (ex. Python). Even then, we might still offer a legacy "show running man icon" option (off by default).

**Additional context**
@scnwwu mentioned this may be challenging, since there isn't a good way to get the active document in the tab group. The first time they tried to implement, could only get the focused document, so if focus was in OUTPUT window, it would submit the text in the OUTPUT window! It looks like there is a separate toolbar for each tab group and the running man icon is one of the buttons on each of those toolbars. We would need a way to get the **active** (regardless of whether it had focus) tab in the tabgroup where the running man icon was clicked. More research needed.

**Environment**
SAS version - Viya and V9

scnwwu:
**Summary**
Fix #433

When a command is invoked from a menu, VS Code will pass the URI of the triggered resource as the first parameter.
We should utilize it over `vscode.window.activeTextEditor` (which is the editor with focus, thus could be SAS log).

**Testing**

- Be able to directly click the running man icon when focus is not in the SAS code. The code submitted is correct.
- Not be able to run when it's already running (running man icon disabled)

### DIFF
--- a/package.json
+++ b/package.json
@@ -346,6 +346,7 @@
           "light": "icons/light/submitSASCode.svg",
           "dark": "icons/dark/submitSASCode.svg"
         },
+        "enablement": "!SAS.running",
         "category": "SAS"
       },
       {
@@ -355,6 +356,7 @@
           "light": "icons/light/submitSASCode.svg",
           "dark": "icons/dark/submitSASCode.svg"
         },
+        "enablement": "!SAS.running",
         "category": "SAS"
       },
       {
@@ -364,6 +366,7 @@
           "light": "icons/light/submitSASCode.svg",
           "dark": "icons/dark/submitSASCode.svg"
         },
+        "enablement": "!SAS.running",
         "category": "SAS"
       },
       {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -572,13 +575,13 @@
       ],
       "editor/title/run": [
         {
-          "when": "editorFocus && editorLangId == sas && !SAS.hideRunMenuItem",
+          "when": "editorLangId == sas && !SAS.hideRunMenuItem",
           "command": "SAS.runSelected"
```
scnjdl, line 579: replace all 'SAS.hideRunMenuItem' with SAS.running ?

scnwwu: We're disabling instead of hiding the icon for running. Hiding is still needed when execution is completely not supported (pure web).

scottdover (approved):

Zhirong2022:
The feature works as expected

Zhirong2022 (approved):

PR #436, opened 2023-08-07
chore: minor cleanup

scnwwu:
- Remove redundant `sasOptions` definition in configuration
- "react-infinite-scroller" is not used anymore with ag grid now


smorrisj (approved):
LGTM

PR #447, opened 2023-08-08
feat: convert .sasnb to .flw

sahbic:
**Summary**
Implementation of a solution to convert .sasnb files to .flw (SAS Studio Flows)
The flow is then associated with SAS Studio, it appears in SAS Studio and can be executed and modified in SAS Viya GUI.

**Testing**
I have tested this change with two examples of .sasnb files: the first with one cell and the second with 3 cells (python, sas, sql)

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -257,6 +257,14 @@ class ContentDataProvider
     return this.model.saveContentToUri(uri, new TextDecoder().decode(content));
   }
 
+  public associateFlow(
+    name: string,
+    uri: Uri,
+    parent: ContentItem,
+  ): void | Promise<void> {
+    return this.model.associateFlowObject(name, uri, parent);
+  }
+
   public async deleteResource(item: ContentItem): Promise<boolean> {
     if (!(await closeFileIfOpen(item))) {
       return false;
--- a/client/src/components/ContentNavigator/ContentModel.ts
+++ b/client/src/components/ContentNavigator/ContentModel.ts
@@ -354,6 +354,61 @@ export class ContentModel {
     }
   }
 
+  public async associateFlowObject(
+    name: string,
+    uri: Uri,
+    parent: ContentItem,
+  ): Promise<void> {
+    // get the zone with format "GMT+02:00"
+    const date = new Date(); // Create a Date object representing the current date and time
+    const timeZoneOffset = date.getTimezoneOffset();
+    // Convert the time zone offset to the desired format (e.g., "GMT+02:00")
+    const hoursOffset = Math.floor(Math.abs(timeZoneOffset) / 60);
+    const minutesOffset = Math.abs(timeZoneOffset) % 60;
+    const formattedTimeZoneOffset = `GMT${
+      timeZoneOffset >= 0 ? "-" : "+"
+    }${hoursOffset.toString().padStart(2, "0")}:${minutesOffset
+      .toString()
+      .padStart(2, "0")}`;
+    try {
+      const res = await this.connection.post(
+        "/studio/sessions",
+        JSON.stringify({
+          baseUri: this.connection.getUri() + "/SASStudio/",
+          locale: "en_US",
+          zone: formattedTimeZoneOffset,
+        }),
+        {
+          headers: {
+            "Content-Type": "application/json",
+            Accept: "application/json",
+          },
+        },
+      );
+      const sessionId = res.data.id;
+      await this.connection.post(
+        "/SASStudio/sasexec/{sessionId}/associateFlowObj".replace(
+          `{${"sessionId"}}`,
+          sessionId,
+        ),
+        JSON.stringify({
+          name: name,
+          uri: "sascontent:" + getResourceId(uri),
+          parentUri: "sascontent:" + getResourceIdFromItem(parent),
+          currentParentUri: null,
+        }),
+        {
+          headers: {
+            "Content-Type": "application/json",
+            Accept: "*/*",
+          },
+        },
+      );
+    } catch (error) {
+      console.log(error);
+    }
+  }
+
   public async getUri(item: ContentItem, readOnly: boolean): Promise<Uri> {
     if (item.type !== "reference") {
       return getUri(item, readOnly);
--- a/client/src/components/ContentNavigator/ContentModel.ts
+++ b/client/src/components/ContentNavigator/ContentModel.ts
@@ -331,7 +331,6 @@ export class ContentModel {
         "If-Unmodified-Since": this.fileTokenMaps[resourceId].lastModified,
       };
     } else {
-      // create timestamp for now in this format Mon, 07 Aug 2023 13:28:40 GMT
       const now = new Date();
       const timestamp = now.toUTCString();
       headers = {
--- a/client/src/components/ContentNavigator/ContentModel.ts
+++ b/client/src/components/ContentNavigator/ContentModel.ts
@@ -165,10 +165,9 @@ export class ContentModel {
     if (!ancestorsLink) {
       return;
     }
-
     const resp = await this.connection.get(ancestorsLink.uri);
-    if (resp.data.ancestors && resp.data.ancestors.length > 0) {
-      return resp.data.ancestors[0];
+    if (resp.data && resp.data.length > 0) {
+      return resp.data[0];
     }
   }
 
@@ -324,17 +323,36 @@ export class ContentModel {
 
   public async saveContentToUri(uri: Uri, content: string): Promise<void> {
     const resourceId = getResourceId(uri);
-    const res = await this.connection.put(resourceId + "/content", content, {
-      headers: {
+    let headers = {};
+    if (resourceId in this.fileTokenMaps) {
+      headers = {
         "Content-Type": "text/plain",
         "If-Match": this.fileTokenMaps[resourceId].etag,
         "If-Unmodified-Since": this.fileTokenMaps[resourceId].lastModified,
-      },
-    });
-    this.fileTokenMaps[resourceId] = {
-      etag: res.headers.etag,
-      lastModified: res.headers["last-modified"],
-    };
+      };
+    } else {
+      // create timestamp for now in this format Mon, 07 Aug 2023 13:28:40 GMT
+      const now = new Date();
+      const timestamp = now.toUTCString();
+      headers = {
+        "Content-Type": "text/plain",
+        "If-Unmodified-Since": timestamp,
+      };
+    }
+    try {
+      const res = await this.connection.put(resourceId + "/content", content, {
+        // if resourceId in this.fileTokenMaps
+        headers: headers,
+      });
+      if (resourceId in this.fileTokenMaps) {
+        this.fileTokenMaps[resourceId] = {
+          etag: res.headers.etag,
+          lastModified: res.headers["last-modified"],
+        };
+      }
+    } catch (error) {
+      console.log(error);
+    }
   }
 
   public async getUri(item: ContentItem, readOnly: boolean): Promise<Uri> {
--- a/client/src/components/ContentNavigator/convert.ts
+++ b/client/src/components/ContentNavigator/convert.ts
@@ -1,12 +1,149 @@
 import * as fs from "fs";
 
+const stepRef: Record<string, string> = {
+  sas: "/dataFlows/steps/a7190700-f59c-4a94-afe2-214ce639fcde",
+  python: "/dataFlows/steps/ab59f8c4-af9a-4608-a5d5-a8365357bb99",
+};
+
+interface Entry {
+  language: string;
+  code: string;
+}
+
+interface FlowData {
+  creationTimeStamp: string;
+  modifiedTimeStamp: string;
+  createdBy: string;
+  modifiedBy: string;
+  version: number;
+  id: null | string;
+  name: string;
+  description: null | string;
+  properties: Record<string, string>;
+  links: any[]; // Replace 'any' with the appropriate type when known
+  nodes: Record<string, any>; // Replace 'any' with the appropriate type when known
+  parameters: Record<string, any>; // Replace 'any' with the appropriate type when known
+  connections: any[]; // Replace 'any' with the appropriate type when known
+  extendedProperties: Record<string, any>; // Replace 'any' with the appropriate type when known
+  stickyNotes: any[]; // Replace 'any' with the appropriate type when known
+}
+
+function generateFlowData(inputList: Entry[], outputFile: string): FlowData {
+  const flowData: FlowData = {
+    creationTimeStamp: "2023-07-25T15:19:29.821Z",
+    modifiedTimeStamp: "2023-07-25T15:38:15.771Z",
+    createdBy: "user@sas.com",
+    modifiedBy: "user@sas.com",
+    version: 2,
+    id: null,
+    name: outputFile,
+    description: null,
+    properties: {
+      UI_PROP_DF_OPTIMIZE: "false",
+      UI_PROP_DF_ID: "120081fc-2d7f-4cfc-bcd3-47f9684e9763",
+      UI_PROP_DF_EXECUTION_ORDERED: "false",
+    },
+    links: [],
+    nodes: {},
+    parameters: {},
+    connections: [],
+    extendedProperties: {},
+    stickyNotes: [],
+  };
+
+  for (let idx = 0; idx < inputList.length; idx++) {
+    const entry = inputList[idx];
+    const nodeId = `id-${idx + 1}`;
+
+    const stepNode = {
+      nodeType: "step",
+      version: 1,
+      id: nodeId,
+      name: `${entry.language} Program`,
+      note: {
+        version: 1,
+        id: nodeId,
+        name: null,
+        description: null,
+        properties: {
+          UI_NOTE_PROP_HEIGHT: "0",
+          UI_NOTE_PROP_IS_EXPANDED: "false",
+          UI_NOTE_PROP_IS_STICKYNOTE: "false",
+          UI_NOTE_PROP_WIDTH: "0",
+        },
+      },
+      priority: idx,
+      properties: {
+        UI_PROP_COLORGRP: "0",
+        UI_PROP_IS_INPUT_EXPANDED: "false",
+        UI_PROP_IS_OUTPUT_EXPANDED: "false",
+        UI_PROP_NODE_DATA_ID: `id-${idx + 2}`,
+        UI_PROP_NODE_DATA_MODIFIED_DATE: "1684001468354",
+        UI_PROP_XPOS: String((idx + 1) * 150), // Adjust the X position as needed
+        UI_PROP_YPOS: "75", // Adjust the Y position as needed
+      },
+      portMappings: [],
+      stepReference: {
+        type: "uri",
+        path: stepRef[entry.language],
+      },
+      arguments: {
+        codeOptions: {
+          code: entry.code,
+          contentType: "embedded",
+          logHTML: "",
+          resultsHTML: "",
+          variables: [
+            {
+              name: "_input1",
+              value: {
+                portIndex: 0,
+                portName: "inTables",
+                referenceType: "inputPort",
+              },
+            },
+            {
+              name: "_output1",
+              value: {
+                arguments: {},
+                portIndex: 0,
+                portName: "outTables",
+                referenceType: "outputPort",
+              },
+            },
+          ],
+        },
+      },
+    };
+
+    flowData.nodes[nodeId] = stepNode;
+
+    if (idx > 0) {
+      const connection = {
+        sourcePort: {
+          node: `id-${idx}`,
+          portName: "outTables",
+          index: 0,
+        },
+        targetPort: {
+          node: nodeId,
+          portName: "inTables",
+          index: 0,
+        },
+      };
+      flowData.connections.push(connection);
+    }
+  }
+  return flowData;
+}
+
 export function convert_sasnb_to_flw(
-  filePath: string,
-): { code: string; language: string }[] {
+  content: string,
+  outputName: string,
+): Uint8Array {
   const codeList = [];
 
   try {
-    const content = fs.readFileSync(filePath, "utf-8");
     const notebookContent = JSON.parse(content);
 
     for (const cell of notebookContent) {
@@ -30,5 +167,10 @@ export function convert_sasnb_to_flw(
     console.error("Error reading or parsing the .sasnb file:", error);
   }
 
-  return codeList;
+  const flowData = generateFlowData(codeList, outputName);
+  // encode json to utf8 bytes without new lines and spaces
+  const flowDataString = JSON.stringify(flowData, null, 0);
+  const flowDataUint8Array = new TextEncoder().encode(flowDataString);
+
+  return flowDataUint8Array;
 }
--- a/client/src/components/ContentNavigator/convert.ts
+++ b/client/src/components/ContentNavigator/convert.ts
@@ -15,29 +15,35 @@ interface Entry {
   code: string;
 }
 
-interface FlowData {
-  creationTimeStamp: string;
-  modifiedTimeStamp: string;
-  createdBy: string;
-  modifiedBy: string;
-  version: number;
-  id: null | string;
-  name: string;
-  description: null | string;
-  properties: Record<string, string>;
-  links: any[];
-  nodes: Record<string, any>;
-  parameters: Record<string, any>;
-  connections: any[];
-  extendedProperties: Record<string, any>;
-  stickyNotes: any[];
+// create a function that takes idx as input
+// and returns a json depending if idx is first, last or in between
+function getPropPort(idx: number, inputList: Entry[]): Record<string, string> {
+  if (idx === 0) {
+    return {
+      "UI_PROP_PORT_DESCRIPTION|outTables|0": "Output tables",
+      "UI_PROP_PORT_LABEL|outTables|0": "Output table 1",
+    };
+  } else if (idx === inputList.length - 1) {
+    return {
+      "UI_PROP_PORT_DESCRIPTION|inTables|0": "Input tables",
+      "UI_PROP_PORT_LABEL|inTables|0": "Input table 1",
+    };
+  } else {
+    return {
+      "UI_PROP_PORT_DESCRIPTION|inTables|0": "Input tables",
+      "UI_PROP_PORT_LABEL|inTables|0": "Input table 1",
+      "UI_PROP_PORT_DESCRIPTION|outTables|0": "Output tables",
+      "UI_PROP_PORT_LABEL|outTables|0": "Output table 1",
+    };
+  }
 }
 
-function generateFlowData(inputList: Entry[], outputFile: string): FlowData {
+function generateFlowData(inputList: Entry[], outputFile: string) {
   const now = new Date();
   const nowString = now.toISOString();
   const nowTimestamp = String(now.getTime());
-  const flowData: FlowData = {
+  const arrayIdNode: string[] = [];
+  const flowData = {
     creationTimeStamp: nowString,
     modifiedTimeStamp: nowString,
     createdBy: "user@sas.com",
@@ -60,17 +66,19 @@ function generateFlowData(inputList: Entry[], outputFile: string): FlowData {
   };
 
   for (let idx = 0; idx < inputList.length; idx++) {
+    const idNode = `id-${Date.now()}-${Math.floor(Math.random() * 10000)}`;
+    arrayIdNode.push(idNode);
+    const idNote = `id-${Date.now()}-${Math.floor(Math.random() * 10000)}`;
     const entry = inputList[idx];
-    const nodeId = `id-${idx + 1}`;
-
+    const propports = getPropPort(idx, inputList);
     const stepNode = {
       nodeType: "step",
       version: 1,
-      id: nodeId,
+      id: idNode,
       name: stepTitle[entry.language],
       note: {
         version: 1,
-        id: nodeId,
+        id: idNote,
         name: null,
         description: null,
         properties: {
@@ -87,6 +95,7 @@ function generateFlowData(inputList: Entry[], outputFile: string): FlowData {
         UI_PROP_IS_OUTPUT_EXPANDED: "false",
         UI_PROP_NODE_DATA_ID: stepRef[entry.language],
         UI_PROP_NODE_DATA_MODIFIED_DATE: nowTimestamp,
+
         UI_PROP_XPOS: String((idx + 1) * 150),
         UI_PROP_YPOS: "75",
       },
@@ -132,18 +141,18 @@ function generateFlowData(inputList: Entry[], outputFile: string): FlowData {
         },
       },
     };
-
-    flowData.nodes[nodeId] = stepNode;
+    stepNode.properties = { ...stepNode.properties, ...propports };
+    flowData.nodes[idNode] = stepNode;
 
     if (idx > 0) {
       const connection = {
         sourcePort: {
-          node: `id-${idx}`,
+          node: arrayIdNode[idx - 1],
           portName: "outTables",
           index: 0,
         },
         targetPort: {
-          node: nodeId,
+          node: arrayIdNode[idx],
           portName: "inTables",
           index: 0,
         },
@@ -168,13 +177,9 @@ export function convert_sasnb_to_flw(
       if (code !== "") {
         const language = cell.language;
         if (language === "sql") {
-          code = `
-          PROC SQL;
-              ${code}
-              ;
-          QUIT;
-          RUN;
-          `;
+          code = `PROC SQL;
+${code};
+QUIT;`;
         }
         codeList.push({ code, language });
       }
--- a/client/src/components/ContentNavigator/convert.ts
+++ b/client/src/components/ContentNavigator/convert.ts
@@ -1,8 +1,13 @@
-import * as fs from "fs";
-
 const stepRef: Record<string, string> = {
   sas: "/dataFlows/steps/a7190700-f59c-4a94-afe2-214ce639fcde",
+  sql: "/dataFlows/steps/a7190700-f59c-4a94-afe2-214ce639fcde",
   python: "/dataFlows/steps/ab59f8c4-af9a-4608-a5d5-a8365357bb99",
+};
+
+const stepTitle: Record<string, string> = {
+  sas: "SAS Program",
+  sql: "SQL Program",
+  python: "Python Program",
 };
 
 interface Entry {
@@ -20,12 +25,12 @@ interface FlowData {
   name: string;
   description: null | string;
   properties: Record<string, string>;
-  links: any[]; // Replace 'any' with the appropriate type when known
-  nodes: Record<string, any>; // Replace 'any' with the appropriate type when known
-  parameters: Record<string, any>; // Replace 'any' with the appropriate type when known
-  connections: any[]; // Replace 'any' with the appropriate type when known
-  extendedProperties: Record<string, any>; // Replace 'any' with the appropriate type when known
-  stickyNotes: any[]; // Replace 'any' with the appropriate type when known
+  links: any[];
+  nodes: Record<string, any>;
+  parameters: Record<string, any>;
+  connections: any[];
+  extendedProperties: Record<string, any>;
+  stickyNotes: any[];
 }
 
 function generateFlowData(inputList: Entry[], outputFile: string): FlowData {
@@ -59,7 +64,7 @@ function generateFlowData(inputList: Entry[], outputFile: string): FlowData {
       nodeType: "step",
       version: 1,
       id: nodeId,
-      name: `${entry.language} Program`,
+      name: stepTitle[entry.language],
       note: {
         version: 1,
         id: nodeId,
@@ -149,9 +154,8 @@ export function convert_sasnb_to_flw(
     for (const cell of notebookContent) {
       let code = cell.value;
       if (code !== "") {
-        let language = cell.language;
+        const language = cell.language;
         if (language === "sql") {
-          language = "sas";
           code = `
           PROC SQL;
               ${code}
--- a/client/src/components/ContentNavigator/convert.ts
+++ b/client/src/components/ContentNavigator/convert.ts
@@ -1,7 +1,7 @@
 const stepRef: Record<string, string> = {
-  sas: "/dataFlows/steps/a7190700-f59c-4a94-afe2-214ce639fcde",
-  sql: "/dataFlows/steps/a7190700-f59c-4a94-afe2-214ce639fcde",
-  python: "/dataFlows/steps/ab59f8c4-af9a-4608-a5d5-a8365357bb99",
+  sas: "a7190700-f59c-4a94-afe2-214ce639fcde",
+  sql: "a7190700-f59c-4a94-afe2-214ce639fcde",
+  python: "ab59f8c4-af9a-4608-a5d5-a8365357bb99",
 };
 
 const stepTitle: Record<string, string> = {
@@ -34,9 +34,12 @@ interface FlowData {
 }
 
 function generateFlowData(inputList: Entry[], outputFile: string): FlowData {
+  const now = new Date();
+  const nowString = now.toISOString();
+  const nowTimestamp = String(now.getTime());
   const flowData: FlowData = {
-    creationTimeStamp: "2023-07-25T15:19:29.821Z",
-    modifiedTimeStamp: "2023-07-25T15:38:15.771Z",
+    creationTimeStamp: nowString,
+    modifiedTimeStamp: nowString,
     createdBy: "user@sas.com",
     modifiedBy: "user@sas.com",
     version: 2,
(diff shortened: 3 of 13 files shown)


client/src/components/ContentNavigator/index.ts
```diff
@@ -247,6 +250,71 @@ class ContentNavigator implements SubscriptionProvider {
...
+              );
+              window.showInformationMessage(
+                Messages.SasnbToFlwConversionSuccess,
+              );
+            } catch (error) {
+              window.showErrorMessage(Messages.SasnbToFlwConversionError);
+            }
+          } else {
+            window.showErrorMessage(Messages.InvalidSasnbFileError);
+          }
```
scottdover, line 315: Can we make this a method inside `ContentDataProvider`?

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+    extendedProperties: {},
+    stickyNotes: [],
+  };
+
+  for (let idx = 0; idx < inputList.length; idx++) {
+    const idNode = `id-${Date.now()}-${Math.floor(Math.random() * 10000)}`;
+    arrayIdNode.push(idNode);
+    const idNote = `id-${Date.now()}-${Math.floor(Math.random() * 10000)}`;
+    const entry = inputList[idx];
+    const propports = getPropPort(idx, inputList);
```
scottdover, line 74:
```suggestion
    const propPorts = getPropPort(idx, inputList);
```

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+                portIndex: 0,
+                portName: "outTables",
+                referenceType: "outputPort",
+              },
+            },
+          ],
+        },
+      },
+    };
+    stepNode.properties = { ...stepNode.properties, ...propports };
```
scottdover, line 145:
```suggestion
    stepNode.properties = { ...stepNode.properties, ...propPorts };
```

client/src/components/ContentNavigator/index.ts
```diff
@@ -247,6 +250,71 @@ class ContentNavigator implements SubscriptionProvider {
...
+              return;
+            }
+
+            // check that name does not exist
+            const parent = await this.contentDataProvider.getParent(resource);
+            const children = await this.contentDataProvider.getChildren(parent);
+            if (children.some((child) => child.name === name)) {
+              window.showErrorMessage(Messages.FileAlreadyExistsError);
+              return;
+            }
```
scottdover, line 281: We could skip this check and use `NewFileCreationError` after calling `createFile` (we do this elsewhere in code)

client/src/components/ContentNavigator/index.ts
```diff
@@ -247,6 +250,71 @@ class ContentNavigator implements SubscriptionProvider {
...
+                return;
+              }
+              const newUri = await this.contentDataProvider.createFile(
+                parent,
+                name,
+              );
+              await this.contentDataProvider.writeFile(
+                newUri,
+                flowDataUint8Array,
+              );
```
scottdover, line 301: We should be able to call _only_ `createFile` with the flow data as the third parameter, I believe

scottdover (commented):
This looks like a good start. The biggest question I had is, can we assume a user _wants_ to connect all their nodes created from cells in their notebook. I don't have a great answer to that, because I'm not super familiar with the sas notebook use-case. @scnwwu may have better feedback with that

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+    links: [],
+    nodes: {},
+    parameters: {},
+    connections: [],
+    extendedProperties: {},
+    stickyNotes: [],
+  };
+
+  for (let idx = 0; idx < inputList.length; idx++) {
+    const idNode = `id-${Date.now()}-${Math.floor(Math.random() * 10000)}`;
```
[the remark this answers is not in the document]
sahbic: This choice is based on examples of .flw files generated by SAS Studio. uuids v4 are used for resource ids but node ids and note ids use this format of "id-timestamp-XXXX"

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+QUIT;`;
+        }
+        codeList.push({ code, language });
+      }
+    }
+  } catch (error) {
+    console.error("Error reading or parsing the .sasnb file:", error);
+  }
+
+  const flowData = generateFlowData(codeList, outputName);
```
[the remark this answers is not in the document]
sahbic: My assumption here is that when a user wants to convert a notebook to a flow, the aim is to use execute it in production from top to bottom. The links in the flow allow to "force" an execution order of the nodes.

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+function generateFlowData(inputList: Entry[], outputFile: string) {
+  const now = new Date();
+  const nowString = now.toISOString();
+  const nowTimestamp = String(now.getTime());
+  const arrayIdNode: string[] = [];
+  const flowData = {
+    creationTimeStamp: nowString,
+    modifiedTimeStamp: nowString,
+    createdBy: "user@sas.com",
+    modifiedBy: "user@sas.com",
```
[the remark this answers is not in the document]
sahbic: This information is actually overwritten by SAS Studio after the association API call that associates the .flw file with SAS Studio. It's replaced by the user email after that step. So we only need a placeholder here.

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+      },
+      priority: idx,
+      properties: {
+        UI_PROP_COLORGRP: "0",
+        UI_PROP_IS_INPUT_EXPANDED: "false",
+        UI_PROP_IS_OUTPUT_EXPANDED: "false",
+        UI_PROP_NODE_DATA_ID: stepRef[entry.language],
+        UI_PROP_NODE_DATA_MODIFIED_DATE: nowTimestamp,
+
+        UI_PROP_XPOS: String((idx + 1) * 150),
```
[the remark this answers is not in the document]
sahbic: fixed in e0fae359ae8827f10f5fa650824020d69e4432da

client/src/components/ContentNavigator/convert.ts
sahbic: changed variable name in cf89c943731e7c2fef458becd0ea636475db0f59

client/src/components/ContentNavigator/convert.ts
sahbic: changed variable name in cf89c943731e7c2fef458becd0ea636475db0f59

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+        },
+      },
+      priority: idx,
+      properties: {
+        UI_PROP_COLORGRP: "0",
+        UI_PROP_IS_INPUT_EXPANDED: "false",
+        UI_PROP_IS_OUTPUT_EXPANDED: "false",
+        UI_PROP_NODE_DATA_ID: stepRef[entry.language],
+        UI_PROP_NODE_DATA_MODIFIED_DATE: nowTimestamp,
+
```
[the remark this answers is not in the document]
sahbic: fixed here 28f4783660103e5f42576b66a2dd0dfb473860a0

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+          index: 0,
+        },
+      };
+      flowData.connections.push(connection);
+    }
+  }
+  return flowData;
+}
+
+export function convert_sasnb_to_flw(
```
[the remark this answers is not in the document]
sahbic: changed function name, resolved by 0b8685e7f0c72ed3e9facd767af935242e7acf3a

client/src/components/ContentNavigator/index.ts
```diff
@@ -247,6 +250,74 @@ class ContentNavigator implements SubscriptionProvider {
...
+            // check that name does not exist
+            const parent = await this.contentDataProvider.getParent(resource);
+            const children = await this.contentDataProvider.getChildren(parent);
+            if (children.some((child) => child.name === name)) {
+              window.showErrorMessage(Messages.FileAlreadyExistsError);
+              return;
+            }
+
+            try {
+              const contentUri: string =
```
scnwwu, line 284: this variable name is misleading, it looks not a Uri

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,201 @@
...
+): Uint8Array {
+  const codeList = [];
+
+  try {
+    const notebookContent = JSON.parse(content);
+
+    for (const cell of notebookContent) {
+      let code = cell.value;
+      if (code !== "") {
+        const language = cell.language;
```
scnwwu, line 182: so what about markdown? how about converting markdown to a Note?

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,201 @@
...
+          index: 0,
+        },
+      };
+      flowData.connections.push(connection);
+    }
+  }
+  return flowData;
+}
+
+export function convertSASNotebookToFlow(
```
scnwwu, line 170: I guess it would be great if a test can be created for this, so that there's a clear view of the expected input/output.

client/src/components/ContentNavigator/convert.ts
sahbic: This is possible but needs to be further studied.
Notes are associated to a Node in SAS Studio but can be inserted at any stage in a Notebook.
Should we assume that all Markdown cells preceding (or following ?) a Code cell are related to that code and added as a Note to the Code Node ?
Markdown titles and subtitles (if available) could also be used as Swimlane / Node titles.

This feature would add its share of complexity should we implement it or keep it as a future improvement ?

client/src/components/ContentNavigator/index.ts
sahbic: resolved by b8c3d3cb4c704147184535b15badb5388f1d8fb5

client/src/components/ContentNavigator/convert.ts
sahbic: Yes I would go with the option of ignoring markdown for now, fixed in ef20e4fa61c12fab7ed885b70c890a7656c3668b. Thanks

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -165,10 +165,9 @@ export class ContentModel {
     if (!ancestorsLink) {
       return;
     }
-
     const resp = await this.connection.get(ancestorsLink.uri);
-    if (resp.data.ancestors && resp.data.ancestors.length > 0) {
-      return resp.data.ancestors[0];
+    if (resp.data && resp.data.length > 0) {
```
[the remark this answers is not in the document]
sahbic: I suppose it was a bug. *ancestors* does not exist in the response data. I also noted that this function was not used anywhere else.

client/src/components/ContentNavigator/index.ts
sahbic: Done in 048298e018a51e572e6320a9e44ad08515e02ae5.
It seems that the .flw files are not "seen" anymore by VSCode after the SAS Studio association step.
SAS Studio also handles duplicated file names by adding a suffix.

client/src/components/ContentNavigator/index.ts
sahbic: Yes indeed, done in 3ec847bd25ab2f905bbe0d735eac5ec89c4f6c03

client/src/components/ContentNavigator/index.ts
```diff
@@ -247,6 +250,71 @@ class ContentNavigator implements SubscriptionProvider {
...
+
+            if (!name) {
+              // User canceled the input box
+              return;
+            }
+
+            if (!name.endsWith(".flw")) {
+              window.showErrorMessage(Messages.InvalidFlwFileNameError);
+              return;
+            }
```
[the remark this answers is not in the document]
sahbic: added input validation in 4fe5cdafe05262209ba32ce48e0daaff69cd9798

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -324,17 +323,90 @@ export class ContentModel {
...
+        "Content-Type": "text/plain",
+        "If-Unmodified-Since": timestamp,
+      };
+    }
+    try {
+      const res = await this.connection.put(resourceId + "/content", content, {
+        // if resourceId in this.fileTokenMaps
+        headers: headers,
+      });
+      if (resourceId in this.fileTokenMaps) {
```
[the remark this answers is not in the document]
sahbic: removed in 778508b2f58aed24a51a3a9c329d5ec71ece4657

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -324,17 +323,90 @@ export class ContentModel {
...
-    };
+      };
+    } else {
+      const now = new Date();
+      const timestamp = now.toUTCString();
+      headers = {
+        "Content-Type": "text/plain",
+        "If-Unmodified-Since": timestamp,
+      };
+    }
```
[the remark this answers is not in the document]
sahbic: created a separate function in 778508b2f58aed24a51a3a9c329d5ec71ece4657

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,198 @@
...
+    stickyNotes: [],
+  };
+
+  for (let idx = 0; idx < inputList.length; idx++) {
+    const idNode = `id-${Date.now()}-${Math.floor(Math.random() * 10000)}`;
+    arrayIdNode.push(idNode);
+    const idNote = `id-${Date.now()}-${Math.floor(Math.random() * 10000)}`;
+    const entry = inputList[idx];
+    const propports = getPropPort(idx, inputList);
+    const stepNode = {
```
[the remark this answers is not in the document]
sahbic: added base flow and node to separate static and changing parameters in 484a6427d3f666a3d659ef40742286843fef5b2b

client/src/components/ContentNavigator/convert.ts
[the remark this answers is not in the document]
sahbic: removed the placeholder email in 484a6427d3f666a3d659ef40742286843fef5b2b

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -324,17 +323,90 @@ export class ContentModel {
...
+          etag: res.headers.etag,
+          lastModified: res.headers["last-modified"],
+        };
+      }
+    } catch (error) {
+      console.log(error);
+    }
+  }
+
+  public async associateFlowObject(
```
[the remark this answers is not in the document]
sahbic: This makes sense, the files allowing to handle APIs are in `src/connection/rest/api`. `DataAccessApi` for example is defined in `compute.ts`. We could have `studio.ts` there for studio related calls. However the files in this folder seem to be generated automatically ?
I do not fully understand how these files are generated, if this change is required I could use some help for this task.

client/src/components/ContentNavigator/index.ts
sahbic: done in 2c900b3fdb3790fe93dce7067ca4ebb5ed93d544. I have also moved the `handleCreationResponse` function to `ContentDataProvider`

client/src/components/ContentNavigator/convert.ts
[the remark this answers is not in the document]
scottdover: I had the opportunity to chat with a few folks overseeing UX of this project. It was their recommendation to setup your script such that it creates a flow where each node is in its own swimlane, with a title giving some indication of where it came from. Here's an example of what the conversion process may look like for a notebook with a sas program & python program.

Screenshot 2023-08-17 at 9 41 51 AM

client/src/components/ContentNavigator/ContentModel.ts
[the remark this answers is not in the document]
scottdover: Yeah, you're right. Those are generated automatically. I would suggest putting this in something like `src/connection/studio/index.ts` and having related files w/n that folder

client/src/components/ContentNavigator/convert.ts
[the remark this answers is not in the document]
scottdover: I think it's safe to use `v4()` here, as this is what studio currently uses when creating new nodes

client/src/components/ContentNavigator/index.ts
```diff
@@ -247,6 +247,44 @@ class ContentNavigator implements SubscriptionProvider {
...
+        "SAS.convertSasnbToFlw",
+        async (resource: ContentItem) => {
+          const resourceUri = getUri(resource);
+          // Ensure that the command is only available for .sasnb files
+          if (resourceUri.path.endsWith(".sasnb")) {
+            // Open window to chose the name and location of the new .flw file
+            const name = await window.showInputBox({
+              prompt: "Enter the name for the new .flw file",
+              value: resource.name.replace(".sasnb", ".flw"),
+              validateInput: fileValidator,
```
scottdover, line 260: Can we make a `flowFileValidator` here to test the file and show `InvalidFlwFileNameError` there instead of checking in the if statement below?

scottdover (commented):
This is looking good. The only bigger thing that stands out is in what kind of flow this creates. I've attached a screenshot that'll hopefully give an idea of what's expected :)

Outside of that, I've noticed some variables/errors/etc throughout the code that use `flw` as part of a variable name (i.e. `InvalidFlwFileNameError`). Can we update those references to `flow` (like `InvalidFlowFileNameError`).

Thank you for taking on this work. It'll make a great addition to the extension :)

client/src/components/ContentNavigator/index.ts
sahbic: done in 3bf40d5b4ae1b22e902919ff178e8e49ae207131

client/src/components/ContentNavigator/convert.ts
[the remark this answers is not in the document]
sahbic: done in c74837c6e2e3c7a1dc955763b3b1474082637aa8

client/src/components/ContentNavigator/ContentModel.ts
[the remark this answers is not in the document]
sahbic: Moved SAS Studio related calls to a new file here 99ce799ce36a68ecaf83c26fea5962261ccc7e21

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -334,6 +343,51 @@ class ContentDataProvider
...
+    const parent = await this.getParent(item);
+    const resourceUri = getUri(item);
+    try {
+      // get the content of the .sasnb file
+      const contentString: string = await this.provideTextDocumentContent(
+        resourceUri,
+      );
+      // convert the .sasnb file to a flowData object
+      const flowDataUint8Array = convertSASNotebookToFlow(contentString, name);
+      // error if content is empty
```
scottdover, line 372: This comment is unneeded

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -322,19 +325,61 @@ export class ContentModel {
...
+  public async associateFlowFile(
+    name: string,
+    uri: Uri,
+    parent: ContentItem,
+  ): Promise<string | undefined> {
+    try {
+      // create a studio session
+      const sessionId = await createStudioSession(this.connection);
+      // associate flow object
+      const flowUri = associateFlowObject(
```
scottdover, line 372: We _could_ just return here instead of assigning to a variable

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -322,19 +325,61 @@ export class ContentModel {
...
+    }
+  }
+
+  public async associateFlowFile(
+    name: string,
+    uri: Uri,
+    parent: ContentItem,
+  ): Promise<string | undefined> {
+    try {
+      // create a studio session
```
scottdover, line 369: Unneeded comment

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -322,19 +325,61 @@ export class ContentModel {
...
+      const sessionId = await createStudioSession(this.connection);
+      // associate flow object
```
scottdover, line 371: Unneeded comment

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,241 @@
...
+        },
+        targetPort: {
+          node: arrayIdNode[idx],
+          portName: "inTables",
+          index: 0,
+        },
+      };
+      flowData.connections.push(connection);
+    }
+  }
```
scottdover, line 204: I'm guessing this will change when we move to using swimlanes and _not_ creating connections?

client/src/components/ContentNavigator/convert.ts
scottdover: Yeah, I'd love to see more testing in this PR

client/src/components/ContentNavigator/index.ts
```diff
@@ -247,6 +259,38 @@ class ContentNavigator implements SubscriptionProvider {
           "workbench.actions.treeView.contentdataprovider.collapseAll",
         );
       }),
+      commands.registerCommand(
+        "SAS.convertSasnbToFlw",
+        async (resource: ContentItem) => {
+          const resourceUri = getUri(resource);
+          // Ensure that the command is only available for .sasnb files
+          if (resourceUri.path.endsWith(".sasnb")) {
```
scottdover, line 267: Do we need this? My expectation is that we'll only show the context menu triggering this action when `convertSasnbToFlw` is present in the context

client/src/components/ContentNavigator/index.ts
```diff
@@ -32,6 +32,18 @@ const fileValidator = (value: string): string | null =>
   )
     ? null
     : Messages.FileValidationError;
+
+const flowFileValidator = (value: string): string | null => {
+  let res = null;
+  if (!/^([^/<>;\\{}?#]+)\.\w+$/.test(value)) {
+    res = Messages.FileValidationError;
+  }
```
scottdover, line 40: Instead of duplication, we should call `fileValidator(value)` here

client/src/components/ContentNavigator/index.ts
scottdover: If we _do_ need this, we should bail from this function early instead. Something like...

```
if (!resourceUri.path.endsWith(".sasnb")) {
    return;
}
```

...and unindent the rest

client/src/connection/studio/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  const hoursOffset = Math.floor(Math.abs(timeZoneOffset) / 60);
+  const minutesOffset = Math.abs(timeZoneOffset) % 60;
+  // Convert the time zone offset to the desired format (e.g., "GMT+02:00")
+  const formattedTimeZoneOffset = `GMT${
+    timeZoneOffset >= 0 ? "-" : "+"
+  }${hoursOffset.toString().padStart(2, "0")}:${minutesOffset
+    .toString()
+    .padStart(2, "0")}`;
+  // Create the request body
+  const studioSessionRequest = JSON.stringify({
```
scottdover, line 21: You _could_ return here instead of variable assignment

client/src/connection/studio/index.ts
```diff
@@ -0,0 +1,77 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { AxiosInstance } from "axios";
+
+function getStudioSessionRequest(connection: AxiosInstance) {
+  // Create a Date object representing the current date and time
+  const date = new Date();
+  // Get the time zone offset
```
scottdover, line 9: unneeded comment

client/src/connection/studio/index.ts
```diff
@@ -0,0 +1,77 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { AxiosInstance } from "axios";
+
+function getStudioSessionRequest(connection: AxiosInstance) {
+  // Create a Date object representing the current date and time
```
scottdover, line 7: unneeded comment

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -334,6 +343,51 @@ class ContentDataProvider
...
+  ): Promise<void> {
+    if (!newUri) {
+      window.showErrorMessage(errorMessage);
+      return;
+    }
+
+    this.reveal(resource);
+  }
+
+  public async convertSasnbToFlw(
```
scottdover, line 359: Just duplicating this here so it gets the little resolve button :) 

> I've noticed some variables/errors/etc throughout the code that use flw as part of a variable name (i.e. InvalidFlwFileNameError). Can we update those references to flow (like InvalidFlowFileNameError).

In general, if function/variable/constant names can use the full names of things (i.e. SASNotebook, Flow), that would be ideal

sahbic: done in 9a191686d48ab21e54aa2247f2b7abd764b0d242

client/src/connection/studio/index.ts
sahbic: done in f5253ce014a5fec429456474725f717e0f2609a0

client/src/connection/studio/index.ts
sahbic: done in f5253ce014a5fec429456474725f717e0f2609a0

client/src/connection/studio/index.ts
sahbic: done in f5253ce014a5fec429456474725f717e0f2609a0

client/src/components/ContentNavigator/index.ts
sahbic: done in f5253ce014a5fec429456474725f717e0f2609a0

client/src/components/ContentNavigator/index.ts
sahbic: Right, not needed anymore since the command is only available for `.sasnb` files as defined in `components/ContentNavigator/utils.ts` . Fixed in f5253ce014a5fec429456474725f717e0f2609a0

client/src/components/ContentNavigator/convert.ts
sahbic: Added a new default conversion mode in user settings. Default is **swimlane** conversion but user can switch to **node** conversion. Added in 8cb6a7941c05fbc3634489bca0abfb5609c49876

client/src/components/ContentNavigator/ContentModel.ts
sahbic: done in f5253ce014a5fec429456474725f717e0f2609a0

client/src/components/ContentNavigator/ContentModel.ts
sahbic: done in f5253ce014a5fec429456474725f717e0f2609a0

client/src/components/ContentNavigator/ContentModel.ts
sahbic: done in f5253ce014a5fec429456474725f717e0f2609a0

client/src/components/ContentNavigator/ContentDataProvider.ts
sahbic: done in f5253ce014a5fec429456474725f717e0f2609a0

sahbic:
Thank you for your comments and suggestions to enhance the code.

Studio Flows are only available for recent versions of SAS Viya, I would like to restrict this feature to newer versions of SAS Viya only. How would you suggest to proceed ?

I see there is a private API Call to get the ViyaCadence in `ContentNavigator/ContentModel.ts` and I would like to use that output in `ContentNavigator/utils.ts` ?

scottdover:
> Thank you for your comments and suggestions to enhance the code.
> 
> Studio Flows are only available for recent versions of SAS Viya, I would like to restrict this feature to newer versions of SAS Viya only. How would you suggest to proceed ?
> 
> I see there is a private API Call to get the ViyaCadence in `ContentNavigator/ContentModel.ts` and I would like to use that output in `ContentNavigator/utils.ts` ?

My opinion is that we should use `viyaCadence` as minimally as possible. What happens if you try to create a flow for an instance that doesn't support it? I'm wondering if it returns an error response we could interpret and respond accordingly?

@scnwwu may have other thoughts on this

client/src/components/ContentNavigator/index.ts
```diff
@@ -247,6 +256,32 @@ class ContentNavigator implements SubscriptionProvider {
           "workbench.actions.treeView.contentdataprovider.collapseAll",
         );
       }),
+      commands.registerCommand(
+        "SAS.convertNotebookToFlow",
+        async (resource: ContentItem) => {
+          // Open window to chose the name and location of the new .flw file
+          const name = await window.showInputBox({
+            prompt: "Enter the name for the new .flw file",
```
scottdover, line 264: This should have an entry in `Messages` as well

scottdover (commented):
The only remaining things that stand out to me (outside of the one comment) are...
 - adding tests around nb -> flw conversion
 - updating things to target only some versions of viya

Everything else looks good.

package.json
```diff
@@ -120,6 +120,20 @@
             "default": true,
             "description": "%configuration.SAS.session.outputHtml%"
           },
+          "SAS.flowConversionMode": {
+            "order": 0,
```
scnwwu, line 124: just noticed this, I guess it's not needed to force the order of this option.

scnwwu:
> > Thank you for your comments and suggestions to enhance the code.
> > Studio Flows are only available for recent versions of SAS Viya, I would like to restrict this feature to newer versions of SAS Viya only. How would you suggest to proceed ?
> > I see there is a private API Call to get the ViyaCadence in `ContentNavigator/ContentModel.ts` and I would like to use that output in `ContentNavigator/utils.ts` ?
> 
> My opinion is that we should use `viyaCadence` as minimally as possible. What happens if you try to create a flow for an instance that doesn't support it? I'm wondering if it returns an error response we could interpret and respond accordingly?
> 
> @scnwwu may have other thoughts on this

I guess it depends on if we can tell from the error that whether it's not supported at all or just temporarily not available.

client/src/components/ContentNavigator/index.ts
sahbic: done in 33035265761fa800663eeb3c9a709e589a66d92a

client/src/components/ContentNavigator/convert.ts
sahbic: I have added some tests in eec49d4ec0f52dc16f9836aebfd4b92839892cfa.
Please let me know if this is the kind of tests you have in mind ?

Tests:
- Converting SAS Notebooks to code lists: multi cell + single cell
- Converting code lists to Flows: swimlane mode + node mode / multi cell + single cell

Some of the challenges encountered:
- handling special characters before parsing
- removing unique ids, timestamps and dates from comparison

sahbic:
Concerning the behavior for older versions of SAS Viya (tested on 3.5), we have an error when trying to create a new SAS Studio session: `Request failed with status code 404`
Right now, the user sees an error message `Error converting the .sasnb file to .flw format.`but a .flw file is still created. (but cannot be opened of course)
I also see an error `fail to retrieve the viya cadence`  in the logs when using SAS Viya 3.5

client/test/components/notebook/convert.test.ts
```diff
@@ -0,0 +1,250 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
```
scottdover, line 1: The location of this file seems off. If my test lives at `client/test/components/notebook/convert.test.ts`, my expectation would be at `client/src/components/notebook/convert.ts`

client/test/components/notebook/convert.test.ts
```diff
@@ -0,0 +1,250 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+  generateCodeListFromPythonNotebook,
+  generateCodeListFromSASNotebook,
+  generateFlowDataNode,
+  generateFlowDataSwimlane,
+} from "../../../src/components/ContentNavigator/convert";
+import * as assert from "assert";
```
scottdover, line 10: Maybe just import `{ equal }` here

client/test/components/notebook/convert.test.ts
```diff
@@ -0,0 +1,250 @@
...
+
+const contentSASNotebook1 = `[{"language": "python","value": "a, b = 4, 2\r\nprint('Result: ', a*10 + b)"},{"language": "sas","value": "data work.prdsale;\r\n\tset sashelp.PRDSALE;\r\nrun;\r\n\r\nproc means data=work.prdsale;\r\nrun;"},{"language": "markdown","value": "# test"},{"language": "sql","value": "CREATE TABLE WORK.QUERY_PRDSALE AS\r\n    SELECT\r\n        (t1.COUNTRY) LABEL='Country' FORMAT=$CHAR10.,\r\n        (SUM(t1.ACTUAL)) FORMAT=DOLLAR12.2 LENGTH=8 AS SUM_ACTUAL\r\n    FROM\r\n        WORK.PRDSALE t1\r\n    GROUP BY\r\n        t1.COUNTRY"}]`;
```
scottdover, line 12: It may be more readable to start with these being json objects, and stringify as needed for tests. And then we can use backticks for strings and use tab/newlines/etc as needed. So we could do something like...

```javascript
const contentSASNotebook1 = [{
	"language": "python",
	"value": `a, b = 4, 2
print('Result: ', a*10 + b)`
}, {
	"language": "sas",
	"value": `data work.prdsale;
		set sashelp.PRDSALE;
run;

proc means data=work.prdsale;
run;`
}, {
	"language": "markdown",
	"value": "# test"
}, {
	"language": "sql",
	"value": `CREATE TABLE WORK.QUERY_PRDSALE AS
SELECT
(t1.COUNTRY) LABEL='Country' FORMAT=$CHAR10.,
(SUM(t1.ACTUAL)) FORMAT=DOLLAR12.2 LENGTH=8 AS SUM_ACTUAL
FROM
WORK.PRDSALE t1
GROUP BY
t1.COUNTRY`
}];
```

and that would change
```javascript
contentSASNotebook1
        .replace(/\n/g, "\\n")
        .replace(/\t/g, "\\t")
        .replace(/\r/g, "\\r"),
```

to something more like `JSON.stringify(contentSASNotebook1)` (newlines/tabs/etc should be properly encoded without replacements)

client/test/components/notebook/convert.test.ts
```diff
@@ -0,0 +1,250 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+  generateCodeListFromPythonNotebook,
+  generateCodeListFromSASNotebook,
+  generateFlowDataNode,
+  generateFlowDataSwimlane,
+} from "../../../src/components/ContentNavigator/convert";
```
scottdover, line 9: In general, we shouldn't be exporting functions from files _just_ for the purposes of testing. In this case, I'm wondering if we can test all the things through `convertNotebookToFlow` since that is the only function that is used in non-test code.

sahbic: done in 5f3a455e8798a4a89997b40e894f660e63b00135

client/test/components/notebook/convert.test.ts
sahbic: done in 5f3a455e8798a4a89997b40e894f660e63b00135

client/test/components/notebook/convert.test.ts
sahbic: done in 5f3a455e8798a4a89997b40e894f660e63b00135

client/test/components/ContentNavigator/convert.test.ts
```diff
@@ -0,0 +1,466 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { convertNotebookToFlow } from "../../../src/components/ContentNavigator/convert";
+import { deepEqual } from "assert";
+import { workspace } from "vscode";
+
+const contentSASNotebook1 = [
```
scnwwu, line 8: nit: just providing an idea, will it look better to make separate files in `client/testFixture` directory.
Say `test.sasnb`, `test.ipynb`, `test.step`, `test.flw` etc. And read these files in this test.
The test fixture files can be opened in real world applications thus indicate it tests real world files and make this test file cleaner.

client/test/components/ContentNavigator/convert.test.ts
```diff
@@ -0,0 +1,466 @@
...
+
+async function updateWorkspaceSettings(Mode) {
+  const configuration = workspace.getConfiguration(); // Get the workspace configuration
+  await configuration.update("SAS.flowConversionMode", Mode, true);
+}
+
+describe("Notebook Conversion", async () => {
+  it("convert the multi cell sas notebook to node flow", async () => {
+    await updateWorkspaceSettings("Node");
+    const [flowDataNode, flowData1Node] = convertAndRemoveIds(
```
scnwwu, line 402: nit: (applies to variable names in many places) would it be better to name them with `expected` and `actual` to make them cleaner?

client/src/components/ContentNavigator/convert.ts
```diff
@@ -0,0 +1,349 @@
...
+function generateCodeListFromPythonNotebook(content: string): Entry[] {
+  const codeList = [];
+  try {
+    const notebookContent = JSON.parse(content);
+
+    for (const cell of notebookContent.cells) {
+      const code = cell.source.join("");
+      const cellType = cell.cell_type;
+      if (code !== "" && cellType === "code") {
+        const language = "python";
```
scnwwu, line 322: We have [sas_kernel](https://github.com/sassoftware/sas_kernel) for Jupyter. Is it possible to have sas here?

scnwwu:
> Concerning the behavior for older versions of SAS Viya (tested on 3.5), we have an error when trying to create a new SAS Studio session: `Request failed with status code 404` Right now, the user sees an error message `Error converting the .sasnb file to .flw format.`but a .flw file is still created. (but cannot be opened of course) I also see an error `fail to retrieve the viya cadence` in the logs when using SAS Viya 3.5

It sounds like we could try create Studio session at the start of the conversion, and if any error, show user something like "SAS Studio service not available."?

client/src/components/ContentNavigator/convert.ts
sahbic: done in 5d9b2307a23ae99f5bed029eb98f4a11face2a07

client/test/components/ContentNavigator/convert.test.ts
sahbic: done in 4b45ea3cd0c6ebd1a21ab73e963182ae250eb717

client/test/components/ContentNavigator/convert.test.ts
sahbic: done in 4b45ea3cd0c6ebd1a21ab73e963182ae250eb717

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -322,19 +325,67 @@ export class ContentModel {
...
+    }
+  }
+
+  public async associateFlowFile(
+    name: string,
+    uri: Uri,
+    parent: ContentItem,
+  ): Promise<string | undefined> {
+    try {
+      const sessionId = await createStudioSession(this.connection);
```
scnwwu, line 378: oh, is it possible to re-use the session from testStudioConnection?

scnwwu (commented):
Thank you for the great work! Looks good to me. Just one minor comment.

client/src/components/ContentNavigator/ContentModel.ts
sahbic: done in 3eb7d56f1e76ea8d93fc8d3de5ba2f866b97008e

sahbic:
I have added translations to German and Chinese using Google Translate.. This can be enhanced later by fluent speakers.

scnwwu:
Looks good to me (including the Chinese translation:smile:).
Will let @scottdover finally vet and approve. Thanks!

scottdover (approved):
After a rebase (to address package.json issues), I think this is good to go. Great work, @sahbic !

package.nls.de.json
```diff
@@ -6,6 +6,7 @@
   "commands.SAS.authorize": "Anmelden",
   "commands.SAS.close": "Schließen der aktuellen Verbindung",
   "commands.SAS.collapseAll": "Alles einklappen",
+  "commands.SAS.convertNotebookToFlow": "In Flow konvertieren...",
```
scottdover, line 9: Hey @sahbic. Can you revert this changes in this file for now? We have someone who is able to translate things into German, and having these left out will make it easier to see which ones to translate in the future (i.e. since our build process points these out)

sahbic: Should I also revert the changes to `l10n/bundle.l10n.de.json` ?

scottdover: Yes, please. Sorry I didn't mention that one

sahbic: done in 58a2508d839cc91690c8d15be92340dac3a29561

scottdover (approved):
Looks good to me :). Thank you again for working on this one!

scottdover:
Just a heads-up. I updated this branch to clear up the translation check failures. Nothing else should've changed here :)

Zhirong2022:
Considering to generate the Flow nodes and the node name associates with the language which the user selects through VSC 'Configure Display Language'
image

Zhirong2022:
A separate issue has been raised to track the feature updates.  Close this one to go to next stage.

scnwwu:
@sahbic, please resolve conflicts then we're going to merge it. Thank you!

clangsmith:
Disclaimer: While you can inspect the SAS Studio Flow JSON format (for example, by opening a .flw file in an editor), it is not a publicly documented or publicly supported format/schema. It is still internal/proprietary. We strive to keep the Flow format backwards compatible, but there is a possibility that future changes to the internal model could break this “convert SAS Notebook to SAS Studio Flow” feature in VS Code, requiring additional changes to the SAS extension to fix this feature. In other words, when you are using the Flow format/schema directly (as in this feature), you are using it as-is and at your own risk.

PR #473, opened 2023-08-24
feat: provide an easy way for users to specify the ods style of the results and automatically use a style that reasonably matches the vs code theme by default

clangsmith:
**Summary**
Added a new option to allow users to specify the ODS style to use with HTML5 results. User can specify "auto" to have the extension pick an ODS style that matches the VS Code color theme as closely as possible, "server default" to instruct the extension not so specify a style (in which case the default style configured on the server is used), or user can explicitly specify a style by selecting one in the option UI combobox or type one in the settings.json file if it doesn't exist in the option UI combobox.

**Testing**
I submitted SAS programs and SAS notebooks with various values for the new option, various color themes for the VS Code app, ensuring (visually and in the log) the appropriate ODS style was used for each scenario.


package.json
```diff
@@ -120,6 +120,23 @@
...
+          "SAS.results.html.style": {
+            "order": 1,
+            "type": "string",
+            "default": "(auto)",
+            "enum": [
+              "(auto)",
+              "(server default)",
+              "HighContrast",
+              "Ignite",
+              "Illuminate"
```
clangsmith, line 132: I'm waiting to hear back from others on which of the ODS styles we should include here in the option UI. I'm thinking "Ignite", "Illuminate", and "HighContrast" at a minimum. I ran the SAS code "proc sql; select * from sashelp.vstyle; quit;" on latest Viya 4 (2023.09) using SAS Studio and there were 60 ODS styles were defined. I was thinking we should include a subset of the most popular rather than trying to include all, since it would inevitably get out of sync. Fortunately, a user can still use styles not in the option UI by simply specifying it in the settings.json.

package.json
```diff
@@ -120,6 +120,23 @@
             "default": true,
             "description": "%configuration.SAS.session.outputHtml%"
           },
+          "SAS.results.html.style": {
```
clangsmith, line 123: I'm going to submit another PR after this one to change "SAS.session.outputHtml" to "SAS.results.html.enabled" to follow the pattern introducing in this pattern. "results" makes a lot more sense than "session" here, and this new pattern makes potentially adding additional ODS results types in the future easier. Also re-ordering the options in that follow-on PR, so that the connection profiles and certs come first, then the results related options.

clangsmith:
Options UI (configuration) is documented here:
https://code.visualstudio.com/api/references/contribution-points#contributes.configuration

client/src/util.ts
```diff
@@ -0,0 +1,53 @@
...
+          }
+          case ColorThemeKind.Dark: {
+            odsStyle = "Ignite";
+            break;
+          }
+          case ColorThemeKind.HighContrast: {
+            odsStyle = "HighContrast";
+            break;
+          }
+          case ColorThemeKind.HighContrastLight: {
```
scnwwu, line 31: nit: how about `default:`

package.json
```diff
@@ -120,6 +120,23 @@
...
+              "(auto)",
+              "(server default)",
+              "HighContrast",
+              "Ignite",
+              "Illuminate"
+            ],
+            "enumDescriptions": [
+              "%configuration.SAS.results.html.style.(auto)%",
+              "%configuration.SAS.results.html.style.(server default)%"
+            ],
```
clangsmith, line 137: We can also use "enumItemLabels" here if we want to provide translatable enum labels. It may be appropriate, at least for the "(auto)" and "(server default)" values? If we do that, it displays both the label and the enum value, so it would look something like this:
image
I noticed a lot of other options in VS Code use "auto", not in parens, and apparently not translated.

clangsmith: Since "HighContrast" is the actual name of the style used in SAS code (as well as the other explicit styles listed), I would expect those to be translated.

client/src/util.ts
```diff
@@ -0,0 +1,53 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
```
scnwwu, line 1: There's currently a `utils.ts` file under `components/` directory. This introduces a question about the positioning for the 2 utility files. I guess the SAS code utility will be larger moving forward thus deserves a separate file.
So how about name it something like `sasCodeUtil.ts` and put it into the `components/` directory?

client/src/components/notebook/Controller.ts
```diff
@@ -133,13 +134,9 @@ const getCode = (doc: vscode.TextDocument) => {
   } else if (doc.languageId === "python") {
     code = wrapPython(code);
   }
-  return outputHtml ? wrapHTML5(code) : code;
+  return outputHtml ? wrapCode(code) : code;
```
scnwwu, line 137: nit: as `wrapCode` will take care of the outputHtml option, how about simply call `wrapCode` so that we don't need to get settings in this file?

package.json
clangsmith: It is also worth noting that any style selected here will be used in the style= option in the ODS statement regardless of whether the code is being submitted in Viya 4, Viya 3.5, or V9. So, if the style is not defined, you’ll get a warning in the log and a fallback style will be used. For example, “WARNING: Style BLAH not found; HTMLEncore style will be used instead.”

scnwwu: A problem for enum is that it will produce warnings if user wants to specify a style in the settings.json file.

WARNING

clangsmith: Yes, I did notice that. It still works, but yes, the warning could be alarming. I initially looked for a combobox where user could type in a value if the desired value did not already exist in the combobox, but it doesn't look like vscode supports that kind of control. Another alternative would be to add another string option, in addition to the combobox, which would allow user to type in any desired style. If they typed a value in this "override HTML style" option, then it would override any value specified in the combobox. Wasn't crazy about needing two UI options for user to specify a style, but could run this by @skyeaussie and @crlewi.

client/src/util.ts
clangsmith: If window.activeColorTheme.kind does not match one of the cases/enums for ColorThemeKind, then odsStyle will remain "" (what is initialized to), thus no style= statement will be generated and the server default style will be used. This seems reasonable to me. I'll go ahead and add default though, so it's more explicit and readable.

client/src/components/notebook/Controller.ts
clangsmith: Ah, good point. I'll make that change, thanks!

client/src/util.ts
clangsmith: Good point. Sure, I'll move it into components/ directory. What about if we name it "utils_sasCode.ts", so that any "utils*" files will appear adjacent in explorer?

package.json
clangsmith: Added a number of other styles recommended by @cleester4, so here is the list we are going to go with:
Aqua, Daisy, Dove, HighContrast, HighContrastLarge, HTMLBlue, HTMLEncore, Ignite, Illuminate, Journal, Journal2, Journal3, Marine, Midnight, Moonflower, Opal, Pearl, Sapphire

The list above are all the styles defined on Viya 4 by default, except for these (which are ideally for other ODS destinations or otherwise don't fit HTML5 well):
Default, EGDefault, Excel, ExcelAqua, ExcelHighContrast, ExcelIgnite, ExcelIlluminate, ExcelMarine, ExcelMidnight, ExcelOpal, ExcelUmstead, Listing, PearlJ, Plateau, PowerPointAqua, PowerPointDark, PowerPointHighContrast, PowerPointIgnite, PowerPointIlluminate, PowerPointLight, PowerPointMarine, PowerPointMidnight, PowerPointOpal, PowerPointUmstead, Printer, Raven, Rtf

As noted above, any styles (built-in or user-defined) can still be used by explicitly specifying in the settings.json file. For example...
"SAS.results.html.style": "EGDefault"

For now, not going to worry about the warning that appears when you specify a value in the settings.json that does not correspond with an enum, since it is just a warning and still works.

client/src/util.ts
clangsmith: I moved and renamed it to components/utils_sasCode.ts.

client/src/components/utils_sasCode.ts
```diff
@@ -0,0 +1,57 @@
...
+          case ColorThemeKind.HighContrast: {
+            odsStyle = "HighContrast";
+            break;
+          }
+          case ColorThemeKind.HighContrastLight: {
+            odsStyle = "Illuminate";
+            break;
+          }
+          default: {
+            odsStyle = "";
```
scottdover, line 36: Do we need `odsStyle = "";` here since we started with an empty string up top?

client/src/components/utils_sasCode.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
```
scottdover, line 1: I didn't realize we already have https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/utils.ts . Can we add this function there? 

Or, if we want to keep these files more tiny/focused, then maybe we create `utils/throttle.ts` (to replace `utils.ts`) and `utils/sasCode.ts`?

client/src/components/utils_sasCode.ts
clangsmith: I didn't have the default section there initially for that reason. I added it at @scnwwu's suggestion. Perhaps a little more readable.

client/src/components/utils_sasCode.ts
```diff
@@ -0,0 +1,57 @@
...
+            break;
+          }
+          default: {
+            odsStyle = "";
+            break;
+          }
+        }
+        break;
+      case "(server default)":
+        odsStyle = "";
```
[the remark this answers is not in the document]
clangsmith: I mainly put it there for readability, so every case would be handled by the switch.

client/src/components/utils_sasCode.ts
```diff
@@ -0,0 +1,57 @@
...
+  const htmlStyle: string = workspace
+    .getConfiguration("SAS")
+    .get("results.html.style");
+
+  if (outputHtml) {
+    let odsStyle = "";
+    switch (htmlStyle) {
+      case "(auto)":
+        switch (window.activeColorTheme.kind) {
+          case ColorThemeKind.Light: {
```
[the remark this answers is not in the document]
clangsmith: Yeah, I'm not a fan of that either. Will remove, thanks.

client/src/components/utils_sasCode.ts
clangsmith: Either way is fine with me. I'll move it into existing utils.ts then, since it isn't very big currently. Can always refactor later if it does get too large.

scottdover (approved):

Zhirong2022:
1. The ODS html style works as expected.
2. The ODS html style is shown properly in log.

Zhirong2022 (approved):

Issue #356, opened 2023-06-19
Can't use `$` in local COM


ShuguangSun:
**Describe the bug**
`format make $10.;` will be submitted as `format make .;`. It seems `$10` will be omitted.

**Steps to reproduce**

1. Create a local COM profile
2. Create a SAS file with 

```sas
proc print data=sashelp.cars; 
    format make $10.;
run;
```

3. run sas code

```
6          proc print data=sashelp.cars;
7              format make .;
                           _
                           22
                           200
ERROR 22-322: Syntax error, expecting one of the following: a name, a format name, ;, -, :, _ALL_, _CHARACTER_, _CHAR_, _NUMERIC_.  
ERROR 200-322: The symbol is not recognized and will be ignored.
```

**Environment (please complete the following information):**
Client OS: windows 2019
Extension version: 1.0.0
SAS version: 9.4M7 from OA

smorrisj:
The $ symbol in powershell are a reserved symbol. I suspect that the code string will need to replace this with escaped versions in the $code parameter (script.ts line 76). It would probably be good to add code to the powershell script Run function to do this before passing the code string into SAS.LanguageService.Submit

smorrisj:
If the above doesnt work, could also consider adding it to the com connection provider itself:

`    //write an end mnemonic so that the handler knows when execution has finished
    const codeWithEnd = `${codeWithODSPath}\n%put ${endCode};`;
    const codeToRun = `$code=@"\n${codeWithEnd}\n"@\n`;`

codeToRun could be augmented to do the escaping in addition.

sasutils:
A better title would be cannot use $ in SAS code.  It is "eating" any token that starts with $.

ShuguangSun:
`%put "ab'c";` not work as well.

ShuguangSun:
> If the above doesnt work, could also consider adding it to the com connection provider itself:
> 
> ```ts
>  //write an end mnemonic so that the handler knows when execution has finished
>  const codeWithEnd = `${codeWithODSPath}\n%put ${endCode};`; 
>  const codeToRun = `$code=@"\n${codeWithEnd}\n"@\n\`;  
> ```
> 
> codeToRun could be augmented to do the escaping in addition.

It seems breaking the quotation.

I also tried 
```ts
const codeToRun = `$code=@'\n${codeWithEnd}\n'@\n`;  
```
It can solve the `$` issue.

But it has the problem with `'` sometimes. 
[...]

PR #474, opened 2023-08-25
fix(core): escape $ symbol in powershell code for local connection

ShuguangSun:
Signed-off-by: Shuguang Sun <shuguang79@qq.com>

**Summary**
See https://github.com/sassoftware/vscode-sas-extension/issues/356 . 

**Testing**
Not provide test code. But tested below scenarios:

- `$` symbol
- `%put "ab'c";`

### DIFF
--- a/client/src/connection/com/index.ts
+++ b/client/src/connection/com/index.ts
@@ -8,8 +8,7 @@ import { BaseConfig, RunResult } from "..";
 import { scriptContent } from "./script";
 import { Session } from "../session";
 
-const beginningCode = `\`'--vscode-sas-extension-submit-begining--\`'`;
-const endCode = `\`'--vscode-sas-extension-submit-end--\`'`;
+const endCode = "--vscode-sas-extension-submit-end--";
 let sessionInstance: COMSession;
 
 /**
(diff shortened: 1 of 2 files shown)


client/src/connection/com/index.ts
```diff
@@ -118,8 +119,8 @@ export class COMSession extends Session {
       );
 
       //write an end mnemonic so that the handler knows when execution has finished
-      const codeWithEnd = `${codeWithODSPath}\n%put ${endCode};`;
-      const codeToRun = `$code=@"\n${codeWithEnd}\n"@\n`;
+      const codeWithEnd = `%put ${beginningCode};\n${codeWithODSPath}\n%put ${endCode};`;
+      const codeToRun = `$code=@'\n${codeWithEnd}\n'@\n`;
```
scnwwu, line 123: A Here-string in PowerShell should start with a new line.
```suggestion
      const codeToRun = `$code=\n@'\n${codeWithEnd}\n'@\n`;
```
Then no need for extra ``` `' ```

ShuguangSun: you are right. It works.

scottdover:
Looks like there are some test failures that need to be resolved for this one. Otherwise, lgtm

Zhirong2022:
Symbol '$' can be recognized correctly.

Zhirong2022 (approved):

PR #487, opened 2023-09-04
fix: correctly re-open files for content rename

### LINKED ISSUE #445: SASNotebook file contents were not shown properly if trying to rename the file name
**Describe the bug**
Install the SAS extension base on the official release. Create a new .sasnb file, the file contents were not shown properly if trying to rename the file name.

**Steps to reproduce**
1. Create a new SASNotebook file
2. Input something and save
3. Rename the file name 

**Expected behavior**
The contents of SASNotebook file were shown properly 

**Screenshots**
renameSASNotebookFile

Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.1.0]

scnwwu:
**Summary**
Fix #445

- File should be opened by "vscode.open" command instead of `window.showTextDocument` for non-text document.
- The dirtyFiles check is not needed. The DataProvider will `closeFileIfOpen` - native way to prompt user if they want to save.
- We should only re-open the file if it was opened before rename.
- The type for `closeFileIfOpen` is not exact, it should be `true | Thenable<boolean>`. We don't need to write it, TS will infer correctly.

**Testing**
Test case in #445

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -28,6 +28,7 @@ import {
   Uri,
   l10n,
   window,
+  commands,
 } from "vscode";
 import { profileConfig } from "../../commands/profile";
 import { SubscriptionProvider } from "../SubscriptionProvider";
@@ -244,12 +245,18 @@ class ContentDataProvider
     item: ContentItem,
     name: string,
   ): Promise<Uri | undefined> {
-    if (!(await closeFileIfOpen(item))) {
+    const closing = closeFileIfOpen(item);
+    if (!(await closing)) {
       return;
     }
     const newItem = await this.model.renameResource(item, name);
     if (newItem) {
-      return getUri(newItem);
+      const newUri = getUri(newItem);
+      if (closing !== true) {
+        // File was open before rename, so re-open it
+        commands.executeCommand("vscode.open", newUri);
+      }
+      return newUri;
     }
   }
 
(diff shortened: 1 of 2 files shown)


client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -244,12 +245,18 @@ class ContentDataProvider
     item: ContentItem,
     name: string,
   ): Promise<Uri | undefined> {
-    if (!(await closeFileIfOpen(item))) {
+    const closing = closeFileIfOpen(item);
+    if (!(await closing)) {
```
scottdover, line 249:
```suggestion
    const closing = await closeFileIfOpen(item);
    if (!closing) {
```

scnwwu: Well I intended to get the variable before promise resolve, so that I can tell if the file was open at line 255.

scottdover: I don't think I understand

scnwwu: If I wrote `const closing = await closeFileIfOpen(item);`, then if  `closing === true`, there're 2 possibilities.
One is that the file was not open, the `closeFileIfOpen` function directly returned true, in this case, it should not re-open the file.
The other is, the file was open and the user confirmed save or not save and closed the file, the function returned a promise and resolved to true, in this case, it should re-open the file.
I can't distinguish the 2 cases if I only get awaited true.

Zhirong2022 (approved):

PR #495, opened 2023-09-07
feat: improve translation check

scottdover:
**Summary**
This makes a few change to how we check for translations:
 - It splits this into a distinct workflow that is ran on a push to `main` (instead of checking for every pull request action)
 - It also adds a `workflow_dispatch` trigger such that we can manually run this check with any branch
 - It updates the output to only list the files that need to be updated for translations. See the two examples below

Example with no translations needed:
Screenshot 2023-09-07 at 9 08 29 AM

Example with translations needed: 
Screenshot 2023-09-07 at 9 08 09 AM

**Testing**
 - [x] Made sure github actions passed
 - [x] Temporarily added this branch to the translation check
 - [ ] Will test `workflow_dispatch` on main (workflow_dispatch triggers have to be in the default branch before they can be triggered via actions)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -19,14 +19,3 @@ jobs:
       - run: xvfb-run -a npm run test
         if: runner.os == 'Linux'
       - run: npx @vscode/vsce package
-  check-translations:
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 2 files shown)


scnwwu (approved):
Looks great to me!

PR #496, opened 2023-09-07
feat: change output html option to more logical/consistent ID and re-order options

clangsmith:
**Summary**
Enabling HTML output has nothing to do with "session"; so, changed the ID to results.html.enabled which makes more sense, follows pattern with results.html.style, and easier to extend to other results formats in the future. Also, I think it makes more sense for the connection profiles (and related certificates) to appear first, since that is usually the first thing you do when getting going with the extension.

**Testing**
- Ensured the extension settings are now in the desired order. (Note: When you open the extension settings from the extension in the Extensions pane, it looks like the settings are in alphabetical order rather than the specified order. To see the specified order, type "sas" in the search field or select Extensions->SAS in the settings tab.)
- Ensure the "HTML: Enabled" and "HTML: Style" settings still work as desired.

### DIFF
--- a/client/src/commands/run.ts
+++ b/client/src/commands/run.ts
@@ -116,7 +116,7 @@ async function runCode(selected?: boolean, uri?: Uri) {
 
   const outputHtml = !!workspace
     .getConfiguration("SAS")
-    .get("session.outputHtml");
+    .get("results.html.enabled");
 
   const code = getCode(selected, uri);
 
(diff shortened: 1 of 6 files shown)


scottdover (approved):

scnwwu:
I'm not against this PR. It just brings a question about do we plan to set explicit order for all settings in future? Or do we only want to make the connectionProfiles and certificates first and leaving others in natural order unless a specific reason?

clangsmith:
> I'm not against this PR. It just brings a question about do we plan to set explicit order for all settings in future? Or do we only want to make the connectionProfiles and certificates first and leaving others in natural order unless a specific reason?

I think order should be considered. If options are related somehow, I'd certainly expect them to be adjacent, rather than appear in the order in which they were added.

Zhirong2022:
The items order at Settings are shown as expected.

Issue #329, opened 2023-06-09
Dragging a file from SAS Content into a SAS program should result in a filename statement for easy programmatic access


clangsmith:
Requested/suggested by @boemska-nik, and I really like the idea!

**Is your feature request related to a problem? Please describe.**
No, it would just be a nice additional convenience feature. Rather than having to remember the syntax to reference a file in SAS Content and type a full filename statement, it would be convenient to be able to just drag the file into a SAS program and have the syntax automatically inserted.

**Describe the solution you'd like**
Similar to the way you can currently drag a table from the Libraries pane into a SAS program and it results in "<library>.<member>" text being inserted into the program (to easily reference that table programmatically), it would be nice/convenient to be able to drag any file from the Explorer pane (SAS Content) into a SAS program and for that to result (when dropped) in filename statement code (with FILESRVC access method, since Files Service) being inserted in the SAS program at the spot it was dropped.

To verify:

- Run the following SAS code in VS Code to create a sample .csv in your deployment:
`/* Write SASHELP.CLASS to a .csv in SAS Content */
filename dest filesrvc folderpath='/Public/' filename='blah.csv';
proc export data=sashelp.class dbms=csv file=dest;
run;` 
- In the Explorer pane, click the Refresh icon an confirm the file blah.csv appears in the SAS Content\Public folder.
- Then, copy this code into a new SAS program:
`data _null_;
  infile myfile;
  input;
  put _infile_;
run;`
- Drag the blah.csv file from Explorer pane and drop it on the top line of the SAS program in the previous step.
- Ensure a filename statement similar to the following is inserted at that spot in the SAS program, where the folderpath corresponds to the folder location of the file and the filename corresponds to the name of the file:
`filename myfile filesrvc folderpath='/Public/' filename='blah.csv';`
- Run the program to make sure it runs successfully. (The contents of the .csv should be printed to the log in this case.)

**Describe alternatives you've considered**
Users can just continue typing the whole thing out themselves and have to remember the syntax.

**Additional context**
It would be nice for dragging a library from Libraries pane to result in "<library>" being inserted into the code as well, but we can track that separately.

**Environment**
Anywhere the Explorer pane (SAS Content, aka Files Service) is supported -- currently Viya 4 and Viya 3.5

Issue #22, opened 2022-07-20
honor Caps Lock status when autocompleting


KDeser:
I'm oldschool and like to write SAS code in ALL CAPS. 

Can the autocomplete abide by this? If the user has Caps Lock on, autocompleted code should be entered in ALL CAPS.

mail2fei:
We will evaluate the feature to see the feasibility and cost in the extension.

scnwwu:
An update: VS Code said they won't do in client side.
https://github.com/microsoft/vscode/issues/156125

PR #510, opened 2023-09-14
feat: allow dragging sas content into editor

scottdover:
**Summary**
This adds the ability to drag sas content into sas programs. See https://github.com/sassoftware/vscode-sas-extension/issues/329 for more details

**Testing**
 - [x] Tested dragging a folder into a SAS program (should not provide any code)
 - [x] Tested dragging a file into a SAS program from my folder
 - [x] Tested dragging a file into a SAS program from my favorites
 - [x] Tested dragging files into a SAS program from nested folder paths
 - [x] Tested dragging two files from the same folder into a SAS program


client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -138,8 +148,31 @@ class ContentDataProvider
...
+      return undefined;
+    }
+
+    const fileFolderPath = await this.model.getFileFolderPath(contentItem);
+    if (!fileFolderPath) {
+      return undefined;
+    }
+
+    return {
+      insertText: `filename myfile filesrvc folderpath='${fileFolderPath}' filename='${contentItem.name}';\n`,
```
scnwwu, line 170: nit: just put a note here that we have customers who like uppercase as #22
I don't know is it a good idea to insert uppercase if the word before or after the position is uppercase?
I'll be OK with simply moving forward and revisit if customer reports.

scnwwu (approved):

client/src/components/ContentNavigator/ContentDataProvider.ts
scnwwu: oh, another thing is that is `myfile` good enough? So if user drop 2 files they have to change the name. Or how about simply using `${contentItem.name}`?

scottdover: Good question. I originally copied from the issue w/o _really_ thinking this part (the name) through. Will take a look. As for uppercase or lowercase, I'm inclined to keep it as-is because it seems like the simplest approach. But yeah, if a customer reports it, lets adjust :)

scottdover: Addressed in https://github.com/sassoftware/vscode-sas-extension/pull/510/commits/41a228c36f53edb8f8cdd803f1a7412e3451e9a8

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -138,8 +148,36 @@ class ContentDataProvider
...
+    }
+
+    const fileFolderPath = await this.model.getFileFolderPath(contentItem);
+    if (!fileFolderPath) {
+      return undefined;
+    }
+
+    const filename = (parse(contentItem.name).name || contentItem.name).replace(
+      /[^a-zA-Z0-9]/g,
+      "",
```
scnwwu, line 171: Well, there's a problem with this. If the item name is pure Chinese, say `中文.sas`, it will result in no name.

scottdover: That's a good point. What would you recommend in that instance? We could always do something like...
```
`file${Date.now()}`
```

That said, `file1695040695541` isn't the most readable variable name.

Or, we could start with a variable name something like `csvfile1`, and increment until we get one that _isn't_ in the document already? Maybe I'll mess around with that approach a bit.

I'll also look into the capitalization bit you mentioned previously. How would you make that determination? Could we look at the document, check if it has uppercase words, and use uppercase in that instance. Otherwise, default to lowercase?

scottdover: I added a different approach and more tests, here - https://github.com/sassoftware/vscode-sas-extension/pull/510/commits/6715b3aed01dd84acb56838dca7a8d5244250864

client/src/components/ContentNavigator/utils.ts
```diff
@@ -106,3 +108,64 @@ export const isItemInRecycleBin = (item: ContentItem): boolean =>
...
+  ) {
+    /* empty */
+  }
+
+  return `${partialFileName}${idx}`;
+};
+
+// A document uses uppercase letters _if_ there is more than one
+// word (where word means gte 3 characters) that is all uppercase
+const documentUsesUppercase = (documentContent: string) =>
```
scnwwu, line 152: Well, if you try with the `client/testFixture/SampleCode.sas`, it will insert uppercase. Because there're datalines with uppercase but not code.
Looks challenging.
How about only do uppercase if not found lowercase word?

client/src/components/ContentNavigator/utils.ts
```diff
@@ -106,3 +108,64 @@ export const isItemInRecycleBin = (item: ContentItem): boolean =>
...
+): string => {
+  const filePieces = parse(contentItemName);
+  const partialFileName = (
+    filePieces.name.replace(/[^a-zA-Z0-9]/g, "") ||
+    `${filePieces.ext.replace(".", "")}file`
+  ).toLocaleLowerCase();
+
+  let idx = 0;
+  const lowercaseDocumentContent = documentContent.toLocaleLowerCase();
+  while (
```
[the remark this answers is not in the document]
scnwwu: Sounds good to me (yeah, the shorter, the better:smile:)

client/src/components/ContentNavigator/utils.ts
```diff
@@ -106,3 +108,64 @@ export const isItemInRecycleBin = (item: ContentItem): boolean =>
...
+ *
+ * @returns string our new variable name.
+ */
+const extractVariableName = (
+  contentItemName: string,
+  documentContent: string,
+): string => {
+  const filePieces = parse(contentItemName);
+  const partialFileName = (
+    filePieces.name.replace(/[^a-zA-Z0-9]/g, "") ||
```
scnwwu, line 133: underscore is also valid, isn't it?

client/src/components/ContentNavigator/utils.ts
scottdover: Updated in https://github.com/sassoftware/vscode-sas-extension/pull/510/commits/1e445783cf30d8ac6246174bf0b17a6bc8f5b209

client/src/components/ContentNavigator/utils.ts
scottdover: Yep! Updated in https://github.com/sassoftware/vscode-sas-extension/pull/510/commits/1e445783cf30d8ac6246174bf0b17a6bc8f5b209

client/src/components/ContentNavigator/utils.ts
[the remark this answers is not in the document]
scottdover: I got rid of the uniqueness checks and added a `<hour><minute><second>` suffix to all variable names in https://github.com/sassoftware/vscode-sas-extension/pull/510/commits/1e445783cf30d8ac6246174bf0b17a6bc8f5b209

scnwwu (approved):

Zhirong2022:
It failed to run the generated FILENAME statement if the _fileref_ name having more than 8 bytes.  And _fileref_ name must conform to the rules for SAS names.
Error details:
ERROR: Invalid logical name.
ERROR: Error in the FILENAME statement.

scottdover:
@Zhirong2022 Can you provide an example of an invalid file name? Also, do you know where the "rules for SAS names" might be (like, is it documented somewhere)?

scottdover:
> It failed to run the generated FILENAME statement if the _fileref_ name having more than 8 bytes. And _fileref_ name must conform to the rules for SAS names. Error details: ERROR: Invalid logical name. ERROR: Error in the FILENAME statement.

I scrapped the file name generation part of this in favor of using a tab stop to fill in the fileref. Now, when a user drags a file into an editor, we'll automatically select/place a cursor at fileref where they can rename as needed.

Screenshot 2023-09-26 at 2 40 56 PM

Zhirong2022:
> @Zhirong2022 Can you provide an example of an invalid file name? Also, do you know where the "rules for SAS names" might be (like, is it documented somewhere)?
@scottdover Please visit the SAS help to find the detail. And "rules for SAS names" means 
names can be 1 to 8 characters long, must begin with a letter (A-Z, either uppercase or lowercase) or an underscore (_), and can continue with any combination of numbers, letters, or underscores.
image

Zhirong2022:
Quotation marks mismatch if the file name contains them and it will fail to run the code as well.
image

scottdover:
> Quotation marks mismatch if the file name contains them and it will fail to run the code as well. image

This has been addressed in https://github.com/sassoftware/vscode-sas-extension/pull/510/commits/214d11e674b705234f092ac4aaf9a922aab080a6.

That said, a sas program like the following will still not run:

```
filename fileref2 filesrvc folderpath='/Users/anon/My Folder' filename="test-'file'.csv";
data _null_; infile fileref2; input; put _infile_; run
```

I attempted to find a workable solution by using the file step in studio to produce code. The code it produced is similar to the code above, and both failed to run with this error:
```
ERROR: Physical file does not exist, test-'file'.csv.
```

All this is to say, I don't think there's anything I can do in this PR to address, but I just wanted to bring it up.

Zhirong2022:
> > Quotation marks mismatch if the file name contains them and it will fail to run the code as well. image
> 
> This has been addressed in [214d11e](https://github.com/sassoftware/vscode-sas-extension/commit/214d11e674b705234f092ac4aaf9a922aab080a6).
> 
> That said, a sas program like the following will still not run:
> 
> ```
> filename fileref2 filesrvc folderpath='/Users/anon/My Folder' filename="test-'file'.csv";
> data _null_; infile fileref2; input; put _infile_; run
> ```
> 
> I attempted to find a workable solution by using the file step in studio to produce code. The code it produced is similar to the code above, and both failed to run with this error:
> 
> ```
> ERROR: Physical file does not exist, test-'file'.csv.
> ```
> 
> All this is to say, I don't think there's anything I can do in this PR to address, but I just wanted to bring it up.

Well, it might run into such kind of scenarios if the file/folder name contains quotation marks.  Ideally, it would be perfect to have a statement without any errors. The user can fix it manually.  Normally, we do not suggest to new a file/folder having special characters otherwise it will cause some troubles unexpectedly.

Zhirong2022:
The filepath is not changed accordingly in filename statement in certain case, please follow the steps below to reproduce

1.Create a new file folder named like 'NewFolderTest' under My Folder, new a .sas file named like 'NewFileTest.sas' under 'NewFileFolderTest'. 
2.Delete file folder 'NewFileFolderTest' 
3.Drag and drop NewFileTest.sas into editor to create filename statement, it works as expected
4.Restore the file folder 'NewFileFolderTest' 
5.Drag and drop NewFileTest.sas into editor, the folderpath still pointed to Recycle Bin

scottdover:
> The filepath is not changed accordingly in filename statement in certain case, please follow the steps below to reproduce
> 
> 1.Create a new file folder named like 'NewFolderTest' under My Folder, new a .sas file named like 'NewFileTest.sas' under 'NewFileFolderTest'. 2.Delete file folder 'NewFileFolderTest' 3.Drag and drop NewFileTest.sas into editor to create filename statement, it works as expected 4.Restore the file folder 'NewFileFolderTest' 5.Drag and drop NewFileTest.sas into editor, the folderpath still pointed to Recycle Bin

This was due to a little cache that I put together that wasn't _really_ saving us that much. I've since removed the cache and things _should_ be working as expected now.

Zhirong2022:

Zhirong2022 (approved):

PR #512, opened 2023-09-14
fix: update high contrast data viewer theming

scottdover:
**Summary**
Prior to this change, we checked to see if we were using a dark theme and applied the `ag-alpine-theme-dark` theme. This changes makes it such that we target `ag-alpine-theme-dark` for both dark and high contrast dark themes, and default to the light theme `ag-alpine-theme` in other cases

**Testing**
 - [x] Test locally using light/dark theme
 - [x] Test via vsix install for high contrast

### DIFF
--- a/client/src/webview/DataViewer.css
+++ b/client/src/webview/DataViewer.css
@@ -50,23 +50,27 @@ body,
 
 /* DARK MODE ICONS */
 
-.vscode-dark .header-icon.float {
+.vscode-dark .header-icon.float,
+.vscode-high-contrast .header-icon.float {
   background: url(../../../icons/dark/tableHeaderNumericTypeDark.svg) center
     no-repeat;
 }
 
 .header-icon.time,
 .header-icon.date-time,
-.vscode-dark .header-icon.date {
+.vscode-dark .header-icon.date,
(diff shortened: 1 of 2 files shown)


scnwwu:
I'm not sure if it's what we want. High Contrast theme has accessibility requirement. SAS High Contrast theme aims to meet WCAG 2.1 AAA color contrast ratios.
I'm not sure if `ag-alpine-theme-dark` provides enough contrast to be high contrast.
We have `ag-theme-sas` in SAS products which match all SAS themes. Not sure if we can take it here. With that it will mtach SAS Studio Next.

scottdover:
Yeah, I wasn't quite sure how to approach this with ag grid. The themes used are "high contrast" according to ag grid (here: https://www.ag-grid.com/javascript-data-grid/themes/#provided-themes). That said, it would be great to be able to use the sas theme from nova-commons.

Maybe this is, at least, a better iteration than the one before it

scnwwu (approved):
OK, good to me.

Zhirong2022:
Specify the color theme as 'SAS High Contrast', the 'ag-theme-alpine-dark' will be applied to the table

Zhirong2022 (approved):

PR #513, opened 2023-09-15
feat: option to control the default placement of Result panel

### LINKED ISSUE #62: Provide an option to allow user to control the default placement of Results
There is one thing some user has been trying to do:  To get the results to appear in the SAME tab group as the code.

Provide a setting to allow the result output tab to show up in the same tab group with the code tab

**Additional context**

777376a7-c8b5-49a3-a6a8-c505ef3b9839

scnwwu:
**Summary**
Resolve #62
Added `Display results to the side of the code` setting. If uncheck, it will display Result in the same tab group to the submitted code.
Added `Reuse single panel to display results` setting. If uncheck, it will display new Result panel each time.

**Testing**
By default the behavior is same to current version. Tested combinations of different settings.

### DIFF
--- a/client/src/components/ResultPanel.ts
+++ b/client/src/components/ResultPanel.ts
@@ -34,7 +34,7 @@ export const showResult = (html: string, uri?: Uri, title?: string) => {
       resultPanel.title = title;
     }
     resultPanel.reveal(
-      sideResult ? ViewColumn.Beside : editor.viewColumn,
+      sideResult ? ViewColumn.Beside : editor?.viewColumn,
       true,
     );
   }
--- a/client/src/components/ResultPanel.ts
+++ b/client/src/components/ResultPanel.ts
@@ -12,11 +12,14 @@ let resultPanel: WebviewPanel | undefined;
 export const showResult = (html: string, uri?: Uri, title?: string) => {
   const sideResult = isSideResultEnabled();
   const singlePanel = isSinglePanelEnabled();
+  if (!title) {
+    title = l10n.t("Result");
+  }
 
   if (!singlePanel || !resultPanel) {
     resultPanel = window.createWebviewPanel(
       "SASSession", // Identifies the type of the webview. Used internally
-      title ?? l10n.t("Result"), // Title of the panel displayed to the user
+      title, // Title of the panel displayed to the user
       {
         preserveFocus: true,
         viewColumn: sideResult ? ViewColumn.Beside : ViewColumn.Active,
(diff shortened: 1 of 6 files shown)


client/src/components/ResultPanel.ts
```diff
@@ -0,0 +1,42 @@
...
+  isSinglePanelEnabled,
+} from "./Helper/SettingHelper";
+
+let resultPanel: WebviewPanel | undefined;
+
+export const showResult = (html: string, uri?: Uri, title?: string) => {
+  const sideResult = isSideResultEnabled();
+  const singlePanel = isSinglePanelEnabled();
+
+  if (!singlePanel || !resultPanel) {
```
scottdover, line 16: Am I understanding this code, right? Basically, it's saying:
 - If there is not a panel, create or
 - else, update the current panel title and contents? 

If that's the case, it might be helpful to have something like `const panelExists = !singlePanel || !resultPanel` (my brain was having a bit of trouble understanding it)

scottdover (approved):
This looks great. I only had one super nitpicky thing if you want to address it

scottdover:
Also, is there any way to we can test this behavior (I'm not sure how straightforward it would be to test placement of things/etc)?

client/src/components/ResultPanel.ts
scnwwu: Well, there're 2 conditions.
`!resultPanel` means there's not a panel.
`!singlePanel` means user don't want to reuse. So we'll create no matter panel exists.
`resultPanel` is the panel instance. `singlePanel` is the setting value. Does it make sense?

scnwwu:
> Also, is there any way to we can test this behavior (I'm not sure how straightforward it would be to test placement of things/etc)?

Hmm... I think it's mostly VS Code's behavior (definition of ViewColumn.Active vs. ViewColumn.Beside). We don't need to test it.

Zhirong2022:
Enable 'Reuse single panel to display results' only, Run SAS tasks through terminal. 
It failed to run the task and got 'Cannot read properties of undefined (reading 'viewColumn')'.

scnwwu:
> Enable 'Reuse single panel to display results' only, Run SAS tasks through terminal. It failed to run the task and got 'Cannot read properties of undefined (reading 'viewColumn')'.

Thanks. Fixed in latest code.

Zhirong2022:

Zhirong2022 (approved):

Issue #537, opened 2023-09-28
After the 1.4 release, .ipynb files are opening via the sas-notebook renderer, which leads to errors


smorrisj:
**Describe the bug**
In the 1.4 release, we're incorrectly associating ipynb files to use the sas-notebook component.

**Steps to reproduce**
Preconditions: A vscode system with the SAS extension and Jupyter Extension bundle installed

1. Open any ipynb file.
2. Observe the toast that pops up that indicates that multiple editors have been associated to .ipynb files. Continue opening the file. It will use the sas-notebook component to render the .ipynb file which is not desired behavior.
3. Observe the error: "i.map is not a function".  See screenshot.

**Expected behavior**
We should only associate sasnb files with the sas-notebook component.

**Screenshots**
![ipynb_error](https://github.com/sassoftware/vscode-sas-extension/assets/117842602/0ba29141-793f-4399-bd16-401511b4cd13)

**Additional context**

PR #538, opened 2023-09-28
fix: associate sas-notebook with sasnb only

smorrisj:
**Summary**
Removes conversion of .ipynb to flow. We only want to support converting .sasnb files to flow.
Removes default association of .ipynb to the sas-notebook component since this causes .ipynb files to no longer render when opened.

**Testing**
Steps in #537

### DIFF
--- a/client/src/components/ContentNavigator/convert.ts
+++ b/client/src/components/ContentNavigator/convert.ts
@@ -309,26 +309,6 @@ QUIT;`;
   return codeList;
 }
 
-function generateCodeListFromPythonNotebook(content: string): Entry[] {
-  const codeList = [];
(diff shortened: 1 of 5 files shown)


scottdover (approved):

scottdover (approved):

dahils:
Approving.  
With this fix, I successfully opened an .ipynb (Jupyter notebook) file with no errors,
I was still able to open a .sasnb (SAS Notebook) file,
and after dropping a local-file .sasnb onto My Folder in the SAS content tree, I was able to access the Convert to Flow context-menu item and successfully made a .flw file from the .sasnb.

PR #550, opened 2023-10-04
feat: add upload/download for sas content

scottdover:
**Summary**
This makes a few changes around uploading/downloading files in the sas extension:
 - This adds an `Upload` context menu item to SAS content: This feature allows a user to upload one or more files to sas content
 - This adds a `Download` context menu item to SAS content: This allows you to download one or more files or folders, and specify where to save them 
   - **NOTE**: This doesn't allow you to specify file name when you are specifying a single file to download. This is to make it such that we have consistency between downloading a single file or multiple files. The single file will be persisted with the same name from SAS content 
 - This adds a `Download` context menu item to the results pane. Clicking this will allow the user to save the results as an html file
 - This introduces `ResultsPanelSubscriptionProvider` which is responsible for handling commands against the results panel

**Testing**
 - [x] Tested uploading single and multiple files
 - [x] Tested downloading single files, files + folder, files + folder + a subset of children in the folder
 - [x] Tested downloading results (tested w/ tables & sgplots)


package.json
```diff
@@ -592,6 +607,12 @@
       }
     ],
     "menus": {
+      "webview/context": [
+        {
+          "command": "SAS.saveHTML",
+          "when": "webviewSection == 'resultsView'"
```
scnwwu, line 613: Looking at VS Code's doc, `webviewSection` is for different sections in a webview, which we don't need.
I think we should primarily check `webviewId`, which is the first parameter passed to `window.createWebviewPanel`.
It is currently set to "SASSession", perhaps we can revise it to something like "SASResult".

package.json
```diff
@@ -769,6 +800,14 @@
...
           "command": "SAS.downloadTable"
+        },
+        {
+          "when": "false",
+          "command": "SAS.downloadResource"
+        },
+        {
+          "when": "false",
+          "command": "SAS.uploadResource"
         }
```
scnwwu, line 811: do we need an entry for saveHTML here?

client/src/components/ResultPanel/index.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { Disposable, Uri, commands, window } from "vscode";
+
+import { createWriteStream } from "fs";
```
scnwwu, line 5: It would be great to use `vscode.workspace.fs` whenever possible to work across virtual file systems VS Code supports.

client/src/components/ResultPanel/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    return [
+      commands.registerCommand(
+        "SAS.saveHTML",
+        async (context: ResultsContext) => {
+          const html = resultsHtml[context.uuid] || "";
+          if (!html) {
+            return;
+          }
+          const uri = await window.showSaveDialog({
+            defaultUri: Uri.file(`results.html`),
```
scnwwu, line 21: nit: I guess it would be great if we can prepend the SAS code file name to it as SAS Studio does.

package.json
scottdover: Done in https://github.com/sassoftware/vscode-sas-extension/pull/550/commits/242da852b3b7fa434704cf23f8c561752ab37048

client/src/components/ResultPanel/index.ts
scottdover: It seems like that could be more of a challenge since we could have saved files, and files that have yet to be saved. That said, we could, at the very least, make it such that the default uri better matches the webview panel title. I'll look into that.

client/src/components/ResultPanel/index.ts
scottdover: Done https://github.com/sassoftware/vscode-sas-extension/pull/550/commits/7834fea3dd0fa8e2b567e73b159125587d079ba9

package.json
scottdover: Done https://github.com/sassoftware/vscode-sas-extension/pull/550/commits/7834fea3dd0fa8e2b567e73b159125587d079ba9

client/src/components/ResultPanel/ResultPanel.ts
```diff
@@ -2,14 +2,41 @@
...
+import { v4 } from "uuid";
+
 import {
   isSideResultEnabled,
   isSinglePanelEnabled,
-} from "./Helper/SettingHelper";
+} from "../Helper/SettingHelper";
 
 let resultPanel: WebviewPanel | undefined;
+export const resultsHtml: Record<string, string> = {};
```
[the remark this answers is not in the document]
scottdover: Updated this to store panel references instead - https://github.com/sassoftware/vscode-sas-extension/pull/550/commits/3d27e482a7d437ebe922fd595211a00086b20ec5

client/src/components/ResultPanel/ResultPanel.ts
```diff
@@ -42,4 +69,5 @@ export const showResult = (html: string, uri?: Uri, title?: string) => {
     );
   }
   resultPanel.webview.html = html;
+  resultPanels[vscodeContext.uuid] = resultPanel;
```
scnwwu, line 72: Here we're re-using previous resultPanel. So there's already a uuid entry referring the instance.
When dispose, it only deletes the initial uuid entry. So it looks like preserving disposed references. Not sure if it's a memory leak issue.

scottdover: Not sure if I understand what you're saying. Can you elaborate?

scnwwu: When the first time show result, say `vscodeContext.uuid === 001`. We create a resultPanel instance say `instance1`.
So
```
resultPanels === {
    '001': instance1
}
```
And we register
```
instance1.onDidDispose(() => {
    delete resultPanels['001'];
});
```
When the 2nd time show result, say `vscodeContext.uuid === 002`. And user set `singlePanel === true`. The `instance1` is re-used.
So
```
resultPanels === {
    '001': instance1
    '002': instance1
}
```
Then when instance1 disposed, the `resultPanels` still has entries referring to `instance1`.
```
resultPanels === {
    '002': instance1
}
```

scottdover: Ah, got it. Thanks for explaining. I've updated things to use `viewType` (the first parameter in `createWebviewPanel`) to store result panels and have verified things are stored as expected in https://github.com/sassoftware/vscode-sas-extension/pull/550/commits/0e50c8d79562e2a63a66d2d8d8f62042df51cc13

scnwwu (approved):

Zhirong2022:
It failed to download files with Windows OS.
Error details:
Error running command SAS.downloadResource: Unable to resolve filesystem provider with relative file path 'c:\VSCCode\NewFile.sas'. This is likely caused by the extension that contributes SAS.downloadResource.

Zhirong2022:
Local files are not shown on file selection window if trying to Upload one with Windows OS.

Zhirong2022:
'Uploading files...' keeps showing in some scenarios. 
1. If trying to upload the same file folder and its subs on the second time to the same place
2. If trying to upload a file with large size. It is more than 20M.

scottdover:
> It failed to download files with Windows OS.
> Error details:
> Error running command SAS.downloadResource: Unable to resolve filesystem provider with relative file path 'c:\VSCCode\NewFile.sas'. This is likely caused by the extension that contributes SAS.downloadResource.

> Local files are not shown on file selection window if trying to Upload one with Windows OS.

I'm currently waiting on a windows VM to be able to test/debug these.

scottdover:
> > It failed to download files with Windows OS.
> > Error details:
> > Error running command SAS.downloadResource: Unable to resolve filesystem provider with relative file path 'c:\VSCCode\NewFile.sas'. This is likely caused by the extension that contributes SAS.downloadResource.
> 
> > Local files are not shown on file selection window if trying to Upload one with Windows OS.
> 
> I'm currently waiting on a windows VM to be able to test/debug these.

These should both be resolved now (by https://github.com/sassoftware/vscode-sas-extension/pull/550/commits/22dadbdf89d3403b7648574ccc7d11f2a1d5a091 and https://github.com/sassoftware/vscode-sas-extension/pull/550/commits/6c39b076755f05f3bbd3cbe8346c474f8e8fc679). The behavior of uploads has been tweaked a bit. For Mac, a user can use the `Upload` menu item and choose a mix of files and folders to upload.

On Windows & Linux, a user will be presented with two options: `Upload File(s)` and `Upload Folder(s)` (mixing and matching isn't allowed on those OSes). Those options will allow you to choose one type to upload at a time.

scottdover:
@Zhirong2022 This is ready to test again when you have a moment

Zhirong2022:
@scottdover  Thank you so much for your investigation and help.  I verified the issues in Windows OS, it can upload files and folders through Upload file(s) and Upload folder(s) on the context menu.  For the error prompt, it can be enhanced.  I have no other concerns.

Zhirong2022 (approved):

Issue #521, opened 2023-09-18
Identify path of sas program file being submitted - at least when using Local SAS 9.4


jbodart-argenx:
**Is your feature request related to a problem? Please describe.**
My programs need to identify from which path they are executed, and the SAS program file name, in order to retrieve metadata from a parent/sibling folder.  When submitting from SAS 9.4 Enhanced Editor this information can be retrieved from environment variable `SAS_EXECFILEPATH`.  When submitting in batch mode it can be retrieved from `%sysfunc(getoption(sysin))`.  In other circumstances it can be retrieved from `SASHELP.VEXTFL`.  In LSAF, it can be retrieved from global macro-variable `&_SASFILEPATH_`.

**Describe the solution you'd like**
An environment variable, or global macro variable, should be set to the path of the SAS program from which code is being submitted.

**Describe alternatives you've considered**
Alternatively the path should be retrievable from a SASHELP view or by running some macro or data step code or some SAS procedure, or even by executing some Java method.

**Additional context**

**Environment**
PC SAS version 9.4 (windows 10 desktop)

scnwwu:
There's a global macro variable `workDir` storing the working directory. You can run SAS code `%put &=workDir;` to show the path in SAS log.

jbodart-argenx:
Thanks @scnwwu , but this does not help.
Let's say I've opened the program `C:\test1\sas-vscode-test.sas` in VScode editor, with the line:
``` 
%put &workdir;
```
When I run that line of code, the log shows:
```
9          %put &=workdir;
WORKDIR=C:\Users\jbodart\AppData\Local\Temp\SAS Temporary Files\_TD6680_ARGENXLAP3088_\Prc2
```
but what I need to retrieve is the path of my program: `C:\test1\sas-vscode-test.sas`

PR #551, opened 2023-10-05
assign _SASPROGRAMFILE macro-variable to path of submitted SAS notebook code (Local SAS 9.4)

jbodart-argenx:
extends #521

Summary
Add code to set environment variable SAS_EXECFILEPATH to the path of the SAS Notebook code is being submitted from

Testing
manually tested against local SAS 9.4 profile


scnwwu (approved):
looks good to me.

Zhirong2022 (approved):

PR #552, opened 2023-10-05
feat: allow convert to flow for explorer files

scottdover:
**Summary**
This does a few things to improve the sasnb -> dataflow experience:
 - Provides visual indication of converting a notebook to dataflow
 - allows converting sasnb files to notebooks from the explorer view (We default to storing notebooks in "My Folder")
 - updates messaging to indicate _where_ a notebook was created (mainly to clear up confusion about where the dataflow went)
 - updates sas content to display created dataflow files, and updates content model to display `FileOpenError` when trying to open files

**Testing**
 - [x] Attempted to "Convert to flow..." from sas content and local filesystem
 - [x] Attempted to open a flow file in vscode
 - [x] Test renaming dataflow
 - [x] Test renaming non-dataflow file
 - [x] Test deleting dataflow
 - [x] Test favoriting dataflow

### DIFF
--- a/client/src/components/ContentNavigator/ContentModel.ts
+++ b/client/src/components/ContentNavigator/ContentModel.ts
@@ -15,8 +15,8 @@ import {
 } from "../../connection/studio";
 import { SASAuthProvider } from "../AuthProvider";
 import {
+  DATAFLOW_TYPE,
   FAVORITES_FOLDER_TYPE,
-  FILE_TYPE,
   FILE_TYPES,
   FOLDER_TYPE,
   FOLDER_TYPES,
@@ -307,17 +307,14 @@ export class ContentModel {
         );
       }
 
-      const patchResponse = await this.connection.patch(
+      const patchResponse = await this.connection.put(
         uri,
-        { name },
+        { ...res.data, name },
         {
           headers: {
             "If-Unmodified-Since": fileTokenMap.lastModified,
             "If-Match": fileTokenMap.etag,
-            "Content-Type":
-              !isContainer(item) && !itemIsReference
-                ? "application/vnd.sas.file+json"
-                : undefined,
+            "Content-Type": fetchItemContentType(item),
           },
         },
       );
@@ -654,7 +651,7 @@ export class ContentModel {
 
 const getPermission = (item: ContentItem): Permission => {
   const itemType = getTypeName(item);
-  return [FOLDER_TYPE, FILE_TYPE].includes(itemType) // normal folders and files
+  return [FOLDER_TYPE, ...FILE_TYPES].includes(itemType) // normal folders and files
     ? {
         write: !!getLink(item.links, "PUT", "update"),
         delete: !!getLink(item.links, "DELETE", "deleteResource"),
@@ -669,4 +666,17 @@ const getPermission = (item: ContentItem): Permission => {
           itemType !== FAVORITES_FOLDER_TYPE &&
           !!getLink(item.links, "POST", "createChild"),
       };
+};
+
+const fetchItemContentType = (item: ContentItem): string | undefined => {
+  const itemIsReference = item.type === "reference";
+  if (itemIsReference || isContainer(item)) {
+    return undefined;
+  }
+
+  if (item.contentType === DATAFLOW_TYPE) {
+    return "application/json";
+  }
+
+  return "application/vnd.sas.file+json";
 };
--- a/client/src/components/ContentNavigator/ContentModel.ts
+++ b/client/src/components/ContentNavigator/ContentModel.ts
@@ -56,6 +56,10 @@ export class ContentModel {
     this.viyaCadence = "";
   }
 
+  public connected(): boolean {
+    return this.authorized;
+  }
+
   public async connect(baseURL: string): Promise<void> {
     this.connection = axios.create({ baseURL });
     this.connection.interceptors.response.use(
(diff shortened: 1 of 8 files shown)


scottdover:
Hey @sahbic . Would you mind taking a look at this and letting me know if it looks good from your perspective?

sahbic:
> Hey @sahbic . Would you mind taking a look at this and letting me know if it looks good from your perspective?

This is a great addition ! I have tested it quickly and I have two notes:
- the flow convertion handles .sasnb and .ipynb but the option does not show for .ipynb files in local filesystem
- We can see .flw files now in SAS Content explorer but we cannot delete/rename them with right clic

scottdover:
> the flow convertion handles .sasnb and .ipynb but the option does not show for .ipynb files in local filesystem

Ah, we recently removed ipynb support as it was not playing well when the jupyter extension was installed, so that is expected. 

> We can see .flw files now in SAS Content explorer but we cannot delete/rename them with right clic

Ah, thanks for finding that. Will resolve shortly

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -297,17 +307,14 @@ export class ContentModel {
         );
       }
 
-      const patchResponse = await this.connection.patch(
+      const patchResponse = await this.connection.put(
```
scnwwu, line 310: Would you mind provide some reasoning for this change or add some comment to the source code?
It looks like `patch` is more efficient as `res.data` is not changed.

client/src/components/ContentNavigator/index.ts
```diff
@@ -4,21 +4,30 @@ import {
...
   ExtensionContext,
+  ProgressLocation,
   Uri,
   commands,
   l10n,
   window,
   workspace,
 } from "vscode";
 
+import { readFileSync } from "fs";
```
scnwwu, line 15: It would be great if we can use `vscode.workspace.fs` so that virtual file system can also be supported.

package.json
```diff
@@ -614,6 +614,13 @@
           "group": "navigation@1"
         }
       ],
+      "explorer/context": [
+        {
+          "command": "SAS.convertNotebookToFlow",
+          "when": "resourceExtname == .sasnb && SAS.connectionType == rest",
```
scnwwu, line 620: It's not supported in `serverId` case, right?
`resourceExtname == .sasnb && !SAS.connection.direct && SAS.connectionType == rest`

scnwwu:
> > the flow convertion handles .sasnb and .ipynb but the option does not show for .ipynb files in local filesystem
> 
> Ah, we recently removed ipynb support as it was not playing well when the jupyter extension was installed, so that is expected.

The real fix for #537 is only this line
https://github.com/sassoftware/vscode-sas-extension/pull/538/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L879

We should not register `.ipynb` to sas-notebook. It's not needed for the ipynb-to-flow feature. Other code from #538 can be reverted.

smorrisj:
> > > the flow convertion handles .sasnb and .ipynb but the option does not show for .ipynb files in local filesystem
> > 
> > 
> > Ah, we recently removed ipynb support as it was not playing well when the jupyter extension was installed, so that is expected.
> 
> The real fix for #537 is only this line https://github.com/sassoftware/vscode-sas-extension/pull/538/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L879
> 
> We should not register `.ipynb` to sas-notebook. It's not needed for the ipynb-to-flow feature. Other code from #538 can be reverted.

I removed converting ipynb to flow because it seemed like it would confuse the user. @scottdover and I discussed #537 and came to consensus that removal was necessary since the intent was to only support converting .sasnb files to flows. Should we get product management to make a call here?

sahbic:
> > > > the flow convertion handles .sasnb and .ipynb but the option does not show for .ipynb files in local filesystem
> > > 
> > > 
> > > Ah, we recently removed ipynb support as it was not playing well when the jupyter extension was installed, so that is expected.
> > 
> > 
> > The real fix for #537 is only this line https://github.com/sassoftware/vscode-sas-extension/pull/538/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L879
> > We should not register `.ipynb` to sas-notebook. It's not needed for the ipynb-to-flow feature. Other code from #538 can be reverted.
> 
> I removed converting ipynb to flow because it seemed like it would confuse the user. @scottdover and I discussed #537 and came to consensus that removal was necessary since the intent was to only support converting .sasnb files to flows. Should we get product management to make a call here?

My input here is that enabling the feature for .ipynb would help onboarding more python users to the sas viya world. Most existing notebook code is .ipynb not .sasnb. I have added the support later because it was requested by one of my colleagues in customer advisory.

client/src/components/ContentNavigator/ContentModel.ts
scottdover: Yeah, I updated this because `patch` was only available for certain content types. For dataflows, `patch` was not a valid request (but `put` was). So instead of having some logic to determine which to choose (`patch` or `put`), I just went with `put` since all types support that request

package.json
scottdover: You're right. Updated in https://github.com/sassoftware/vscode-sas-extension/pull/552/commits/87f18fa61f9ad249a625be6ecdb403b9b3394532

client/src/components/ContentNavigator/index.ts
scottdover: Done https://github.com/sassoftware/vscode-sas-extension/pull/552/commits/a92335600c9a59552ef5b4defccf362448337f24

scnwwu (approved):
ipynb is being discussed separately.
Looks good to me.

Zhirong2022:

PR #444, opened 2023-08-08
feat: run SAS code via vscode tasks

scnjdl:
**Summary**
run SAS code via vscode tasks

**Testing**
manually

PR #554, opened 2023-10-06
chore: reorganize utils

scottdover:
**Summary**
This merges our utils & helpers into a common folder. See the following comment: https://github.com/sassoftware/vscode-sas-extension/pull/444#discussion_r1310475245

**Testing**
 - [x] Made sure there were no build errors and `npm run test` succeeded as expected


client/src/components/utils/throttle.ts
scnwwu, line 1: Well, this file contains not only `throttle` but also `deferred` which is not related. So `throttle` seems not a good name for the whole file. Perhaps `utils.ts` or `index.ts` is fine?

client/src/components/utils/settingHelper.ts
scnwwu, line 1: How about either `utils/settingUtils` or `helper/settingHelper`? IMO, we don't need 2 terms😄

scottdover: I dropped the `Helper` suffix from these files in https://github.com/sassoftware/vscode-sas-extension/pull/554/commits/2fa00755f4280d4f196a70fbb516baefff492283

scnwwu (approved):

PR #555, opened 2023-10-08
feat: recursive Folding Block

jingjiajie:
**Summary**
Added support for recursive Folding Blocks:

Data in Macro
Proc in Macro
/\*region*/ in Data
/\*region*/ in Proc
/\*region*/ in Macro
Data in /\*region*/
Proc in /\*region*/
Macro in /\*region*/
/\*region*/ in /\*region*/

**Testing**
How did you test this change?
Every block shows a folding icon.
```
*region global;
%macro macro1;
    data table1;
        *region data_inner;
        *endregion data_inner;
    run;

    *region proc;
    proc print data=table1;
        *region proc_inner;
        *endregion proc_inner;
    run;
    *endregion proc;
    
%mend;

*region inner;
    /* nothing */
*endregion inner;

*endregion global;
```


scnwwu:
I noticed 2 issues with below code.

```
*region;
data _null_;
run;
*endregion;
```

1. There's no outline for this code in VS Code's Outline panel.
2. Open context menu from code, select `Run Region`, no response.

jingjiajie:
> I noticed 2 issues with below code.
> 
> ```
> *region;
> data _null_;
> run;
> *endregion;
> ```
> 
> 1. There's no outline for this code in VS Code's Outline panel.
> 2. Open context menu from code, select `Run Region`, no response.

@scnwwu Fixed all.

scnwwu (approved):

Zhirong2022:
Acceptance Criteria:

2.Verify that the DATA regions can be folded correctly
3.Verify that the PROC regions can be folded correctly
4.Verify that Macro regions can be folded as expected
5.Verify that the customized regions can be folded as expected, the matched pair would be   
5.1 /*region [optional description] */ /*endregion [optional description] */
5.2 *region  [optional description]; *endregion  [optional description];
5.3 %*region [optional description]  %*endregion [optional description]; 
6.Verify that the nested regions work as expected
7.Verify that the outline works well
8.Verify that 'Run region' function well
9.Verify 1 to 8 in OS: Windows and Mac
10.Verify 1 to 9 in NLS environment

Zhirong2022:
Type the code below, no outline for the customized region and data block.  
%macro filenaming;
    %put &copyres;
%mend;
%filenaming;
*region;
data _null_;
run;

Zhirong2022:
Type code below and click the first line or the second, open context menu to execute 'Run region'.  It will direct to "proc means" block instead of the whole macro block.
```
%macro reportit(request);
    %if %upcase(&request)=STAT %then
       %do;
          proc means;
             title "Summary of All Numeric Variables";
          run;
       %end;
%mend reportit;

```

Zhirong2022:
Perform 'Run region' action on the code below.  If click the first line and execute 'Run region',  the whole customized block will be ran instead of data block only.
```
*region;
data _null_;
run; 
proc print data=sashelp.cars;
run;
*endregion;
```

Zhirong2022:
Type the code below, the **data** block does not fold properly.
```
*region;
data;
asd;
*endregion;
asd;
run;

```
image

jingjiajie:
> Type the code below, the **data** block does not fold properly.
> 
> ```
> *region;
> data;
> asd;
> *endregion;
> asd;
> run;
> ```
> 
> image

@Zhirong2022 
The folding block doesn't support overlapping. In this case, the custom block and the data block overlaps, so technically either the *region; truncates the data block, or the data block truncates the *region;. But we shouldn't define a behavior for this case because we don't allow this.

Other issues are all fixed~

server/src/sas/LexerEx.ts
```diff
@@ -225,7 +231,7 @@ export class LexerEx {
         i++;
         block = blocks[i];
       } while (block && !this._isBefore(pos, this._startPos(block))); // |[]
-      return --i;
+      return block ? --i : -1;
```
scnwwu, line 234: Hi, may I know the reason for this change? If there's not a `i+1` block, the `i` block is still valid.
There's a problem with current implementation. With code
```
data a;
  a;
run;
data b;
  b;
run;
```
If I put cursor at the start of the line `data b` (line 4 col 1), click `Run Region` from context menu, the `data b` is not run.

Zhirong2022:
The outline is not proper for some code blocks. Please type the code below to take a look.  '%filenaming;' is not a block in such scenario.
```
*region;
%macro filenaming;
    %put &copyres;
%mend;
%filenaming;
/*region proc statement*/
proc print data=sashelp.cars; run;
/*endregion*/
*region;
data _null_;
run;
*endregion;

```

server/src/sas/LexerEx.ts
jingjiajie: @scnwwu  In the recursive senario, if a user clicks the firstline **"*region"**, and run region, it will search the block recursively. First it finds the *region block, then it will check every children blocks [data, proc]. If I let "i block" is still valid when there is not a i-1 block, it will then match the data block instead of the outer ***region** block. This is a fault.
```
*region;
data _null_;
run; 
proc print data=sashelp.cars;
run;
*endregion;
```

jingjiajie:
> The outline is not proper for some code blocks. Please type the code below to take a look. '%filenaming;' is not a block in such scenario.
> 
> ```
> *region;
> %macro filenaming;
>     %put &copyres;
> %mend;
> %filenaming;
> /*region proc statement*/
> proc print data=sashelp.cars; run;
> /*endregion*/
> *region;
> data _null_;
> run;
> *endregion;
> ```

Fixed.

server/src/sas/LexerEx.ts
scnwwu: Hi @jingjiajie,
There's another problem. With below code, put cursor at line 3 col 1, run region will run `*region` which is wrong.
```
*region;
proc print data=sashelp.class;
run;proc print data=sashelp.cars;
*endregion;
```

Zhirong2022:
The outline is not correct for the code below. The most outer block '%*region' is not shown as expected.
```
%*region;
%macro filenaming;
    %put &copyres;
%mend;
%filenaming;    
    /*region proc statement*/
    proc print data=sashelp.cars; run;
    /*endregion*/

    *region;
    data _null_;
    run;
    *endregion;

%*endregion;

```

server/src/sas/LexerEx.ts
jingjiajie: Fixed

jingjiajie:
> The outline is not correct for the code below. The most outer block '%*region' is not shown as expected.
> 
> ```
> %*region;
> %macro filenaming;
>     %put &copyres;
> %mend;
> %filenaming;    
>     /*region proc statement*/
>     proc print data=sashelp.cars; run;
>     /*endregion*/
> 
>     *region;
>     data _null_;
>     run;
>     *endregion;
> 
> %*endregion;
> ```

@Zhirong2022 Fixed.

Zhirong2022:
The data block is not folded as expected with the code below. '%*endregion;' is regarded as a normal comment in the code below. So it should be folded into the data block as a whole. 
```
data cars;
   input MODEL $11. @13 (PREF1-PREF7) (1.);
   CARDS;
ELDORADO    0807990
; 
%*endregion;
title1 'RPVWR00A: REPORT WITH WRAP OPTION';
options ls=78 ps=21;

```

Zhirong2022:
The customized block is not shown proper. 
```
data cars;
   input MODEL $11. @13 (PREF1-PREF7) (1.);
   CARDS;
ELDORADO    0807990
; 
*region;
proc print data=sashelp.cars;run;

```

scnwwu:
> The customized block is not shown proper.
> 
> ```
> data cars;
>    input MODEL $11. @13 (PREF1-PREF7) (1.);
>    CARDS;
> ELDORADO    0807990
> ; 
> *region;
> proc print data=sashelp.cars;run;
> ```

@jingjiajie, I understand that technically it's an overlapping case. However, I think user may easily encounter this case, as `run` is not required to end a data step. Is it possible to make sure the syntax region is correct (take priority over custom region) in such case?

jingjiajie:
> The data block is not folded as expected with the code below. '%*endregion;' is regarded as a normal comment in the code below. So it should be folded into the data block as a whole.
> 
> ```
> data cars;
>    input MODEL $11. @13 (PREF1-PREF7) (1.);
>    CARDS;
> ELDORADO    0807990
> ; 
> %*endregion;
> title1 'RPVWR00A: REPORT WITH WRAP OPTION';
> options ls=78 ps=21;
> ```

Fixed.

jingjiajie:
> > The customized block is not shown proper.
> > ```
> > data cars;
> >    input MODEL $11. @13 (PREF1-PREF7) (1.);
> >    CARDS;
> > ELDORADO    0807990
> > ; 
> > *region;
> > proc print data=sashelp.cars;run;
> > ```
> 
> @jingjiajie, I understand that technically it's an overlapping case. However, I think user may easily encounter this case, as `run` is not required to end a data step. Is it possible to make sure the syntax region is correct (take priority over custom region) in such case?

Fixed. 
@scnwwu In this case, it's not really overlapping. I let the proc block terminate the data block across the custom block. You could pull the latest code to have a try.

Zhirong2022:
It does not work appropriately if trying to run region against the code below.  Click line 4 (%filename) and perform Run Region, the following customized block will be ran at the same time.
```
%macro filenaming;
    %put &copyres;
%mend;
%filenaming;
*region;
data _null_;
run;
*endregion;
```

jingjiajie:
> It does not work appropriately if trying to run region against the code below. Click line 4 (%filename) and perform Run Region, the following customized block will be ran at the same time.
> 
> ```
> %macro filenaming;
>     %put &copyres;
> %mend;
> %filenaming;
> *region;
> data _null_;
> run;
> *endregion;
> ```

@scnwwu @Zhirong2022 There is a question. When a user clicks line 4 "%filenaming;" and perform Run Region, should we only run the line 4, or run the whole program? 
For current implementation, line 4 is compiled as a global region. Global regions are automatically ignored, and because there is no recognized region on the selected line, then the whole program will be run. That's current technical implementation, but seems not very reasonable. I'd reconfirm the requirement.

scnwwu:
> @scnwwu @Zhirong2022 There is a question. When a user clicks line 4 "%filenaming;" and perform Run Region, should we only run the line 4, or run the whole program? For current implementation, line 4 is compiled as a global region. Global regions are automatically ignored, and because there is no recognized region on the selected line, then the whole program will be run. That's current technical implementation, but seems not very reasonable. I'd reconfirm the requirement.

For open code (code that doesn't belong to any region), it doesn't make sense to run region. In SAS Studio, the Run Region menu item is hidden in this case. We probably want to do it but it may be out of the scope of this PR.
If you would like to hide the menu item, it would be great. If you can make the same behavior as current main branch, that is to only run the open code, it would be acceptable. If it's not easy, I think it's also acceptable to run all code, as long as the behavior is consistent for all open code cases, so that it's predictable for user. And we can try hide the menu item in future.

jingjiajie:
> > @scnwwu @Zhirong2022 There is a question. When a user clicks line 4 "%filenaming;" and perform Run Region, should we only run the line 4, or run the whole program? For current implementation, line 4 is compiled as a global region. Global regions are automatically ignored, and because there is no recognized region on the selected line, then the whole program will be run. That's current technical implementation, but seems not very reasonable. I'd reconfirm the requirement.
> 
> For open code (code that doesn't belong to any region), it doesn't make sense to run region. In SAS Studio, the Run Region menu item is hidden in this case. We probably want to do it but it may be out of the scope of this PR. If you would like to hide the menu item, it would be great. If you can make the same behavior as current main branch, that is to only run the open code, it would be acceptable. If it's not easy, I think it's also acceptable to run all code, as long as the behavior is consistent for all open code cases, so that it's predictable for user. And we can try hide the menu item in future.

@scnwwu It feels to me the best solution is to hide the menu item when a user changes cursor position into the open code area. But, I didn't find a method to trigger codes in vscode extension when cursor position changed. Do you have ideas about this? If it can't be done, I can adopt the second solution, to run only the open code.

jingjiajie:
> It does not work appropriately if trying to run region against the code below. Click line 4 (%filename) and perform Run Region, the following customized block will be ran at the same time.
> 
> ```
> %macro filenaming;
>     %put &copyres;
> %mend;
> %filenaming;
> *region;
> data _null_;
> run;
> *endregion;
> ```

Fixed.

scnwwu:
@jingjiajie, thank you for trying this but there's a problem for this approach.
User can directly right click on a line without put cursor on it first. The `setContext` is an async operation so the menu item can't be updated in time.

jingjiajie:
> @jingjiajie, thank you for trying this but there's a problem for this approach. User can directly right click on a line without put cursor on it first. The `setContext` is an async operation so the menu item can't be updated in time.

@scnwwu Yes, I reptoduced this problem. I think we could just run the open code instead of disabling the menu item, if there's no better solution.

scnwwu:
> > @jingjiajie, thank you for trying this but there's a problem for this approach. User can directly right click on a line without put cursor on it first. The `setContext` is an async operation so the menu item can't be updated in time.
> 
> @scnwwu Yes, I reptoduced this problem. I think we could just run the open code instead of disabling the menu item, if there's no better solution.

Agreed

jingjiajie:
> > > @jingjiajie, thank you for trying this but there's a problem for this approach. User can directly right click on a line without put cursor on it first. The `setContext` is an async operation so the menu item can't be updated in time.
> > 
> > 
> > @scnwwu Yes, I reptoduced this problem. I think we could just run the open code instead of disabling the menu item, if there's no better solution.
> 
> Agreed

**done.**

client/src/utils/utils.ts
scnwwu, line 1: well, there's already a `client/src/components/utils/` directory. I would suggest put it there, or don't make this change if not necessary.

client/src/node/extension.ts
scnwwu, line 1: looks unnecessary to change this file now

client/src/commands/run.ts
```diff
@@ -193,7 +135,15 @@ export async function runSelected(uri: Uri): Promise<void> {
 
 export async function runRegion(client: BaseLanguageClient): Promise<void> {
   const selections = await getSelectedRegions(client);
-  window.activeTextEditor.selections = selections;
+  const editor = window.activeTextEditor;
+  const doc = editor.document;
+  if (selections.length === 0) {
+    editor.selections = [
```
scnwwu, line 141: looks like not necessary now?

scnwwu:
There's a problem. With below code
```
%macro;
*region;
data;
a;
*endregion;
%mend;
a;
```
Put cursor at line 6 (start of %mend;), run region, it run whole file. But it should run only the macro region.

jingjiajie:
> There's a problem. With below code
> 
> ```
> %macro;
> *region;
> data;
> a;
> *endregion;
> %mend;
> a;
> ```
> 
> Put cursor at line 6 (start of %mend;), run region, it run whole file. But it should run only the macro region.

Fixed.

client/src/commands/run.ts
scnwwu, line 1: It would be great if you can make the file identical to the base commit (not in diff at all). It will make future blame easier. Thanks.

jingjiajie: done.

client/src/node/extension.ts
```diff
@@ -48,6 +48,7 @@ import { NotebookSerializer } from "../components/notebook/Serializer";
 import { ConnectionType } from "../components/profile";
 import { SasTaskProvider } from "../components/tasks/SasTaskProvider";
 import { SAS_TASK_TYPE } from "../components/tasks/SasTasks";
+import { getSelectedRegions } from "../utils/utils";
```
scnwwu, line 51: @jingjiajie same here, it currently fails the build

jingjiajie: done.

Zhirong2022:
Type code below, click line 5 and execute 'Run region' through context menu.  It should run filename statement only. And now it will run all the code. 
```
*region;
options set=SAS_HADOOP_JAR_PATH="/third_party/Hadoop/jars/lib";
options set=SAS_HADOOP_CONFIG_PATH="/sasdata/cluster1/conf";
*endregion;
filename foo hadoop 'file1.dat' user='user' pass='apass' recfm=s lrecl=1024 bufferlen=1000000;
data _null_;
   infile foo truncover;
input a $1024.;
put a;
run;

Zhirong2022:
All issues have been fixed.

Issue #562, opened 2023-10-10
"Connecting to SAS Session" hang when attempting to connect to an expired deployment


smorrisj:
**Describe the bug**
The "Connecting to SAS Session" toast hangs when attempting to run code on an expired deployment

**Steps to reproduce**
1.) Create a connection profile for viya 4 to a known expired deployment.
2.) Run some SAS code. 
3.) Observe that the toast spins without resolving.

**Expected behavior**
When obtaining a session from the compute server, if the license is expired, the job will be in failed state. We need to make sure to react properly to this scenario so that the correct error can be bubbled up in an error message.

Zhirong2022:
"Connecting to SAS Session" will not hang when attempting to connect to an expired deployment. And now it will show error 'Request failed with status code 400'.

PR #564, opened 2023-10-11
fix: check for failed state during session log stream to prevent unbounded loop

smorrisj:
**Summary**
Check for failed state when streaming the session log to avoid attempting to check for logs in an unbounded loop.

**Testing**
Steps in #562

### DIFF
--- a/client/src/connection/rest/session.ts
+++ b/client/src/connection/rest/session.ts
@@ -204,6 +204,7 @@ export class ComputeSession extends Compute {
       "warning",
       "completed",
       "idle",
+      "failed",
     ];
 
     while (states.indexOf(state) === -1) {
--- a/client/src/commands/run.ts
+++ b/client/src/commands/run.ts
@@ -19,7 +19,12 @@ import {
   wrapCodeWithOutputHtml,
 } from "../components/utils/sasCode";
 import { isOutputHtmlEnabled } from "../components/utils/settings";
-import { OnLogFn, RunResult, getSession } from "../connection";
+import {
+  ErrorRepresentation,
+  OnLogFn,
+  RunResult,
+  getSession,
+} from "../connection";
 import { profileConfig, switchProfile } from "./profile";
 
 interface FoldingBlock {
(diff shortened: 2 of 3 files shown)


client/src/connection/rest/session.ts
```diff
@@ -204,6 +204,7 @@ export class ComputeSession extends Compute {
       "warning",
       "completed",
       "idle",
+      "failed",
```
smorrisj, line 207: failed is one of the possible job states that indicates that the request is finished. Add it here so that the while loop will properly terminate on failed jobs.

smorrisj:
I noticed that after fixing this, we get a rather ugly looking json object rendered in the error message toast. This is due to how we're currently showing error messages.As part of this PR I'd propose that we come up with a way to clean this up a little. There are a few different error models that we could encounter at this level:

- Axios request errors
- Axios response errors
- Axios error in setting up the request (not the same as the first issue above)
- General typescript/node errors

Based on these possible states it seems like we might need some sort of well known translator that knows how to deal with the possibly different types coming in. Thoughts?

I've looked into making the error display in a modal error message, but it looks fairly different than the usual toast messages. The other tricky part specifically in the expired license case is that the context of the error that tells the user what's going on is actually in the details array, whereas the message value on that case is not very clear. 

VSCode's error dialog in the modal option does allow for passing details in a details array. One such approach might be to detect if we are dealing with an error representation, if we are, then see if there are details. On the case of a details array we'd use the modal form. In the case of empty or undefined details array we'd use the normal brief toast message approach. Thoughts?

smorrisj:
Example of what the modal dialog for expired license would look like. Message is the top string and the details array are joined in the detail text below:

image

scnwwu (approved):
Looks good to me

scnwwu:
> VSCode's error dialog in the modal option does allow for passing details in a details array. One such approach might be to detect if we are dealing with an error representation, if we are, then see if there are details. On the case of a details array we'd use the modal form. In the case of empty or undefined details array we'd use the normal brief toast message approach. Thoughts?

Modal dialog looks good to me. I guess the most tricky part is where's the real detail. I remembered some errors where details is also complex objects.

smorrisj:
@scnwwu could I trouble you for another review? I pushed some code that switches over to the modal dialog approach discussed further up.

scnwwu (approved):

Zhirong2022:
"Connecting to SAS Session" will not hang when attempting to connect to an expired deployment. And now it will show error 'Request failed with status code 400'.

Zhirong2022 (approved):

PR #573, opened 2023-10-18
feat: close session on status bar

### LINKED ISSUE #119: add a visual indication whether a connection to a compute server exists or not
Currently there is no visual indication to see whether a connection to the compute server exists or not.
image

One can hover over the icon and then get the "close connection", but you always get this text independant of whether there is a connection or not

scnwwu:
**Summary**
Resolve #119

- Remove the disconnect button from the editor toolbar (next to the run button), as it's not specific to a file.
- Change the icon of the active profile status bar item when a compute session established to provide visual indication.
- Add `Close Session` button on the tooltip of the active profile status bar item when a compute session established.
- Add `SAS Profile` title to the tooltip of the active profile status bar item
- (Engineering) Extract status bar related code to a separate file

**Testing**
Create and close sessions to see the changes.

### DIFF
--- a/client/src/components/StatusBarItem.ts
+++ b/client/src/components/StatusBarItem.ts
@@ -1,0 +1,41 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { MarkdownString, StatusBarAlignment, Uri, l10n, window } from "vscode";
+
+import { profileConfig } from "../commands/profile";
+
+const statusBarItem = window.createStatusBarItem(StatusBarAlignment.Left);
+statusBarItem.command = "SAS.switchProfile";
+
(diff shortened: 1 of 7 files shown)


scnwwu:
@clangsmith mentioned an idea to also have a `Switch Profile` button on the tooltip of the active profile status bar item. So that if user doesn't know that clicking the status bar item can switch profile, they may find it on the tooltip.
I played a bit with this idea but found I'm not a fan of it.
I worry if user first see the tooltip may think it's the only way or our expected way to switch profile. But it's not a good experience to have to hover on the item and wait for the tooltip to popup and move mouse carefully on to a text link, compared to simply click the item.
Every status bar item is clickable in VS Code. It's common. I think we should guide user to take this better approach.

scottdover (approved):
This looks good to me. Sorry this one fell of my radar for so long

Zhirong2022 (approved):

Zhirong2022:

PR #575, opened 2023-10-19
chore: store built vsix

scnwwu:
**Summary**
Store the built vsix file as artifact so that anyone can download and directly play with PR or latest main branch without build effort.

**Testing**
See artifact for this build https://github.com/sassoftware/vscode-sas-extension/actions/runs/6571171460

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -1,11 +1,13 @@
-# This workflow perform basic checks for a pull request
+name: Package VSIX
 
-name: Pull Request Check
-
-on: pull_request
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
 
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -27,3 +26,18 @@ jobs:
             echo $CHANGES
             exit 1
           fi
+
+  build:
```
scottdover, line 30: Translation check was setup for `workflow_dispatch` where you can manually trigger a run of it in github actions as needed. Do we think adding this here will cause any issues with that?

scnwwu: Thanks for pointing it out. I separated them.

scottdover (approved):

Issue #583, opened 2023-10-25
Coding issue of the log of local profile


ShuguangSun:
**Describe the bug**
When I use the local profile (COM), if the code contains non-ascii code (e.g. Chinese or other UTF8 symbols), it would 
1. can't recognized the char correctly in the code, and SAS would raise error
2. If put those chars in the log, VScode would report error like "FlushLog Error"

**Steps to reproduce**

Code example 1: `ERROR: The name '�'n is not a valid SAS name.`
```sas
data _null_;
    put "啊";
run;
```

Code example 2: `Flush log Error`
```sas
data trt;
  input trt $20.;
  cards;
紫杉醇
多柔比星脂质体
拓扑替康
;
run;

data _NULL_;
  set trt;
  put (trt)(=);
run;
```

Code example 3: 
```sas
data trt;
  input trt $20.;
  cards;
紫杉醇
多柔比星脂质体
拓扑替康
;
run;

proc print data = trt noobs;
run;
```

It seems the data is not correct:
```
<html>
<body>
<!--StartFragment-->

trt
--
绱潐閱?
澶氭煍姣旀槦
鎷撴墤鏇垮悍

<!--EndFragment-->
</body>
</html>
```

**Expected behavior**
Put the char correctly in the log window.

**Environment (please complete the following information):**
Client OS: Windows 2019
Extension version: 1.4.1
SAS version: 9.4

**Additional context**

The COM script relies on powershell. It seems the coding the code and log is not well handled. Maybe we can use utf-8 anyway.

Zhirong2022:
@ShuguangSun , I cannot reproduce the issue in my testing environment. Perhaps, you did not configure the language for local 94. Please find the .cfg file at  C:\Program Files\SASHome\SASFoundation\9.4 to make sure zh is set. 
-config "C:\Program Files\SASHome\SASFoundation\9.4\nls\zh\sasv9.cfg"
![image](https://github.com/sassoftware/vscode-sas-extension/assets/121005702/9989ff03-bfdd-4617-b190-06b973c3de1a)

ShuguangSun:
Thanks. My SAS (local/COM) server's encoding is UTF-8 and I can't change it. So I tried to change the encoding of remote poweshell to utf-8 by put `chcp 65001` in '$profile'.  Now example 1 and 3 are well solved.

The exmple 2 still get 'Flush log Error' message. I'll have other tries.

ShuguangSun:
[...]

PR #524, opened 2023-09-19
Assign _SASPROGRAMFILE macro-variable to full path and filename of submitted SAS program (Local SAS 9.4)

jbodart-argenx:
implements #521 

**Summary**
Add code to set environment variable SAS_EXECFILEPATH to the full path and filename of the SAS program code is being submitted from

**Testing**
manually tested against local SAS 9.4 profile

PR #592, opened 2023-10-31
feat: add sas 9 remote (via IOM bridge) support

scottdover:
**Summary**
This adds sas 9 IOM bridge support. There are a few things to note here:
 - This uses ITC to connect to SAS 9 via the IOM bridge protocol. Our current local sas 9 connection uses the same underlying connection. Therefore, instead of recreating the wheel, our current com session was modified.
 - All of the connection files were moved from `com -> itc` to better match connection info
 - fetching file results no longer assumes the file is on your local machine. Instead, it uses the file service to grab the results from the last run

**Testing**
 - [x] Test running multiple sas programs via `SAS 9.4 (remote - IOM)`
 - [x] Test running multiple sas programs via `SAS 9.4 (local)`
 - [x] Test running with and without ods results enabled
 - [x] Tested printing ODS results of various sizes (small tables, tables w/ 1k+ rows)


client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,385 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
```
scottdover, line 1: Fwiw, this was moved from `itc -> com`. Most of the functionality is _not_ new

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). If you introduce breaking changes, please group them together in the "Changed" section using the **BREAKING:** prefix.
 
+## [Unreleased]
+
+## Added
+
+- Added sas 9.4 remote connection support via ITC and the IOM Bridge protocol. This is currently supported for Windows only ([#592](https://github.com/sassoftware/vscode-sas-extension/pull/592))
```
scnwwu, line 11: nit: Is it clear enough to just say Windows only here? Will any users treat it as 9.4 remote on Windows?
It's actually "VS Code Extensions Host on Windows only". So technically a Mac user can tunnel to a Windows VS Code workspace and run the extension there. I guess we don't want to say too much here in changelog, perhaps say something in connect-and-run.md?

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,385 @@
...
+      runResult.html5 = htmlResults;
+      runResult.title = "Result";
+    }
+    this.storePassword();
+    this._runResolve(runResult);
+  };
+}
+
+/**
+ * Creates a new SAS 9 Local Session.
```
scnwwu, line 364: nit: perhaps update the comment to say not only local

connect-and-run.md
```diff
@@ -128,36 +146,41 @@ Note: the default path to the SAS executable (saspath) is /opt/sasinside/SASHome
...
 | **Host** | Indicates SAS 9.4 local instance | Defaults to localhost for com       |
 
-## Add New SAS 9.4 Local Profile
+### Add New SAS 9.4 (local) Profile
+
+Open the command palette (`F1`, or `Ctrl+Shift+P` on Windows). After executing the `SAS.addProfile` command, select the SAS 9.4 (local) connection type to create a new profile.
+
+## Profile: SAS 9.4 (remote - IOM)
+
+On Windows, during the install of SAS 9.4, make sure that "Integration Technologies Client" checkbox is checked. If using an order that doesn't provide this option, make sure the ITC is installed by visiting the following [link](https://support.sas.com/downloads/browse.htm?fil=&cat=56). Make sure to download the 9.4m8 option.
```
scnwwu, line 168: nit: we probably would want to differentiate some wording here from the local com. As the user just want to connect to remote, local install of 9.4 is not a primary task. Installing ITC only or 9.4 with ITC are same level of alternatives.

package.json
```diff
@@ -323,6 +324,41 @@
...
+                            "description": "%configuration.SAS.connectionProfiles.profiles.ssh.host%"
+                          },
+                          "username": {
+                            "type": "string",
+                            "default": "",
+                            "description": "%configuration.SAS.connectionProfiles.profiles.ssh.username%"
+                          },
+                          "port": {
+                            "type": "number",
+                            "default": 22,
```
scnwwu, line 354: 8591?

client/src/node/extension.ts
```diff
@@ -55,8 +55,11 @@ const activeProfileStatusBarIcon = window.createStatusBarItem(
   0,
 );
 
+export let extensionContext: ExtensionContext | undefined;
```
scnwwu, line 58: There's a `client/src/components/ExtensionContext` trying to encapsulate it. How about providing get/setSecret there?

CHANGELOG.md
scottdover: I don't think I follow. Currently, this profile type is only supported in a Window environment. If a user is somehow tunneling through (not exactly what that means), they would then be on a Windows environment, right (technically speaking)?

package.json
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/1d7ee717f30707e321a05af5152a4c8ea2b388b9

client/src/connection/itc/index.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/1d7ee717f30707e321a05af5152a4c8ea2b388b9

client/src/node/extension.ts
scottdover: Ha. Didn't know about this one. Updated in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/551093751ab18c3476a50ad45388a9da33c54df9

connect-and-run.md
scottdover: This has been updated in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/a016339016c3d1233b63ad43417c350f6487893c

CHANGELOG.md
scnwwu: I'm thinking if I'm a user who have no knowledge of ITC and IOM, when I see just "Windows only", there're 3 possbilities to me.

1. Does it mean my SAS 9.4 needs to be run on Windows only?
2. Does it mean I have to use VS Code on Windows only?
3. Does it mean both my VS Code and my SAS 9.4 need to be on Windows only?

The anwser of the 3 questions are all "no". I'm thinking it would be great to explain more details

connect-and-run.md
```diff
@@ -128,36 +146,41 @@ Note: the default path to the SAS executable (saspath) is /opt/sasinside/SASHome
...
 | **Host** | Indicates SAS 9.4 local instance | Defaults to localhost for com       |
 
-## Add New SAS 9.4 Local Profile
+### Add New SAS 9.4 (local) Profile
+
+Open the command palette (`F1`, or `Ctrl+Shift+P` on Windows). After executing the `SAS.addProfile` command, select the SAS 9.4 (local) connection type to create a new profile.
+
+## Profile: SAS 9.4 (remote - IOM)
+
+In order to use this option, you'll need to have "Integration Technologies Client" (ITC) installed. You can do this by making sure the "Integration Technologies Client" checkbox is checked when installing SAS 9.4, or by visiting the following [link](https://support.sas.com/downloads/browse.htm?fil=&cat=56) (Make sure to download the 9.4m8 option).
```
scnwwu, line 168: nit: as my comment for "Windows only", it would be good to make it clear that the ITC has to be installed on the machine that our extension runs on. Not on the SAS 9.4 server, nor the VS Code UI if the user is tunneling.

CHANGELOG.md
scnwwu: By tunneling I'm in general referring to all [remote approaches](https://code.visualstudio.com/docs/remote/remote-overview) provided by VS Code. For Windows, only tunnel and SSH is practical.

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,387 @@
...
+  };
+
+  private fetchPassword = async (): Promise<string> => {
+    if (this._config.protocol === ITCProtocol.COM) {
+      return "";
+    }
+
+    const storedPassword = await getSecret(PASSWORD_KEY);
+    if (storedPassword) {
+      return storedPassword;
```
scnwwu, line 144: You're not setting `this._password` here, so `this._password` can still be `""` right?
When `storePassword()` called, `""` will be stored which is not expected, right?

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,387 @@
...
+    // free objects in the scripting env
+    process.on("exit", async () => {
+      close();
+    });
+
+    return setupPromise;
+  };
+
+  private storePassword = async () =>
+    await setSecret(PASSWORD_KEY, this._password);
```
scnwwu, line 130: Does it need to store by different profile?

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  private clearPassword = async () => {
```
scnwwu, line 132: For security, user may need a way to manually log out.

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,387 @@
...
+    if (file) {
+      workspace.fs.delete(outputFileUri);
+    }
+
+    const runResult: RunResult = {};
+    if (htmlResults.search('<*id="IDX*.+">') !== -1) {
+      runResult.html5 = htmlResults;
+      runResult.title = "Result";
+    }
+    this.storePassword();
```
scnwwu, line 361: how about store password on session creation success? User may close VS Code before run code for whatever reason.

CHANGELOG.md
smorrisj: This conversation leads me to further believe that we need to have a wiki entry with some more intricate details around mapping a users sas "environment" to the underlying technologies that can be used in the connection profiles to carry out the session creation. Not sure if we'd want to do that as part of this effort. @jefurbee  any thoughts here?

client/src/connection/itc/index.ts
smorrisj: Eventually we'll want the ability to store username and password for ssh keys. Does that factor into the naming conversation here?

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,387 @@
...
+      `$filePath = "${resolve(
+        this._workDirectory,
+        this._html5FileName + ".htm",
+      )}"
+$outputFile = "${outputFileUri.fsPath}"
+$runner.FetchResultsFile($filePath, $outputFile)\n`,
+      this.onWriteComplete,
+    );
+
+    const file = await new Promise((resolve) => {
```
smorrisj, line 334: Just wondering why were checking to see if maxTime has passed here?  Are we waiting that long for a failed read before giving up? Just wondering if we'd run up against that for potentially large files that might take longer than 10 seconds to read?

scottdover: Yeah, that's a good point, @smorrisj . The reason we have maxTime here is because `FetchResultsFile` takes _some_ time to store the results. 

I'm wondering what a better approach would look like (since we don't _really_ know when FetchResultsFile completes).

smorrisj: Hmm. Could we invent a way to know when it's done? One such thing might be to have the powershell script echo out a unique line that we look for to know that we fetched results, then react to it if we see that line echoed to the onShellStdOut?

client/src/connection/itc/index.ts
scottdover: Right now, the two are linked. We don't create the session _until_ the user clicks the button to run the code. And, since run code has the `endCode` to check for, I just relied on that to store the password. I suppose I could do something similar for session creation (since I'm likely going to do the same for fetching results)

client/src/connection/itc/index.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/4637097276508551fd84e58a5bc7078ea5423581

client/src/connection/itc/index.ts
scottdover: This has been resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/a9acbc563a97095fbfb19218327fe185f00877b6. I basically created a new concept of a "namespaced" secret storage and am using that now for viya & itc sessions. And since all the things are namespaced, this shouldn't create issues w/ ssh remote support

client/src/connection/itc/index.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/1da05d1a445dde2144bc471b9cc842da3dccebc4

client/src/connection/itc/index.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/1da05d1a445dde2144bc471b9cc842da3dccebc4

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,414 @@
...
+    });
+
+    //write ODS output to work so that the session cleans up after itself
+    const codeWithODSPath = code.replace(
+      "ods html5;",
+      `ods html5 path="${this._workDirectory}";`,
+    );
+
+    //write an end mnemonic so that the handler knows when execution has finished
+    const codeWithEnd = `${codeWithODSPath}\n%put ${LineCodes.RunEndCode};`;
```
scnwwu, line 181: Just saw [this](https://github.com/sassoftware/vscode-sas-extension/discussions/605) today. Perhaps we can move this end code also to powersell?

scottdover: Eventually, I think that's a great idea. This is only _now_ complicated by the fact that we are using async submissions  so that cancel can be enabled. 

My preference is to move this to its own pull request after this code has been merged.

client/src/connection/itc/index.ts
scottdover: For now, clicking the disconnect icon will trigger session.close, and we'll clear the password at that point. There's some discussion about using AuthProvider as a solution for _all the things_. I'll explore that in a separate PR

CHANGELOG.md
scottdover: For now, I've removed the "Windows only" part of this message. I think there's probably a bigger conversation that needs to happen like @smorrisj said.

client/src/connection/itc/script.ts
```diff
@@ -75,13 +77,19 @@ class SASRunner{
...
+        $this.objSAS.LanguageService.Reset()
+        $this.objSAS.LanguageService.Async = $true
         $this.objSAS.LanguageService.Submit($code)
     }catch{
       Write-Error $_.ScriptStackTrace
       throw "Run error"
     }
   }
 
+  [void]SubmitComplete() {
```
scnwwu, line 89: nit: looks like not used?

scnwwu (approved):
Looks good to me. Thanks.

client/src/connection/itc/script.ts
scottdover: 😬 You're right. Updated in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/a69deff13525927c73f45d3447b8ecd95ac2600c

Zhirong2022:
Create a new profile with SAS 9.4 (Remote-IOM). The property description is not consistent with the connection type.  Put mouse over 'host', 'port' and 'username'. 
image

Zhirong2022:
It failed to run the code with error below.
Error detail:
There was an error executing the SAS Program. See console log for more details.

scottdover:
> Create a new profile with SAS 9.4 (Remote-IOM). The property description is not consistent with the connection type. Put mouse over 'host', 'port' and 'username'. image

I tested the connection details you sent me and was unable to test the connection using _SAS Integration Technologies Configuration Wizard_. That leads me to believe this is an issue with the server configuration.

scottdover:
> It failed to run the code with error below. Error detail: There was an error executing the SAS Program. See console log for more details.

I tested the connection details you sent me and was unable to make a successful connection using _SAS Integration Technologies Configuration Wizard_. That leads me to believe this is an issue with the server configuration.

scottdover:
> Create a new profile with SAS 9.4 (Remote-IOM). The property description is not consistent with the connection type. Put mouse over 'host', 'port' and 'username'. image

Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/3947797b37504d3cf02238896f99149d63f76bbb

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,413 @@
...
+        this._html5FileName =
+          line.match(/NOTE: .+ HTML5.* Body.+: (.+)\.htm/)?.[1] ??
+          this._html5FileName;
+        this._onLogFn?.([{ type: "normal", line }]);
+      }
+    });
+  };
+
+  private processLineCodes(line: string): boolean {
+    if (line.includes(LineCodes.RunEndCode)) {
```
scnwwu, line 308: It looks like this condition will enter twice for each run. The log will contain the source
`%put --vscode-sas-extension-submit-end--;`
and then the output log
`--vscode-sas-extension-submit-end--`
So it will fetch result twice.

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,413 @@
...
+    } catch (e) {
+      await workspace.fs.createDirectory(globalStorageUri);
+    }
+
+    const outputFileUri = Uri.joinPath(
+      globalStorageUri,
+      `${this._html5FileName}.htm`,
+    );
+    this._shellProcess.stdin.write(
+      `$filePath = "${resolve(
```
scnwwu, line 360: The `resolve` will try to normalize the path based on the OS it is running on, which is Windows.
But the working directory is on the server that SAS is running on, which could be Linux.
The resulting directory may be unexpected. I believe that is the cause of the error Zhirong encountered.

client/src/connection/itc/index.ts
scottdover: Oh you're right. I changed this _back_ to endsWith in _some_ commit, but I'm guessing bouncing between Mac and Windows meant I loss this along the way. Updated in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/8522fbdadb666dbbbf1e191300776895ec15cc7e

client/src/connection/itc/index.ts
scottdover: You're right. I've addressed this in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/61aab6cf6d682732700612cca59b520afae81d0e . Admittedly, my approach seems a _little_ clunky, but should support both COM/IOM Bridge connection types. 

@Zhirong2022 One thing to mention here is that we'll likely want to test COM connections as part of this

Zhirong2022:
Some unreadable Chinese characters are shown on the result page with Local94 connections (COM). 
Please update the sasv9.cfg as below at C:\Program Files\SASHome\SASFoundation\9.4 to start a Chinese SAS Session to reproduce.
-config "C:\Program Files\SASHome\SASFoundation\9.4\nls\zh\sasv9.cfg"
image

[sampleCode.txt](https://github.com/sassoftware/vscode-sas-extension/files/13447927/sampleCode.txt)

Zhirong2022:
Suppose having a .sas file and the file name contains Chinese characters(Test测试数据.sas).  It will fail to run on a specific remote94 server.
Error:
There was an error executing the SAS Program. See console log for more details.

scottdover:
Hey @Zhirong2022 . Do you mind testing the latest two test cases on sas 9 using COM (on the main branch)? I've been looking into the chinese character in the filename issue you've mentioned, and it _looks_ like that issue was introduced by https://github.com/sassoftware/vscode-sas-extension/pull/524

Zhirong2022:
@scottdover , I tested the two cases on Main Branch.  It works well.
image

ShuguangSun:
VScode help menu => Toggle Developer Tools to find the console error message.

It might be related to https://github.com/sassoftware/vscode-sas-extension/issues/583. The coding of powershell and SAS should be the same. In my case, SAS session is n utf-8 encoding, however powershell is in zh_CH (936?) code which make issue 583. So I put `chcp 65001` to powershell's $PROFILE and solve the coding issue.

scottdover:
> VScode help menu => Toggle Developer Tools to find the console error message.
> 
> It might be related to #583. The coding of powershell and SAS should be the same. In my case, SAS session is n utf-8 encoding, however powershell is in zh_CH (936?) code which make issue 583. So I put `chcp 65001` to powershell's $PROFILE and solve the coding issue.

How did you accomplish this? I updated the code to spawn a child process using `chcp 65001 >NUL & powershell.exe ...`, but that didn't seem to work for me

ShuguangSun:
```powershell
echo "chcp 65001" >> $PROFILE
```

For current user and current host, Windows - $PROFILE is  `$HOME\Documents\PowerShell\Microsoft.PowerShell_profile.ps1`
or `%USERPROFILE%\Documents\PowerShell\Microsoft.PowerShell_profile.ps1`

Zhirong2022:
Unreadable characters in log output. Please run the code below to have a try.
```
proc print data=sashelp.cars (obs=1); run;

```
image

scottdover:
Hey @smorrisj @clangsmith @scnwwu . 

I know we continue to have some issues around encoding that @smorrisj is helping me look into, and @snlwih has mentioned expanding IOM support _while_ we get through these encoding issues. Given the majority of the code works as expected, I'm wondering if we can/should...
 - Create a new branch based on main, called `sas9-iom-integration` (or something similar)
 - Merge these changes into `sas9-iom-integration`

That would allow us to more easily review a more focused PR for things like:
 - handling the encoding issues
 - expanding iom support for libraries/etc

That said, it does have the added overhead of first merging into `sas9-iom-integration`, _then_ merging those changes into main, but it's pretty straightforward

smorrisj:
> Unreadable characters in log output. Please run the code below to have a try.
> 
> ```
> proc print data=sashelp.cars (obs=1); run;
> ```
> 
> image

Could you retest with changes from https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/138ad73095605f3ad0368edca158ff7bf63b615c?

scottdover:
This looks good @smorrisj....just looks like tests need to be patched

Zhirong2022:
@scottdover @smorrisj  Thank you for your update.  I retested the raised issues against the new package. 

New findings below:

Issue1: Specify a remote-IOM connection profile and try to run SAS Code, it keeps showing 'SAS code running' initially
Issue2: Cancel sas code running, Ran man icon will be disabled
Issue3: Unreadable Chinese characters on result page. 
Issue4: Errors or something unexpectedly happened if .sas file name contains Chinese characters.
Issue5: Errors or log issue with code having NLS characters.
Issue6: Unreadable Chinese characters in output log 

Issue3 to issue6 , please check the attached doc file.  I listed different scenarios which had errors or something improperly.

ShuguangSun:
See issue https://github.com/sassoftware/vscode-sas-extension/issues/583 . It is quite important to match the encoding of powershell and SAS session (UTF-8 SAS can't read zh_CN Powershell string, vice versa), and then the SAS local with powershell display language. It is impossible to change the encoding of SAS during the session but only at the startup (e.g. sasv9.cfg). So that it seems not workable to force using UTF-8 in powershell. It might be better to leave it to user decided. See issue https://github.com/sassoftware/vscode-sas-extension/issues/594

ShuguangSun:
> Issue3: Unreadable Chinese characters on result page. 
If possible,  check the original html file in the SAS temp to see whether it is good or not.

Zhirong2022:
It looks that the unreadable Chinese characters appear at a fixed position.  It is encoded as "EF BF BD". Something associated with boundary.

client/src/commands/run.ts
```diff
@@ -156,6 +156,8 @@ async function runCode(selected?: boolean, uri?: Uri) {
     (_progress, cancellationToken) => {
       cancellationToken.onCancellationRequested(() => {
         session.cancel?.();
+        running = false;
+        commands.executeCommand("setContext", "SAS.running", false);
```
scnwwu, line 160: It's done at the `finally` call at line 182 below. Is that the `runCode` promise never fulfilled if cancelled? I think it would be great if it can still be fulfilled even cancelled. That will make upper layer easier.

scottdover: Yeah, it's never fulfilled in that case. I half expected I could do something like `const token = cancellationToken.onCancellationRequested...` and call `token.dispose()` to attempt to resolve things, but that also didn't resolve things

scnwwu: The cancellation token is from VS Code UI, the promise is from our own session, they're different.
Perhaps echo something to powsershell when cancel and capture it similar to -submit-end--

scottdover: Good idea. Done in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/bf8ac6a0a1babd95b1c6683b8fbd7ec17c0b9e06

clangsmith:
> Create a new profile with SAS 9.4 (Remote-IOM). The property description is not consistent with the connection type. Put mouse over 'host', 'port' and 'username'. image

I noticed this (incorrectly references "SSH") in the new connection prompts as well (even though I selected IOM in this case, not SSH)...
image

I assume this still needs to be resolved?

clangsmith:
I can successfully run code with MBCS/DBCS characters as long as the SAS server session encoding is set to one that can handle those characters (which one would expect, and how other clients have always behaved). The ODS results also properly render the characters. However, I have not been able to get the DBCS/MBCS characters to render in the log in VS Code in any scenario, even when the VS Code display language matches the SAS session locale and encoding and regardless of whether I explicitly set the Powershell encoding to Simplified Chinese (936) or utf8 (65001) (in the Powershell $PROFILE). So, there still appears to be a problem with that scenario (rendering DBCS/MBCS characters in the log).

scottdover:
> > Create a new profile with SAS 9.4 (Remote-IOM). The property description is not consistent with the connection type. Put mouse over 'host', 'port' and 'username'. image
> 
> I noticed this (incorrectly references "SSH") in the new connection prompts as well (even though I selected IOM in this case, not SSH)... image
> 
> I assume this still needs to be resolved?

Good catch. That's been resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/1ffda86cece6cf5b78f7bceb69740be46e578b70 . I basically updated the text to say "SAS 9 server" instead of "SAS 9 SSH server." Let me know if we'd prefer to keep the strings _more_ specific (i.e. "SAS 9 IOM server" and "SAS 9 SSH server")

clangsmith:
> I can successfully run code with MBCS/DBCS characters as long as the SAS server session encoding is set to one that can handle those characters (which one would expect, and how other clients have always behaved). The ODS results also properly render the characters. However, I have not been able to get the DBCS/MBCS characters to render in the log in VS Code in any scenario, even when the VS Code display language matches the SAS session locale and encoding and regardless of whether I explicitly set the Powershell encoding to Simplified Chinese (936) or utf8 (65001) (in the Powershell $PROFILE). So, there still appears to be a problem with that scenario (rendering DBCS/MBCS characters in the log).

This has been resolved for me with @scottdover's latest changes, thanks! The DBCS characters are now rendering correctly in the log as well. Thus, DBCS characters are now rendering correctly everywhere I'd expect when an appropriate SAS server session encoding is set -- in the log and ODS results.

clangsmith:
I did find an issue though...
The "Close Session" menu is no longer available after making changes to the active profile, even though I clearly have an open session.

Steps to reproduce:

1. Define an IOM profile (ex. uitsrv05, port 8591)
2. Submit any code (ex. "%put hello;")
3. Hover over the profile on the status bar at the bottom of VS Code and confirm the "Close Session" menu is present, as expected (but do not click it).
4. Open the settings.json file and change the port to another server context on that server (ex. port 8596, still on uitsrv05).
5. Save the settings.json file.
6. Hover over the profile again on the status bar at the bottom of VS Code and confirm the "Close Session" menu is NO LONGER present.
7. Submit the same program again.
8. Hover over the profile again on the status bar at the bottom of VS Code.

Problem: Notice the "Close Session" menu is still not there. I expected it to re-appear after running with the new port, since I thought that would have resulted in a new session.

If you explicitly close the session (in step #3), then the problem does not occur.

You can recover the "Close Session" menu by explicitly running the command from command palette to close the SAS session, then make a new session.

I just realized, it looks like I'm not getting a new session when I only change the port (and confirmed same even if I change the host), since I was not re-prompted for password. So, it looks like changing any active profile properties doesn't close the active session, which is ok, but the "Close Session" menu should not have gone away.

scnwwu:
> I did find an issue though... The "Close Session" menu is no longer available after _any_ changes to the settings.json file are saved, even though I clearly still have an open session.
> 
> Steps to reproduce:
> 
> 1. Define an IOM profile (ex. uitsrv05, port 8591)
> 2. Submit any code (ex. "%put hello;")
> 3. Hover over the profile on the status bar at the bottom of VS Code and confirm the "Close Session" menu is present, as expected (but do not click it).
> 4. Open the settings.json file and change the port to another server context on that server (ex. port 8596, still on uitsrv05).
> 5. Save the settings.json file.
> 6. Hover over the profile again on the status bar at the bottom of VS Code and confirm the "Close Session" menu is NO LONGER present.
> 7. Submit the same program again.
> 8. Hover over the profile again on the status bar at the bottom of VS Code.
> 
> Problem: Notice the "Close Session" menu is still not there. I expected it to re-appear after running with the new port, since I thought that would have resulted in a new session.
> 
> If you explicitly close the session (in step #3), then the problem does not occur.
> 
> You can recover the "Close Session" menu by explicitly running the command from command palette to close the SAS session, then make a new session.
> 
> I just realized, it looks like I'm not getting a new session when I only change the port (and confirmed same even if I change the host), since I was not re-prompted for password. So, it looks like changing any active profile properties doesn't close the active session, which is ok, but the "Close Session" menu should not have gone away.

It's not specific to this PR and can be reproduced in main branch with Viya profile too. It can be addressed separately.

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,438 @@
...
+        const parts = line.split("WORKDIR=");
+        this._workDirectory = parts[1].trim();
+        this._runResolve();
+        updateStatusBarItem(true);
+        return;
+      }
+      if (!this.processLineCodes(line)) {
+        this._html5FileName =
+          line.match(/NOTE: .+ HTML5.* Body.+: (.+)\.htm/)?.[1] ??
+          this._html5FileName;
```
scnwwu, line 317: I think it should reset `this._html5FileName` upon each run, as ssh does
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/connection/ssh/index.ts#L89
Otherwise it will use the name from last run if no new ods result produced

client/src/connection/itc/index.ts
```diff
@@ -0,0 +1,438 @@
...
+    if (file) {
+      workspace.fs.delete(outputFileUri);
+    }
+
+    const runResult: RunResult = {};
+    if (htmlResults.search('<*id="IDX*.+">') !== -1) {
+      runResult.html5 = htmlResults;
+      runResult.title = "Result";
+    }
+    this._runResolve(runResult);
```
scnwwu, line 413: Currently, if I uncheck `SAS.results.html.enabled` settings, simply run `%put foo;`. The running will never complete.
We need to also resolve the promise if no `this._html5FileName` captured, as ssh does
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/connection/ssh/index.ts#L134

client/src/connection/itc/index.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/3b80433ddddc02f46f6b00c2e633547af8fdf477

client/src/connection/itc/index.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/3b80433ddddc02f46f6b00c2e633547af8fdf477

clangsmith:
> > I did find an issue though... The "Close Session" menu is no longer available after _any_ changes to the settings.json file are saved, even though I clearly still have an open session.
> > Steps to reproduce:
> > 
> > 1. Define an IOM profile (ex. uitsrv05, port 8591)
> > 2. Submit any code (ex. "%put hello;")
> > 3. Hover over the profile on the status bar at the bottom of VS Code and confirm the "Close Session" menu is present, as expected (but do not click it).
> > 4. Open the settings.json file and change the port to another server context on that server (ex. port 8596, still on uitsrv05).
> > 5. Save the settings.json file.
> > 6. Hover over the profile again on the status bar at the bottom of VS Code and confirm the "Close Session" menu is NO LONGER present.
> > 7. Submit the same program again.
> > 8. Hover over the profile again on the status bar at the bottom of VS Code.
> > 
> > Problem: Notice the "Close Session" menu is still not there. I expected it to re-appear after running with the new port, since I thought that would have resulted in a new session.
> > If you explicitly close the session (in step #3), then the problem does not occur.
> > You can recover the "Close Session" menu by explicitly running the command from command palette to close the SAS session, then make a new session.
> > I just realized, it looks like I'm not getting a new session when I only change the port (and confirmed same even if I change the host), since I was not re-prompted for password. So, it looks like changing any active profile properties doesn't close the active session, which is ok, but the "Close Session" menu should not have gone away.
> 
> It's not specific to this PR and can be reproduced in main branch with Viya profile too. It can be addressed separately.

Thank @scnwwu. I agree. I'll open a separate bug.

doc/scripts/itc-connection-test.ps1
```diff
@@ -0,0 +1,177 @@
+<#
```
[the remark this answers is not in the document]
scottdover: I'm going to say no. Correct me if I'm wrong @scnwwu. The main reason is that nothing in our `tools/` folder has the copyright header, nor do any of our `test.ts` files. 

I'm not entirely sure the distinction of when to include or skip the copyright, though.

clangsmith:
I tested against a server configured with encrypted communications (specifically, Triple DES algorithm). I submitted very simple code (ex. "proc print data=sashelp.class; run;"). The code appears to run correctly, but the Results tab does not appear and the ODS results are not displayed no matter what I submit. @scottdover, I'll send you the connection information for that server, so you can test it out.

client/src/components/utils/sasCode.ts
```diff
@@ -48,14 +49,23 @@ export function wrapCodeWithOutputHtml(code: string): string {
...
     return code;
   }
 }
 
+export function extractOutputHtmlFileName(
+  line: string,
+  defaultValue: string,
+): string {
+  return (
+    line.match(/body="(.{8}-.{4}-.{4}-.{4}-.{12}).htm"/)?.[1] ?? defaultValue
```
clangsmith, line 65: This is probably ok for now, but at some point in the future, we should probably use the ODS IOM events instead (namely ODSComplete), similar to the way EG and SAS Studio handle. Example C# code:

ODS_1_1 _odsService = (ODS_1_1)ws.ODS;
CIODSFileEvents_1_1_Event _odsFileEventsFor93 = ws.ODS as CIODSFileEvents_1_1_Event;
_odsService.FileOpen += HandleOdsFileOpen;
_odsService.FileClose += HandleOdsFileClose;
_odsService.DirectoryBegin += HandleOdsDirectoryBegin;
_odsService.AnchorElement += HandleOdsAnchorElement;
_odsFileEventsFor93.ODSComplete += HandleOdsComplete;

scottdover: Hey @clangsmith . Thanks for this. I linked this comment to the IOM library panel issue and will take a look when I'm working on that.

scottdover:
Hey @scnwwu / @Zhirong2022 . Do you both mind taking one more look at this one you get a chance? I've been working with Joe and Casey to test/resolve all the things, but just want to get your final blessing before merging things in

Zhirong2022:
I went through all the issues raised in PR. 
And now only two left for verification:
1. Unreadable Chinese characters, the root cause is that I use SAS Dark theme to render the specific SAS code. So, it is nothing to do with OS locale, VSC language configure or locale94/IOM connection.
2. As @clangsmith  mentioned the server configured with encrypted communications (specifically, Triple DES algorithm). It cannot connect to the server. I have no idea how to do verification.

client/src/connection/itc/script.ts
```diff
@@ -106,6 +124,29 @@ class SASRunner{
...
+    $objStream = $objFile.OpenBinaryStream(1);
+    [Byte[]] $bytes = 0x0
+
+    $endOfFile = $false
+    $byteCount = 0
+    $outStream = [System.IO.StreamWriter] $outputFile
+    do
+    {
+      $objStream.Read(1024, [ref] $bytes)
+      $outStream.Write([System.Text.Encoding]::UTF8.GetString($bytes))
```
scnwwu, line 140: This is a problem. A character in UTF8 may occupy various bytes (from 1 to 4 bytes). The 1024th byte may not be an end of a character. So when you try to get string from the 1024 bytes, the last character may be broken. And when it happens, the first character get from the next 1024 bytes will also be broken.
I believe it is the cause of the issue Zhirong is seeing. It only happens at cases where 1024 happened to cut a character.
It is a regression compared to v1.5 as it also affects local connection. In v1.5 it directly read local file thus no such issue. So I think it would be better to fix it in this PR.
We should collect all bytes before converting it to string.

scottdover:
> As @clangsmith mentioned the server configured with encrypted communications (specifically, Triple DES algorithm). It cannot connect to the server. I have no idea how to do verification.

Fwiw, we believe this one is now resolved. @clangsmith tested this yesterday I believe

client/src/connection/itc/script.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/ad48cd15c82de5327db9f50411a6e69992178f93

clangsmith:
> > As @clangsmith mentioned the server configured with encrypted communications (specifically, Triple DES algorithm). It cannot connect to the server. I have no idea how to do verification.
> 
> Fwiw, we believe this one is now resolved. @clangsmith tested this yesterday I believe

Correct. I have tested it and it is now (after Scott's fix) working as expected for me -- the ODS results are now appearing. I played around with that server using our extension for awhile and did not encounter any other issues.

Zhirong2022:
The issue that Chinese characters were not rendered correctly has been fixed in the latest package. However I found a performance issue with coding running.
1.Suppose I ran the same sample code against old package and new package. 
old package: 5-8 seconds to show result page
new package 50-1miniute30seconds  to show result page
2. If try to run the given code below to show a big table with same IOM connection.
```
proc print data=maps.africa;

```
old package: 1minute50seconds  to show result table 
new package: sometimes it will keep showing 'SAS code running' without any response.
At such scenario, try to cancel the running.  The running man icon will be disabled even witching to a new profile.

scottdover:
> The issue that Chinese characters were not rendered correctly has been fixed in the latest package. However I found a performance issue with coding running. 1.Suppose I ran the same sample code against old package and new package. old package: 5-8 seconds to show result page new package 50seconds-1miniute30seconds to show result page 2. If try to run the given code below to show a big table with same IOM connection.
> 
> ```
> proc print data=maps.africa;
> ```
> 
> old package: 1minute50seconds to show result table new package: sometimes it will keep showing 'SAS code running' without any response. At such scenario, try to cancel the running. The running man icon will be disabled even switching to a new profile.

This was one of my concerns in making a change to "collect all the bytes" first (in https://github.com/sassoftware/vscode-sas-extension/pull/592/commits/ad48cd15c82de5327db9f50411a6e69992178f93). Before this commit, we were reading bytes and writing them directly to a file. _Now_, we are storing all of the bytes in memory (which we'll eventually exhaust) and then attempting to encode all of them at once (which again, could be a problem for a huge number of bytes). 

I'm not sure the best approach to take here. It seems the options are:
 - Write to the file system as often as possible, keeping a lower memory overhead but ending up with some encoding issues, or...
 - Write to the file system once, with a much higher memory footprint but no encoding issues

@clangsmith @smorrisj @scnwwu Thoughts on how to proceed, here?

scnwwu:
> > The issue that Chinese characters were not rendered correctly has been fixed in the latest package. However I found a performance issue with coding running. 1.Suppose I ran the same sample code against old package and new package. old package: 5-8 seconds to show result page new package 50seconds-1miniute30seconds to show result page 2. If try to run the given code below to show a big table with same IOM connection.
> > ```
> > proc print data=maps.africa;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > old package: 1minute50seconds to show result table new package: sometimes it will keep showing 'SAS code running' without any response. At such scenario, try to cancel the running. The running man icon will be disabled even switching to a new profile.
> 
> This was one of my concerns in making a change to "collect all the bytes" first (in [ad48cd1](https://github.com/sassoftware/vscode-sas-extension/commit/ad48cd15c82de5327db9f50411a6e69992178f93)). Before this commit, we were reading bytes and writing them directly to a file. _Now_, we are storing all of the bytes in memory (which we'll eventually exhaust) and then attempting to encode all of them at once (which again, could be a problem for a huge number of bytes).
> 
> I'm not sure the best approach to take here. It seems the options are:
> 
> * Write to the file system as often as possible, keeping a lower memory overhead but ending up with some encoding issues, or...
> * Write to the file system once, with a much higher memory footprint and potential for failed runs, but no encoding issues
> 
> @clangsmith @smorrisj @scnwwu Thoughts on how to proceed, here?

Some thoughts:

1. How about simply read files with Nodejs for local connection (same as main branch)? Currently it already checks `this._config.protocol === ITCProtocol.COM` to resolve file path. I don't know if it will be faster than the IOM approach. If we're not able to make a perfect solution for remote IOM, at least it's not a regression compared to v1.5.
2. How about replace `System.IO.StreamWriter` with `System.IO.FileStream`? So we read bytes and write bytes thus can be safely sliced. I guess this is a desired way if it works, as it supports text and non-text files uniformly.
3. If the above doesn't work, how about replace `$objFile.OpenBinaryStream` with `$objFile.OpenTextStream`? It *looks* like designed for characters.

clangsmith:
> > > The issue that Chinese characters were not rendered correctly has been fixed in the latest package. However I found a performance issue with coding running. 1.Suppose I ran the same sample code against old package and new package. old package: 5-8 seconds to show result page new package 50seconds-1miniute30seconds to show result page 2. If try to run the given code below to show a big table with same IOM connection.
> > > ```
> > > proc print data=maps.africa;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > old package: 1minute50seconds to show result table new package: sometimes it will keep showing 'SAS code running' without any response. At such scenario, try to cancel the running. The running man icon will be disabled even switching to a new profile.
> > 
> > 
> > This was one of my concerns in making a change to "collect all the bytes" first (in [ad48cd1](https://github.com/sassoftware/vscode-sas-extension/commit/ad48cd15c82de5327db9f50411a6e69992178f93)). Before this commit, we were reading bytes and writing them directly to a file. _Now_, we are storing all of the bytes in memory (which we'll eventually exhaust) and then attempting to encode all of them at once (which again, could be a problem for a huge number of bytes).
> > I'm not sure the best approach to take here. It seems the options are:
> > 
> > * Write to the file system as often as possible, keeping a lower memory overhead but ending up with some encoding issues, or...
> > * Write to the file system once, with a much higher memory footprint and potential for failed runs, but no encoding issues
> > 
> > @clangsmith @smorrisj @scnwwu Thoughts on how to proceed, here?
> 
> Some thoughts:
> 
> 1. How about simply read files with Nodejs for local connection (same as main branch)? Currently it already checks `this._config.protocol === ITCProtocol.COM` to resolve file path. I don't know if it will be faster than the IOM approach. If we're not able to make a perfect solution for remote IOM, at least it's not a regression compared to v1.5.
> 2. How about replace `System.IO.StreamWriter` with `System.IO.FileStream`? So we read bytes and write bytes thus can be safely sliced. I guess this is a desired way if it works, as it supports text and non-text files uniformly.
> 3. If the above doesn't work, how about replace `$objFile.OpenBinaryStream` with `$objFile.OpenTextStream`? It _looks_ like designed for characters.

@scottdover and I stepped through the EG code and we are now doing essentially the same thing that EG is (pseudo-code): 

```
binaryStream = SAS.IFileref.OpenBinaryStream(Mode.StreamOpenModeForReading)
while (true)
{
	binaryStream.Read(CommonDownloadHelper.ChunkSize, out data); // EG ChunkSize is 8192
	if (data == null || data.Length == 0)
		break;
	fileStream.Write((byte[])data, 0, data.Length);
}
```
The previous issue is that we were encoding the byte chunks as we received them, which would break up DBCS characters if they fell on the border.

The testing I have done since the last fix appears to have resolved the issue. I also feel good knowing this is now the same way that EG handles it, considering EG has been doing it for many years (field-proven experience).

(Btw, given that printing maps.africa generates a 20MB results file, I consider that to be an edge-case scenario. EG doesn't even attempt to open results files larger than 5MB by default. I still tried this scenario in the extension with the latest changes, and it was successful, but took about a minute total before I saw the results. This is not unreasonable in my opinion.)

@scnwwu, @Zhirong2022, when you all return next week, please take another look, and if you feel this is reasonably addressed, don't encounter any other serious issues, or have any other concerns, please feel free to go ahead and merge this in.

Zhirong2022:
All issues have been fixed properly.

Zhirong2022 (approved):

scottdover:
Thanks, @scnwwu and @Zhirong2022 🎉

PR #615, opened 2023-11-14
chore: dependabot for /client /server and github-actions

scnwwu:
**Summary**
Looks like our current dependabot doesn't cover `/client` and `/server`. Guess it's a good idea to also keep them updated.

**Testing**
N/A


smorrisj (approved):
LGTM

PR #648, opened 2023-11-27
feat: ods graphics on; in wrapper

smorrisj:
**Summary**
Adds ods graphics on; to ods wrapper code to better align with SAS Studio

**Testing**

1. Run a program  (example below).
```
title 'Shoe Sales Over $500,000';
data work.shoes;
   set sashelp.shoes;
   where sales>500000;
run;
```
2. Check the Output Log. Verify that ods graphics on; appears in the output log as a line.

### DIFF
--- a/client/src/components/utils/sasCode.ts
+++ b/client/src/components/utils/sasCode.ts
@@ -47,7 +47,7 @@ export function wrapCodeWithOutputHtml(code: string): string {
   if (outputHtml) {
     const htmlStyleOption = generateHtmlStyleOption();
     return `title;footnote;ods _all_ close;
-ods graphics / imagemap;
+ods graphics on;
 ods html5${htmlStyleOption} options(bitmap_mode='inline' svg_mode='inline');
 ${code}
 ;*';*";*/;run;quit;ods html5 close;`;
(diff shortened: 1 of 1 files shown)


scnwwu:
Looks like SAS Studio doesn't wrap this. Do we need to make a settings option for user to choose whether they want it? I don't know. Just want to raise this question.

scottdover (approved):
The code looks to me, assuming things are good w/ PM (i.e. The option @scnwwu mentioned)

smorrisj:
> Looks like SAS Studio doesn't wrap this. Do we need to make a settings option for user to choose whether they want it? I don't know. Just want to raise this question.

One alternative is that it can get added to the autoexec on profile options if it's something that we dont want to add to the wrapper.

scnwwu (approved):
Looks like ods graphics is by default on. So there's not a difference unless user explicitly turned it off. I may not understand the use case for this. But since it matches SAS Studio, I'm good with it😄

Zhirong2022:
ods graphics on; appears in the log output as a line.

Zhirong2022 (approved):

smorrisj:
> Looks like ods graphics is by default on. So there's not a difference unless user explicitly turned it off. I may not understand the use case for this. But since it matches SAS Studio, I'm good with it😄

Yep! In my testing it was already enabled by default. The addition is more or less to just sync with SAS Studio's preamble.

PR #654, opened 2023-11-29
Italian Localisation

LRocc:
**Summary**
Added localisation changed for the italian language to bundle.l10n.it.json and package.nls.it.json 
**Testing**
booted the debug environment to check if the localisation matched and consulted the dictionary extensively


l10n/bundle.l10n.it.json
```diff
@@ -0,0 +1,121 @@
+{
+  "The SAS session has closed.": " La sessione SAS è stata chiusa.",
```
scnwwu, line 2: I have no knowledge of Italian. I saw leading whitespace in several messages in this file. Just want to double check is it intentional?

LRocc: No actually :?, i think this might have been a formatting error when pasting it from the shared word document we used, i'll rectify it

scnwwu (approved):

PR #664, opened 2023-12-03
fix(core): ods html5 output path to work directory

ShuguangSun:
**Summary**
Add `ods html5;` to function `wrapCodeWithOutputHtml` and make it similar to SAS EG.

The reason to make the change is that the SAS will generate html files in th 'current' diretory instead of 'work' directory. For exampl, I can change the current directory with `%put %sysfunc(dlgcdir("&__prjpath./SAS"));`, and then SAS will generate the html file in the path of '&__prjpath./SAS', however, vscode-sas-extension wants find the file in the work directory.

There is a function in 'client/src/connection/com/index.ts' which try to correct the ods output path and depends on `ods html5;`
```ts
      //write ODS output to work so that the session cleans up after itself
      const codeWithODSPath = code.replace(
        "ods html5;",
        `ods html5 path="${this._workDirectory}";`,
      );
```

**Testing**
Using the COM in main branch or IOM in sas9-itc branch

```sas
%put %sysfunc(dlgcdir("&__prjpath./SAS"));
proc print data=sashelp.cars(obs=10) noobs;
run;
```

### DIFF
--- a/client/src/components/utils/sasCode.ts
+++ b/client/src/components/utils/sasCode.ts
@@ -53,7 +53,7 @@ export function wrapCodeWithOutputHtml(code: string): string {
   const activeProfile = profileConfig.getActiveProfileDetail();
   const outputDestination =
     activeProfile &&
-      activeProfile.profile.connectionType !== ConnectionType.Rest
+    activeProfile.profile.connectionType !== ConnectionType.Rest
       ? ` body="${v4()}.htm"`
       : "";
 
--- a/client/src/components/utils/sasCode.ts
+++ b/client/src/components/utils/sasCode.ts
@@ -61,9 +61,9 @@ export function wrapCodeWithOutputHtml(code: string): string {
     const htmlStyleOption = generateHtmlStyleOption();
     return `title;footnote;ods _all_ close;
 ods graphics on;
-ods html5${htmlStyleOption} options(bitmap_mode='inline' svg_mode='inline')${outputDestination};
+ods html5(id=_vscode_sas)${htmlStyleOption} options(bitmap_mode='inline' svg_mode='inline')${outputDestination};
 ${code}
-;*';*";*/;run;quit;ods html5 close;`;
+;*';*";*/;run;quit;ods html5(id=_vscode_sas) close;`;
   } else {
     return code;
   }
--- a/client/src/components/utils/sasCode.ts
+++ b/client/src/components/utils/sasCode.ts
@@ -53,7 +53,7 @@ export function wrapCodeWithOutputHtml(code: string): string {
   const activeProfile = profileConfig.getActiveProfileDetail();
   const outputDestination =
     activeProfile &&
-    activeProfile.profile.connectionType !== ConnectionType.Rest
+      activeProfile.profile.connectionType !== ConnectionType.Rest
       ? ` body="${v4()}.htm"`
       : "";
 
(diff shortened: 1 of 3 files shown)


ShuguangSun:
Add `(id=--vscode-sas-extension-htm)` to the html5 which make the temporary file collect all outputs until the explicite closing. So we can make other ods html/html5 or `ods html5 close` or `ods listing on` in any where of the code.

The issue left is in `onShellStdOut`, the html5FileName below catch the file name from the **last** `ods html5` setting. However, we want the **very first** one.
```ts
        this._html5FileName =
          line.match(/NOTE: .+ HTML5.* Body.+: (.+)\.htm/)?.[1] ??
          this._html5FileName;
```

scnwwu:
So looks like the regexp can be enhanced to include the id, something like
```
new RegExp(`NOTE: .+ HTML5(${id}).* Body.+: (.+)\.htm`)
```

ShuguangSun:
Test code:
```sas
proc print data=sashelp.cars(obs=10) noobs;
  title "The block prior to ods html5 spec.";
run;

ods html5 file="abc.html";
proc print data=sashelp.cars(obs=10) noobs;
  title "The block with ods html5 spec";
run;
ods html5 close;

proc print data=sashelp.cars(obs=10) noobs;
  title "The block after ods html5 spec and ods html5 close.";
run;
```

client/src/connection/com/index.ts
```diff
@@ -114,12 +114,17 @@ export class COMSession extends Session {
 
       //write ODS output to work so that the session cleans up after itself
       const codeWithODSPath = code.replace(
-        "ods html5;",
-        `ods html5 path="${this._workDirectory}";`,
+        /\bods html5\b([^;]*;)/i,
+        `ods html5(id=_vscode_sas) path="${this._workDirectory}"$1`,
```
scnwwu, line 118: I guess it would be good to wrap the id in the `wrapCodeWithOutputHtml` function.
It's different to the `path=` which specific to IOM. The test case for id also applies to Viya.

ShuguangSun: Now make the change of id to `wrapCodeWithOutputHtml`

scottdover (approved):
This looks okay to me

scnwwu:
Hi @ShuguangSun, thank you very much for your patience. I just got feedback from the team and we'd like to have the id be just "vscode", that is `ods html5(id=vscode)`. Do you mind make it? Thank you!

scnwwu:
@ShuguangSun, please let me know if you'd like any help

ShuguangSun:
I'll update it this weekend.

ShuguangSun:
Updated but not sure the test.

client/src/components/utils/sasCode.ts
```diff
@@ -53,17 +53,17 @@ export function wrapCodeWithOutputHtml(code: string): string {
   const activeProfile = profileConfig.getActiveProfileDetail();
   const outputDestination =
     activeProfile &&
-    activeProfile.profile.connectionType !== ConnectionType.Rest
+      activeProfile.profile.connectionType !== ConnectionType.Rest
```
scnwwu, line 56: the build is failing due to format check. Have you updated dependencies (Prettier version in particular) and npm run format on this? looks like this line of change is not needed

ShuguangSun: fixed now.

scnwwu:
Hi @ShuguangSun, the test script will also need to be updated with id=vscode to pass
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/test/connection/itc/index.test.ts#L163

Zhirong2022:
The result page is displayed as expected if the user specify a personal work directory for html5 output.

Zhirong2022 (approved):

scnwwu:
@ShuguangSun, great contribution! Cheers!

PR #677, opened 2023-12-06
chore: update locale tool to support absolute paths

scottdover:
**Summary**
Self-explanatory

**Testing**
 - Tested running script with relative and absolute file paths

### DIFF
--- a/tools/locale.mjs
+++ b/tools/locale.mjs
@@ -1,7 +1,7 @@
 import { readFileSync, writeFileSync } from "fs";
 import glob from "glob";
 import csv from "papaparse";
-import { dirname, join } from "path";
+import { dirname, isAbsolute, join } from "path";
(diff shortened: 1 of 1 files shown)


tools/locale.mjs
```diff
@@ -61,7 +61,10 @@ const csvTranslationMap = (source) => {
     return {};
   }
 
-  const csvData = readFileSync(join(csvPath, importCSV)).toString();
+  const csvData = readFileSync(
+    // User can choose to read from an absolute path, or a relative path
+    importCSV.startsWith("/") ? importCSV : join(csvPath, importCSV),
```
scnwwu, line 66: Looks like not work for Windows. How about the `isAbsolute` function in the `path` module?

scnwwu (approved):

Issue #647, opened 2023-11-27
SAS Output Log channel is continually focused periodically for long running executions hijacking focus from other panels while the submission runs


smorrisj:
**Describe the bug**
As a user, I'd like to be able to use other panels (such as terminal) uninterrupted while waiting for my long-running SAS Program to finish its execution without losing focus.

**Steps to reproduce**

1. Run the following SAS code:
```
  data _null_;
      slept=sleep(10,1);
  run;
  
  proc print data=sashelp.cars;
  run;
  
  data _null_;
      slept=sleep(10,1);
  run;
  
  proc print data=sashelp.air;
  run;
```
2. Click on terminal and interact with the CLI.
3. Observe that as log output is flushed, focus is automatically shifted back to Output/SAS Log.
4. Click on the terminal and continue interaction.
5. Observe that subsequent log output flushes also steal focus.

**Expected behavior**
Right now we're calling "show" on the LogChannel component for every chunk/LogLine that gets flushed to the OutputChannel. This call will reveal our SAS Log Output Channel in the UI, taking focus away from whatever the user might be already doing. This will become especially annoying for programs that have heavy log output. 

Ideally there would be some profile options to allow the user to tailor the experience here.  One option to "focus log on beginning of execution" and perhaps another option to focus on the output window once execution has finished.

scnwwu:
We're calling `outputChannel.show(true)` where the `true` parameter tells the output to NOT take focus. You can keep typing code in the editor while the log append, it won't take your focus.
It's just that the Terminal by default sits together with Output. So to just "show" the Output (not taking focus), it has to hide Terminal.
A workaround is to drag the Terminal or Output to the right side pane, so that they're separated, like
https://github.com/sassoftware/vscode-sas-extension/issues/581#issuecomment-1791818069

snlwih:
We definitely need to provide the user with an option to tailor their experience to some extend. Supporting some settings for the SAS extension to specify if/when SAS Output or Table Viewer (once we have that added for IOM connections) should automatically open when new content is available, and perhaps even a suboption to specify if some of those tabs/panels should be allowed to take focus when new content is available in that tab.

See https://github.com/sassoftware/vscode-sas-extension/issues/651 for a related feature request.

PR #713, opened 2023-12-19
Additional Options for showing log output panel

smorrisj:
**Summary**
Currently the SAS Log output window is focused for any call made to `LogFn` during execution. While editor focus remains, if a user is interacting with tabs on the bottom panel (terminal for example), focus will be forcibly moved to the SAS Log output for every call to `LogFn`. This PR exposes two new options that will allow the user to specify when the SAS Log Output is focused to tailor the experience:

  `onExecutionStart` - boolean - show Log Output Window when code starts executing? Default: true
  `onExecutionFinish`- boolean - show Log Output Window when code has finished executing? Default: true

User Settings JSON structure:

```json
"logFocus": {
  "onExecutionStart": true,
  "onExecutionFinish": true 
}
```

Additionally, I think that it's time to start considering a more formalized structure for how we manage state in the extension. As such, part of this PR proposes to introduce [zustand](https://github.com/pmndrs/zustand/tree/main) as the state management library. It's lightweight, easy to setup, and we can leverage it for both the react webviews and the non-react components of the extension. The previous `running` state that we're tracking in the run command is moved to a zustand store so that we can update and react to changes to it in a simple manner using observers.

**Tasks**
- [x] introduce zustand store
- [x] create store for run state
- [x] switch run variable state over to store
- [x] subscribe to running state to conditionally show log on execution state changes
- [ ] add new fields to profile
- [ ] doc updates

**Testing**
How did you test this change?


package.nls.json
```diff
@@ -50,6 +50,8 @@
   "configuration.SAS.flowConversionMode": "Choose the conversion mode for notebooks",
   "configuration.SAS.flowConversionModeNode": "Convert each notebook cell to a node",
   "configuration.SAS.flowConversionModeSwimlane": "Convert each notebook cell to a swimlane",
+  "configuration.SAS.logFocus.onExecutionFinish": "Focus SAS Log when Code Execution is finished",
```
scnwwu, line 53: Suggest to say "show" instead of "focus". It does not focus as my [comment](https://github.com/sassoftware/vscode-sas-extension/issues/647#issuecomment-1829154110) explained.
Also, I'm not sure why "Code Execution" needs uppercase.

client/profileExamples/viya4.json
scnwwu, line 1: where is it used?

client/src/components/LogChannel.ts
```diff
@@ -4,14 +4,14 @@ import { OutputChannel, l10n, window } from "vscode";
 
 import { appendLog } from "./LogViewer";
 
-let outputChannel: OutputChannel;
+let log: OutputChannel;
```
scnwwu, line 7: In my opinion, `outputChannel` is a better name than `log`. When I see `log`, I'm confused is it log strings or anything.

client/src/components/LogChannel.ts
```diff
@@ -4,14 +4,14 @@ import { OutputChannel, l10n, window } from "vscode";
...
+if (!log) {
+  log = window.createOutputChannel(l10n.t("SAS Log"), "sas-log");
```
scnwwu, line 9: Simply `const log =` if it means to initialize here.
In my opinion, it would be better to be on-demand.

client/src/components/LogChannel.ts
```diff
@@ -4,14 +4,14 @@ import { OutputChannel, l10n, window } from "vscode";
...
+}
+export { log };
```
scnwwu, line 11: how about subscribe the store inside this component and not necessary to export it

client/src/commands/run.ts
```diff
@@ -34,7 +39,23 @@ interface FoldingBlock {
...
 
-let running = false;
+const { getState, subscribe: onRunStatusChange } = runStore;
+const { isRunning, setIsRunning } = getState();
+
+onRunStatusChange(
+  (state) => state.isRunning,
+  (newValue, oldValue) => {
+    if (newValue && !oldValue) {
+      if (isFocusLogOnExecutionStart()) {
```
scnwwu, line 49: I would suggest to change this option to be "Show output channel when SAS log received", which matches current behavior on main branch.
It doesn't make sense to show output channel if no SAS log received, for example authentication failed.
It makes sense to show output channel if user opens the library pane and got session startup logs.

client/profileExamples/viya4.json
smorrisj: It will be referred to from the documentation. I thought it would be good to build up a repository of settings.json examples that users could refer to.

client/src/components/LogChannel.ts
smorrisj: fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/a3aff3515f0aebd91ff1b1ee9ffecaaad9a5b3fe

client/src/components/LogChannel.ts
smorrisj: Moved back to on demand in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/0ce419ea20fce295cf620a7f6b83f9a0c0ff9104

client/src/components/LogChannel.ts
smorrisj: Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/0ce419ea20fce295cf620a7f6b83f9a0c0ff9104

package.nls.json
smorrisj: See renames in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/0ce419ea20fce295cf620a7f6b83f9a0c0ff9104. Are those names sufficient or should we explore different ones?

scottdover (approved):
This looks good to me, assuming it's good w/ @scnwwu

smorrisj:
Just noticed that the variable `isRunning` thats retrieved from the store is not bound, meaning that updates to it via setting store state do not update the locally scoped variable value. Need to look into that. It's entirely possible that we might need to subscribe to state updates for that.

smorrisj:
> Just noticed that the variable `isRunning` thats retrieved from the store is not bound, meaning that updates to it via setting store state do not update the locally scoped variable value. Need to look into that. It's entirely possible that we might need to subscribe to state updates for that.

Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/7471bd95d1573c6f3df2616289de4081fea9aea9. For the non-react components, if we want to have reactive behavior then we'll need to subscribe to state changes.

client/src/commands/run.ts
scnwwu: any comment to this one?

client/profileExamples/viya4.json
scnwwu: nit: an alternative is to put it into doc/ directory so that it's clear that it's just for doc.

client/src/commands/run.ts
```diff
@@ -34,7 +36,14 @@ interface FoldingBlock {
...
 }
 
-let running = false;
+const { isRunning: initialRunState, setIsRunning } = runStore.getState();
+
+let isCodeRunning: boolean = initialRunState;
+
+subscribe(
+  (state) => state.isRunning,
+  (run) => (isCodeRunning = run),
```
scnwwu, line 45: nit: just want to mention an alternative is to call `getState` each time, as it's not really reactive to do anything.

smorrisj: Good feedback. Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/ad4374265039a8fb7dec2e66b729eee2cb994ede

client/profileExamples/viya4.json
smorrisj: Good feedback. Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/ad4374265039a8fb7dec2e66b729eee2cb994ede

client/src/commands/run.ts
scnwwu: I agree with @clangsmith. I meant to say we need to wait for the first log recieved after run start, otherwise it will show even no log recieved.
Another thing is how a user choose to show the session startup log if they don't run any code but opens the library pane?

smorrisj: All good feedback. https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/82d208aa91034a323949f13fa131e4cfbbdc82df refactors the "on start" option to eval whether or not to show the channel only when the first chunk is written to `LogFn`

client/src/components/logViewer/LogChannel.ts
```diff
@@ -0,0 +1,50 @@
...
+  setProducedLogOutput();
+};
+
+subscribe(
+  (state) => state.isRunning,
+  (running, prevRunning) => {
+    if (!running && prevRunning) {
+      if (showLogOnExecutionFinish()) {
+        outputChannel?.show(true);
+        unsetProducedLogOutput();
```
scnwwu, line 35: This call should be independant of `if (showLogOnExecutionFinish())`, right?

Another thing is that, when running starts, that is
`if (running && !prevRunning)`
it should **also** `unsetProducedLogOutput()`
because, if user expanded the library view and produced log, there's no "done running" happened.

smorrisj: Thanks @scnwwu. I've refactored the approach to split run and log into separate "slices" of the global store to clean things up a bit. I've also split the session log and code execution log action into separate zustand actions so that we can evaluate the "show log" options for code execution log since it seemed like we didnt want to be showing the output panel for session log outputs. See https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/89aca224ac760fe0ed04cc79d8baf8c980113f6d. Still working on writing tests for the actions and initial state pieces.

client/src/components/logViewer/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    for (let i = 0; i < dataTokens.length; i++) {
+      if (legend.tokenTypes.includes(dataTokens[i])) {
+        tokensBuilder.push(document.lineAt(i).range, dataTokens[i]);
+      }
+    }
+    return tokensBuilder.build();
+  },
+};
+
+useStore.subscribe(logSelectors.selectIsOutputChannelOpen, (open) => {
```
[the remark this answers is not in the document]
smorrisj: That is something I wrestled with. It's more of a flag indicating that we should open the output channel. Another alternative is to just have the output channel itself be state in the store, and just use commands to open it. I was trying to keep vscode constructs/imports out of the store, but I do agree this looks weird.

client/src/components/logViewer/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    outputChannel.show(true);
+  }
+});
+
+useStore.subscribe(logSelectors.selectLogLines, (lines, prevLines) => {
+  if (!outputChannel) {
+    outputChannel = window.createOutputChannel(l10n.t("SAS Log"), "sas-log");
+  }
+
+  const delta = lines.filter((x) => !prevLines.includes(x));
```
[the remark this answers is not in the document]
smorrisj: Lines here will be the entire new log lines state. We could try to use `OutputChannel.replace` here if we wanted to go that route (assuming we keep the output channel declaration in logViewer), but we would need to stringify the entire state before calling replace. On the first pass I had tried that, but it seemed inefficient to replace the entire contents of the OutputChannel for every log update.

client/src/store/store.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { AppStoreState, initialState } from "./initialState";
+import { createLogActions } from "./log/actions";
+import { LogActions } from "./log/actions";
+import { logger } from "./middleware";
+import { createRunActions } from "./run/actions";
+import { RunActions } from "./run/actions";
+
+export type AppActions = LogActions & RunActions;
+export type AppStore = AppActions & AppStoreState;
```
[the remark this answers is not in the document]
smorrisj: Zustand recommends having a single store. I was using the slicing pattern defined [here](https://github.com/pmndrs/zustand/blob/main/docs/guides/slices-pattern.md). I agree that it feels kind of weird, but was trying to follow their best practices. I do like the idea of feature areas only needing to know about the state that they need to read/mutate.

client/src/store/selectors.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright © 2022-2024, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export { logSelectors } from "./log/selectors";
+// export { runSelectors } from "./run/selectors";
```
scottdover, line 5: Looks like we can clean this up for now

client/src/store/store.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const store: StateCreator<AppStore, []> = (...params) => ({
+  ...initialState,
+
+  ...createLogActions(...params),
+  ...createRunActions(...params),
+});
+
+export const useStore = createStore<AppStore>()(
+  subscribeWithSelector(logger(store)),
```
scottdover, line 24: Having a bunch of `console.info` statements in our log doesn't seem ideal for "production" environments. How often do we dump things to the console? Should this be something that is limited to development?

client/src/components/logViewer/index.ts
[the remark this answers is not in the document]
smorrisj: One motivating factor for capture log lines as state is that it would allow us down the road to offer a download/export log action which users have been asking for. VSCode has a way to programmatically access it (In digging it seems that the output channel is treated internally as a text document to vscode itself) but having a store action to dump those out seems preferable to having to go rooting around in vscode's document manager?

smorrisj:
Thanks for review feedback. In the latest push I reorganized the store to split back out into a log and run store. As a result, the log viewer now reacts to running state and uses actions to update stores accordingly.

client/src/store/selectors.ts
smorrisj: Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/b53a6ba752b9c595d468a795ee1a79100fc84c3c.

client/src/store/store.ts
smorrisj: fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/b53a6ba752b9c595d468a795ee1a79100fc84c3c.

client/src/store/run/actions.ts
```diff
@@ -0,0 +1,31 @@
...
+  toggleIsExecutingCode: (isExecutingCode) => {
+    const prevExecuting = get().isExecutingCode;
+
+    if (!isExecutingCode && prevExecuting) {
+      if (showLogOnExecutionFinish()) {
+        get().toggleOutputLogVisible(true);
+      }
+
+      set({ producedExecutionOutput: false });
+    }
```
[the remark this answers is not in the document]
smorrisj: See updated approach in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/b53a6ba752b9c595d468a795ee1a79100fc84c3c.

client/src/store/store.ts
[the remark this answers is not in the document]
smorrisj: Split out stores in latest pushes: https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/b53a6ba752b9c595d468a795ee1a79100fc84c3c

client/src/components/logViewer/index.ts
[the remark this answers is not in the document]
smorrisj: See updated approach in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/b53a6ba752b9c595d468a795ee1a79100fc84c3c. I was able to get rid of the strange state.

client/src/commands/run.ts
```diff
@@ -167,18 +168,19 @@ async function runCode(selected?: boolean, uri?: Uri) {
...
+  toggleIsExecutingCode(true);
   commands.executeCommand("setContext", "SAS.running", true);
 
   await runCode(selected, uri)
     .catch((err) => {
       onRunError(err);
     })
     .finally(() => {
-      running = false;
+      toggleIsExecutingCode(false);
```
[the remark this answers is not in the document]
smorrisj: Changed name in latest pushes: https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/b53a6ba752b9c595d468a795ee1a79100fc84c3c.

smorrisj:
@scnwwu any comments on latest approach?

client/src/components/logViewer/index.ts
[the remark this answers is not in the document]
scnwwu: Hmm... to me, storing the log lines doesn't necessarily mean storing as a state. For example a LogViewer can store the log lines internally and provide "append" and "get" methods.
Actually regarding download/export, getting from vscode's document sounds not a bad idea. SAS log may be huge and it can reduce some memory?

package.json
```diff
@@ -419,6 +420,18 @@
...
             ]
+          },
+          "SAS.log.showOnExecutionStart": {
+            "order": 7,
+            "type": "boolean",
+            "default": true,
+            "description": "%configuration.SAS.log.showOnExecutionStart%"
+          },
+          "SAS.log.showOnExecutionFinish": {
+            "order": 8,
```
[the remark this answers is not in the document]
smorrisj: Both settings can be set to true at the same time, this is the default setting for both. We dont want to keep the current behavior, so we didnt provide a method for "always". I went with two explicit settings for simplicity.

client/src/store/log/actions.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const createLogActions: StateCreator<Store, [], [], LogActions> = (
+  set,
+  get,
+) => ({
+  onOutputLog: (logs) => {
+    const tokens = logs.map((x) => x.type);
+    set((existing) => ({
+      logTokens: [...existing.logTokens, ...tokens],
+    }));
```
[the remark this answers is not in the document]
smorrisj: We use two separate actions because we dont want to evaluate log options on session log output.

client/src/components/LibraryNavigator/LibraryModel.ts
```diff
@@ -25,7 +25,7 @@ class LibraryModel {
 
   public async connect(): Promise<void> {
     const session = getSession();
-    session.onLogFn = LogChannelFn;
+    session.onExecutionLogFn = appendSessionLogFn;
```
scnwwu, line 28: why not
```suggestion
    session.onSessionLogFn = appendSessionLogFn;
```

client/src/commands/run.ts
```diff
@@ -226,10 +232,10 @@ export async function runTask(
       await session.cancel();
     }
 
-    running = false;
+    setIsExecutingCode(false);
     commands.executeCommand("setContext", "SAS.running", false);
   });
-  session.onLogFn = onLog ?? LogChannelFn;
+  session.onExecutionLogFn = onLog ?? appendExecutionLogFn;
```
scnwwu, line 238: I guess it's also good to add
```
session.onSessionLogFn = appendSessionLogFn;
```
for task

scnwwu (commented):
In general good to me. Only 2 small comments

client/src/connection/itc/index.ts
```diff
@@ -319,7 +319,7 @@ export class ITCSession extends Session {
           line,
           this._html5FileName,
         );
-        this._onLogFn?.([{ type: "normal", line }]);
+        this._onExecutionLogFn?.([{ type: "normal", line }]);
```
scnwwu, line 322: shall we distinguish session log here? perhaps
```
if (this._workDirectory) {
    this._onExecutionLogFn?.([{ type: "normal", line }]);
} else {
    this._onSessionLogFn?.([{ type: "normal", line }]);
}

Comment (smorrisj) [reply client/src/connection/itc/index.ts]: fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/6d6d1e0dd3d0938d427f339603b093b34cf7d2f0

File: client/src/commands/run.ts (hunk 18 above)
Comment (smorrisj) [reply client/src/commands/run.ts:239]: fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/6d6d1e0dd3d0938d427f339603b093b34cf7d2f0

File: client/src/components/LibraryNavigator/LibraryModel.ts (hunk 17 above)
Comment (smorrisj) [reply client/src/components/LibraryNavigator/LibraryModel.ts]: fixed in https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/6d6d1e0dd3d0938d427f339603b093b34cf7d2f0

**Review by scnwwu — APPROVED** [review]
Good to me. Thank you!

**Review by scottdover — APPROVED** [review]

**Comment by Zhirong2022** [comment]
Output panel is shown unexpectedly in certain case.
```  
 "SAS.log.showOnExecutionStart": true,
 "SAS.log.showOnExecutionFinish": true
```
Reproduce step:
1. Sign in with a valid profile with Viya connection to show Explorer and Library
2. Switch to a profile with Viya connection, but the endpoint is not available and fail to connect. It will show error prompt as expect.
3. Stay in Terminal panel
4. Run some SAS code

Result: It will show Output panel unexpectedly, however there is no log text written on it.

smorrisj:
> Output panel is shown unexpectedly in certain case.
> 
> ```
>  "SAS.log.showOnExecutionStart": true,
>  "SAS.log.showOnExecutionFinish": true
> ```
> 
> Reproduce step:
> 
> 1. Sign in with a valid profile with Viya connection to show Explorer and Library
> 2. Switch to a profile with Viya connection, but the endpoint is not available and fail to connect. It will show error prompt as expect.
> 3. Stay in Terminal panel
> 4. Run some SAS code
> 
> Result: It will show Output panel unexpectedly, however there is no log text written on it.

For step number 2 are you switching to the bad profile and then pressing the run button?

smorrisj:
> Output panel is shown unexpectedly in certain case.
> 
> ```
>  "SAS.log.showOnExecutionStart": true,
>  "SAS.log.showOnExecutionFinish": true
> ```
> 
> Reproduce step:
> 
> 1. Sign in with a valid profile with Viya connection to show Explorer and Library
> 2. Switch to a profile with Viya connection, but the endpoint is not available and fail to connect. It will show error prompt as expect.
> 3. Stay in Terminal panel
> 4. Run some SAS code
> 
> Result: It will show Output panel unexpectedly, however there is no log text written on it.

Thanks @Zhirong2022. Fixed in latest push https://github.com/sassoftware/vscode-sas-extension/pull/713/commits/fb31638a05bfce0b80f543ee3db7f0cf39aff7b6.

Zhirong2022:
The raised issue has been fixed.  No other concerns.

Zhirong2022 (approved):

PR #733, opened 2024-01-03
chore: create library adapter, make library model api agnostic

scottdover:
**Summary**
This refactors LibraryDataProvider in preparation for getting things working with iom/com connections. Here's the general architecture changes made as part of this pr:
 - The concept of a `LibraryAdapter` has been introduced along with `RestLibraryAdapter`. `LibraryAdapter` is an interface that is used by LibraryModel to manage library/table data
 - `LibraryAdapterFactory` has been added to create `LibraryAdapter`s. At the moment, this always creates a `RestLibraryAdapter`, but will eventually create a `ItcLibraryAdapter`
 - `LibraryModel`s functionality has been limited to: working with the file system, paginating api requests, and further processing library adapter results for easier consumption by `LibraryDataProvider`

**Testing**
 - [x] Tested deleting a library
 - [x] Tested refreshing libraries
 - [x] Tested viewing multiple tables
 - [x] Tested drag and drop with table
 - [x] Tested exporting table as CSV

### DIFF
--- a/client/src/components/LibraryNavigator/LibraryAdapterFactory.ts
+++ b/client/src/components/LibraryNavigator/LibraryAdapterFactory.ts
@@ -1,5 +1,7 @@
-// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// Copyright © 2024, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+import { ProgressLocation, l10n, window } from "vscode";
+
 import RestLibraryAdapter from "../../connection/rest/RestLibraryAdapter";
 import { ConnectionType } from "../profile";
 import { LibraryAdapter } from "./types";
@@ -9,8 +11,20 @@ class LibraryAdapterFactory {
     switch (connectionType) {
       case ConnectionType.Rest:
       default:
-        return new RestLibraryAdapter();
+        return new RestLibraryAdapter(this.emitConnectionNotification);
     }
+  }
+
+  private async emitConnectionNotification(
+    callback: () => Promise<void>,
+  ): Promise<void> {
+    await window.withProgress(
+      {
+        location: ProgressLocation.Notification,
+        title: l10n.t("Connecting to SAS session..."),
+      },
+      callback,
+    );
   }
 }
 
--- a/client/src/components/LibraryNavigator/LibraryAdapterFactory.ts
+++ b/client/src/components/LibraryNavigator/LibraryAdapterFactory.ts
@@ -1,7 +1,5 @@
 // Copyright © 2024, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { ProgressLocation, l10n, window } from "vscode";
-
 import RestLibraryAdapter from "../../connection/rest/RestLibraryAdapter";
 import { ConnectionType } from "../profile";
 import { LibraryAdapter } from "./types";
@@ -11,20 +9,8 @@ class LibraryAdapterFactory {
     switch (connectionType) {
       case ConnectionType.Rest:
       default:
-        return new RestLibraryAdapter(this.emitConnectionNotification);
+        return new RestLibraryAdapter();
     }
-  }
-
-  private async emitConnectionNotification(
-    callback: () => Promise<void>,
-  ): Promise<void> {
-    await window.withProgress(
-      {
-        location: ProgressLocation.Notification,
-        title: l10n.t("Connecting to SAS session..."),
-      },
-      callback,
-    );
   }
 }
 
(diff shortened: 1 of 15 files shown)


client/src/components/LibraryNavigator/LibraryAdapterFactory.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
```
scnwwu, line 1: Well, 2024 now 😄

client/src/components/LibraryNavigator/index.ts
```diff
@@ -29,7 +31,7 @@ class LibraryNavigator implements SubscriptionProvider {
   constructor(context: ExtensionContext) {
     this.extensionUri = context.extensionUri;
     this.libraryDataProvider = new LibraryDataProvider(
-      new LibraryModel(),
+      new LibraryModel(this.activeProfileConnectionType()),
```
scnwwu, line 34: nit: just wondering is it a good idea to have LibraryModel taking care of the active connection type internally?

client/src/connection/rest/RestLibraryAdapter.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    await window.withProgress(
+      {
+        location: ProgressLocation.Notification,
+        title: l10n.t("Connecting to SAS session..."),
+      },
+      async () => {
+        await session.setup();
+      },
+    );
```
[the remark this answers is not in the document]
scottdover: I'm glad you mentioned this one. I _do_ need to think of a better way to address this. Ideally, I think it might not be a bad idea to do a better job isolating vscode-isms from our code in "deeper" layers

client/src/components/LibraryNavigator/index.ts
scottdover: Not a bad idea. So maybe instead of `LibraryDataProvider.refresh(connectionType: ConnectionType)`, we have something more like `LibraryDataProvider.useAdapter(adapter: LibraryAdapter)` that gets passed down to the model?

client/src/connection/rest/RestLibraryAdapter.ts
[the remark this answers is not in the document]
scottdover: Done in https://github.com/sassoftware/vscode-sas-extension/pull/733/commits/586f7636bb230e75b7bd39b4458020f006df3c88

client/src/components/LibraryNavigator/index.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/733/commits/93d3eff37b0ed5e1302e7be4f63a9cbe3a4ac847

client/src/connection/rest/RestLibraryAdapter.ts
```diff
@@ -0,0 +1,248 @@
...
+    protected readonly emitConnectionNotification: (
+      callback: () => Promise<void>,
+    ) => Promise<void>,
+  ) {}
+
+  public async connect(): Promise<void> {
+    const session = getSession();
+    session.onLogFn = LogChannelFn;
+
+    await this.emitConnectionNotification(async () => await session.setup());
```
scnwwu, line 38: nit: these 2 lines are still independent of connection type, that is, will be repeated in your itc adapter, right?
If so, how about make it into a base class?
Well actually, I just realized they're also repeated in run command. So perhaps they can be wrapped into the setup function in the base Session class.

scottdover: I'll look at making them part of the base Session class

scottdover: I updated `setup` in the base session class in https://github.com/sassoftware/vscode-sas-extension/pull/733/commits/6b4022f0efc7cb99b92b05911410905188e3b1b8

scnwwu (commented):
- In main branch, when running as a task, it emits to terminal rather than notification.
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/commands/run.ts#L235
So looks like it has to be changed to `session.establishConnection()`.
It made me think that is there a better name for contributors to easily understand when to call `setup` vs. `establishConnection`? For example `setupWithProgress`/`setup`, `setup(silent?: boolean)`.

- The progress call in notebook can be removed now.
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/notebook/Controller.ts#L44

client/src/connection/session.ts
```diff
@@ -8,7 +10,17 @@ export abstract class Session {
...
   }
 
-  abstract setup(): Promise<void>;
+  async setup(): Promise<void> {
+    await window.withProgress(
+      {
+        location: ProgressLocation.Notification,
+        title: l10n.t("Connecting to SAS session..."),
+      },
+      async () => await this.establishConnection(),
```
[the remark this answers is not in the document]
scottdover: D'oh. Updated in https://github.com/sassoftware/vscode-sas-extension/pull/733/commits/c9087c958ecca535d61bc3e7fe00cae3542f361f

scnwwu (approved):
Good to me. Thank you

scottdover:
@Zhirong2022 do you mind testing? This PR didn't add any new features but refactored how we handle libraries/tables. We'll just need to make sure no regressions were introduced.

Zhirong2022:
No regression issues.

Zhirong2022 (approved):

PR #786, opened 2024-01-30
fix: pin node version to avoid compatibility issues

smorrisj:
**Summary**
@vscode/l10n-dev is not compatible with current lts node. This PR pins the version to a known version that works.

**Testing**
Run action locally. L10n files needing localization are now properly identified. Those keys will get updated using our normal L10 process.


scottdover (approved):
Lgtm https://github.com/sassoftware/vscode-sas-extension/actions/runs/7713731625/job/21024320825

PR #813, opened 2024-02-07
feat: SQL/Python/Lua syntax highlight

jingjiajie:
**Summary**
Below four procs are supported syntax highlighting.
&nbsp;&nbsp;**proc sql, proc fedsql, proc python, proc lua**

In this PR only syntax highlighting is supported.
Auto completion, hovering, formatting, etc. will be done in the future but is not included in this PR.

**Testing**

/*
proc python;

proc sql;
&nbsp;&nbsp;select "

cards
*/

proc python restart echo;
submit;
&nbsp;&nbsp;def func(param: str) -> str:
&nbsp;&nbsp;&nbsp;&nbsp;return param
&nbsp;&nbsp;if \_\_name\_\_ == '\_\_main\_\_':
&nbsp;&nbsp;&nbsp;&nbsp;print(func("foo"))
&nbsp;&nbsp;else:
&nbsp;&nbsp;&nbsp;&nbsp;pass
endsubmit;
quit;

proc lua terminate;
submit;
&nbsp;&nbsp;function func()
&nbsp;&nbsp;&nbsp;&nbsp;local var = "hello world"
&nbsp;&nbsp;end
&nbsp;&nbsp;print(func())
endsubmit;
quit;

proc fedsql;
&nbsp;&nbsp;select * from t1
&nbsp;&nbsp;left join t2 on t1.foo = t2.bar
quit;

proc sql;
&nbsp;&nbsp;select * from t1
&nbsp;&nbsp;left join t2 on t1.foo = t2.bar
run;


scottdover:
Hey @jingjiajie . Do you mind fixing build failures and rebasing?

jingjiajie:
I looked at the build error, it's caused by failure of code formatter unit test. I checked out to _main_ branch and tried a similar case , and found a similar error:

```
proc lua;
submit;

endsubmit;
run;
```


Click "Format Document", then the code changes to:

```
proc lua;
    submit;


```

@scnwwu Would you mind having a look?

server/src/sas/Lexer.ts
```diff
@@ -235,37 +244,120 @@ export class Lexer {
...
+          if (
+            this.context.embeddedLangState ===
+              EmbeddedLangState.PROC_PYTHON_SUBMIT_OR_INTERACTIVE ||
+            this.context.embeddedLangState ===
+              EmbeddedLangState.PROC_LUA_SUBMIT_OR_INTERACTIVE
+          ) {
+            endReg =
+              /\b(endsubmit|endinteractive|quit|run)(\s+|\/\*.*?\*\/)*;(\s+|\/\*.*?\*\/)*$/i;
+          } else {
+            endReg = /\b(run|quit)(\s+|\/\*.*?\*\/)*;(\s+|\/\*.*?\*\/)*$/i;
```
scnwwu, line 310: proc sql will only end by `quit` not `run`

Another thing is that, proc can end without explicit quit, when another proc met, refers to sample code
https://go.documentation.sas.com/doc/en/pgmsascdc/v_048/sqlproc/n1ncn0pznd8wrln1tnp3xdxjz9xz.htm

server/src/sas/Lexer.ts
```diff
@@ -235,37 +244,120 @@ export class Lexer {
...
+        if (token.type === "sep" && token.text === ";") {
+          let endReg;
+          if (
+            this.context.embeddedLangState ===
+              EmbeddedLangState.PROC_PYTHON_SUBMIT_OR_INTERACTIVE ||
+            this.context.embeddedLangState ===
+              EmbeddedLangState.PROC_LUA_SUBMIT_OR_INTERACTIVE
+          ) {
+            endReg =
+              /\b(endsubmit|endinteractive|quit|run)(\s+|\/\*.*?\*\/)*;(\s+|\/\*.*?\*\/)*$/i;
```
scnwwu, line 308: should not check quit/run until met endsubmit/endinteractive
Refers to sample code
https://go.documentation.sas.com/doc/en/pgmsascdc/v_048/proc/n01tofnov0sz98n1adq2qarcofrc.htm

scnwwu:
> I looked at the build error, it's caused by failure of code formatter unit test. I checked out to _main_ branch and tried a similar case , and found a similar error:
> 
> ```
> proc lua;
> submit;
> 
> endsubmit;
> run;
> ```
> 
> Click "Format Document", then the code changes to:
> 
> ```
> proc lua;
>     submit;
> ```
> 
> @scnwwu Would you mind having a look?

Thank you for reporting this. I made a PR #849 to fix this. Once it is included in this branch, the build will pass. In the meantime, you can continue work on other issues in this PR.

scnwwu (commented):
Please try sample code at
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/testFixture/formatter/expected.sas
The highlight looks incorrect with this PR.

scnwwu:
With this PR, some SAS specific syntax in PROC SQL will have different colors to other places.
Please refer to sample code at
https://go.documentation.sas.com/doc/en/pgmsascdc/v_048/sqlproc/n1ncn0pznd8wrln1tnp3xdxjz9xz.htm
See format like `date7.`, date literal like `'26JUN70'd`, statement keyword like `title`

scnwwu:
#849 merged. Please rebase with main branch. Thanks.

server/src/sas/Lexer.ts
jingjiajie: Seems it can end by run:  https://go.documentation.sas.com/doc/en/pgmsascdc/v_048/proc/p0o588yrrlokr3n1osm419b7hdtg.htm

scnwwu: If you try run
```
proc sql;
    select * from sashelp.cars;
run;
    select * from sashelp.cars;
```
You'll find that `run` will not end sql proc; The 2nd select statement still works. You'll got below note from SAS log:
`NOTE: PROC SQL statements are executed immediately; The RUN statement has no effect.`

jingjiajie:
> With this PR, some SAS specific syntax in PROC SQL will have different colors to other places. Please refer to sample code at https://go.documentation.sas.com/doc/en/pgmsascdc/v_048/sqlproc/n1ncn0pznd8wrln1tnp3xdxjz9xz.htm See format like `date7.`, date literal like `'26JUN70'd`, statement keyword like `title`

Fixed.

jingjiajie:
> Please try sample code at https://github.com/sassoftware/vscode-sas-extension/blob/main/client/testFixture/formatter/expected.sas The highlight looks incorrect with this PR.

Fixed.

scnwwu:
Try below code
```
/* comment */
proc sql outobs=2;
  title 'Densities of Countries';
  select name format=$20.
      from sql.newcountries;
/* comment */
proc sql outobs=2;
  title 'Densities of Countries';
  select name format=$20.
      from sql.newcountries;
```
Issue 1, colors for `title`, string, comment, format are different
Issue 2, mouse hover the first `outobs=`, the help can appear, but no help on the 2nd `outobs=`
Issue 3, similar to autocomplete, if you type spacebar after the first `proc sql`, the autocomplete list for proc sql options will appear, but not for the 2nd `proc sql`

scnwwu:
Incorrect syntax colors with below code
```
proc sql;
   title 'Weighted Averages from Sample Data';
   select * from sashelp.cars;
```

scnwwu:
Incorrect syntax colors with below code (See `]]`)
```
proc lua;
submit;
    local code = [[
        data sample;
        run;
    ]]
endsubmit;
run;
```

jingjiajie:
> Try below code
> 
> ```
> /* comment */
> proc sql outobs=2;
>   title 'Densities of Countries';
>   select name format=$20.
>       from sql.newcountries;
> /* comment */
> proc sql outobs=2;
>   title 'Densities of Countries';
>   select name format=$20.
>       from sql.newcountries;
> ```
> 
> Issue 1, colors for `title`, string, comment, format are different Issue 2, mouse hover the first `outobs=`, the help can appear, but no help on the 2nd `outobs=` Issue 3, similar to autocomplete, if you type spacebar after the first `proc sql`, the autocomplete list for proc sql options will appear, but not for the 2nd `proc sql` Issue 4, try `Format Document`, some code lost. The formatter has to be adjusted.



> Incorrect syntax colors with below code
> 
> ```
> proc sql;
>    title 'Weighted Averages from Sample Data';
>    select * from sashelp.cars;
> ```



> Incorrect syntax colors with below code (See `]]`)
> 
> ```
> proc lua;
> submit;
>     local code = [[
>         data sample;
>         run;
>     ]]
> endsubmit;
> run;
> ```

Fixed.

scnwwu:
@jingjiajie, there're still difference for the colors, see below, compare line 1 to line 4, line 2 to line 7

Capture1

scnwwu:
wrong color with code
```
*
'''
my proc python
'''
;
proc python;
interactive;
print('first statement after for loop')
endinteractive;
run;
```

```
proc
python;
interactive;
print('first statement after for loop')
endinteractive;
run;
```

```
proc sql;
%put my data;
  describe table dictionary.members;
```

jingjiajie:
> wrong color with code
> 
> ```
> *
> '''
> my proc python
> '''
> ;
> proc python;
> interactive;
> print('first statement after for loop')
> endinteractive;
> run;
> ```

Fixed.

> ```
> proc
> python;
> interactive;
> print('first statement after for loop')
> endinteractive;
> run;
> ```

Maybe this can't be fixed by textmate. Here is the regex of _proc python_, I tried to add "\n" between _proc_ and _python_ in the regex, but it doesn't work. It seems the regex in the textmate doesn't support matching across multiple lines. If that's true, this case can hardly be fixed.
image


> ```
> proc sql;
> %put my data;
>   describe table dictionary.members;
> ```

Fixed the SQL highlighting bug, but the global stmt _%put_ has difficulty to be highlighted yet. See my last comment above.

scnwwu:
> > @jingjiajie, there're still difference for the colors, see below, compare line 1 to line 4, line 2 to line 7
> > Capture1
> 
> > proc sql;
> > %put my data;
> > describe table dictionary.members;
> 
> @scnwwu I meet an issue: the SQL highlighting is implemented by textmate definition. textmate can't hand over to language server when meeting a SAS global stmt inside the SQL code. To highlight SAS global statment inside SQL, we need to write SAS global stmt syntax into the SQL textmate file. I have three ideas:
> 
> 1. Just highlight SAS keywords in the SAS global stmt inside SQL
>    This is easiest way, although not very accurate, it can provide a basic highlight for user
> 2. Write a full SAS global stmt syntax into the textmate definition
>    This could be a heavy work, and I can't promise it can accurately support all SAS global statments because
>    a. Regex has many limitations of abilities.
>    b. proc sql allows not only global stmts, but also any macro stmts, it has too many possibilities.
> 3. Don't use textmate. Just treat PROC SQL as SAS language, and use SAS language server to parse all the SQL language.
>    This can reuse all the existing abilities to parse global stmts, and macro stmts. But the heavy work is to implement parsing SQL language in the language server.
> 
> I just came up with idea 4: 4. Use both language server and textmate for PROC SQL. Textmate will make the SQL highlighting, but leave the SAS global stmts, macro stmts unrecognized. Language server can highlight the SAS global stmts, macro stmts, but leaves the SQL stmts unrecognized. This seems to be the most economic way. The only concern is that if there is a SQL keyword is the same as a SAS keyword, the language server will give a fault semantic highlighting. I'm not very sure if this case exists.
> 
> Please let me know if there's any other ideas.

@jingjiajie, would you try https://github.com/microsoft/vscode-textmate?
It would be good if we can get textmate result within the language server, and provide all tokens from language server. So that we have all control over it. And other language clients can simply get full result from language server without worrying about textmate.

jingjiajie:
> > > @jingjiajie, there're still difference for the colors, see below, compare line 1 to line 4, line 2 to line 7
> > > Capture1
> > 
> > 
> > > proc sql;
> > > %put my data;
> > > describe table dictionary.members;
> > 
> > 
> > @scnwwu I meet an issue: the SQL highlighting is implemented by textmate definition. textmate can't hand over to language server when meeting a SAS global stmt inside the SQL code. To highlight SAS global statment inside SQL, we need to write SAS global stmt syntax into the SQL textmate file. I have three ideas:
> > 
> > 1. Just highlight SAS keywords in the SAS global stmt inside SQL
> >    This is easiest way, although not very accurate, it can provide a basic highlight for user
> > 2. Write a full SAS global stmt syntax into the textmate definition
> >    This could be a heavy work, and I can't promise it can accurately support all SAS global statments because
> >    a. Regex has many limitations of abilities.
> >    b. proc sql allows not only global stmts, but also any macro stmts, it has too many possibilities.
> > 3. Don't use textmate. Just treat PROC SQL as SAS language, and use SAS language server to parse all the SQL language.
> >    This can reuse all the existing abilities to parse global stmts, and macro stmts. But the heavy work is to implement parsing SQL language in the language server.
> > 
> > I just came up with idea 4: 4. Use both language server and textmate for PROC SQL. Textmate will make the SQL highlighting, but leave the SAS global stmts, macro stmts unrecognized. Language server can highlight the SAS global stmts, macro stmts, but leaves the SQL stmts unrecognized. This seems to be the most economic way. The only concern is that if there is a SQL keyword is the same as a SAS keyword, the language server will give a fault semantic highlighting. I'm not very sure if this case exists.
> > Please let me know if there's any other ideas.
> 
> @jingjiajie, would you try https://github.com/microsoft/vscode-textmate? It would be good if we can get textmate result within the language server, and provide all tokens from language server. So that we have all control over it. And other language clients can simply get full result from language server without worrying about textmate.

I think there are two problems if we directly use vscode-textmate:
1. Theme problem. Textmate returns tokens with _**scope**_ , and the theme is applied to the scope. If we use language server to parse the textmate grammar, what we return are the semantic tokens, so the theme configuration doesn't work.

2. We still need to write the same logic to identify if the line inside PROC SQL is a SAS global stmt or a SQL clause(For SAS global stmt, parse by language server, For SQL, parse by textmate). As we still need to do this, using https://github.com/microsoft/vscode-textmate saves no effort than the idea4 talked above.

So for now I still think the idea4 is the best compromise. But I still need to write codes to try if it really works.

jingjiajie:
@scnwwu I had a try. The idea4 seems to work well, and only less than 20 lines of related code are modified. It's easy enough as a solution. I just pushed the code, so you could validate it.

scnwwu:
@jingjiajie, your idea 4 looks good in general. Looks like now that semantic token can handle sas syntax in PROC SQL, for example date and format. Would you please check all your customizations in 'sassql' still needed?

package.json
```diff
@@ -917,6 +917,27 @@
       },
       {
         "id": "sas-log"
+      },
+      {
+        "id": "sassql"
```
scnwwu, line 922: Is there a chance to NOT declare a new language? Is it possible to just inject a textmate to source.sas?
https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#injection-grammars

package.json
```diff
@@ -917,6 +917,27 @@
...
+        "language": "sassql",
+        "scopeName": "source.sassql",
+        "path": "./syntaxes/sassql.tmLanguage.json"
+      },
+      {
+        "language": "sas",
+        "scopeName": "source.sas",
+        "path": "./syntaxes/sas.tmLanguage.json",
+        "embeddedLanguages": {
+          "source.sql": "sql",
```
scnwwu, line 936: is it still needed?

server/src/sas/Lexer.ts
```diff
@@ -233,8 +246,277 @@ export class Lexer {
...
+          let line = this.curr.line;
+          line < this.model.getLineCount();
+          line++
+        ) {
+          let lineContent = this.model.getLine(line);
+          if (line === this.curr.line) {
+            lineContent =
+              " ".repeat(this.curr.column) +
+              lineContent.slice(this.curr.column);
+          }
```
scnwwu, line 421: nit: looks like this piece of code appears multiple times. Any chance to extract it?

syntaxes/sassql.tmLanguage.json
```diff
@@ -0,0 +1,674 @@
...
+    },
+    "saskeywords": {
+      "patterns": [
+        {
+          "match": "(run|quit|%mend)(\\s|/\\*.*\\*/|\\*[^;]*;)*;$",
+          "name": "keyword.section.sas"
+        }
+      ]
+    },
+    "sas-macro-put": {
```
scnwwu, line 648: nit: how about more consistent names? for example
sas-keywords
sas-macro-put
sas-macro-quotes

scnwwu:
@jingjiajie, it would be great if you can try add some unit test for your added part. Thanks!

jingjiajie:
> @jingjiajie, it would be great if you can try add some unit test for your added part. Thanks!

@scnwwu I didn't find a test folder in the /server folder. It feels to me that unit tests for the language server can be more convenient if written into the /server folder, it enables unit test codes easily access Lexer, SyntaxProvider, CodeZoneManager, etc. which can't be directly accessed from /client. For me, I need to access the Lexer to test if the tokenizing of the PROC SQL, PROC PYTHON, and PROC LUA is correct, as well as SemanticTokens from the SyntaxProvider. Would we add a unit test folder into the /server folder?

Zhirong2022:
Acceptance Criteria:

It will cover three types of programming language in SAS code: Python, Lua and Sql
2.Make sure syntax highlighting works correctly for each language
3.Verify that there will be no syntax help for each language
4.Verify that code blocks folding function well for each language
5.Verify that code outline works as expected for each language
6.Verify that code formatting function properly for each language
7.Verify that 'run region' function well for each language
8.Make sure there are no issues for each language if having blocks nesting
9.Verify that SAS code inside Python/Lua constant have correct Python/Lua syntax color
10.Verify that syntax colors work for SAS or non-SAS themes
Sql:
11.Verify that the the valid SAS code is highlighted correctly
12.Verify that the valid SAS code has syntax help if it has
13.Verify 1 to 12 in OS: Windows and Mac
14.Verify 1 to 13 in NLS environment

Zhirong2022:
1.The color of some parameter and variable is not correct 
2.The color of comment should be #647f29
```
/* options ls=72; */
%let a=%nrquote(%nrstr(test%'s     message;));

data _null_;
  input name=$ / 1-8 $10. @ @15 @@ @(b*3) @weekday #2 +5 4. ? ?? name=$ / 1-8
    $10. @ @15 @@ @(b*3) @weekday #2 +5 4. ? ??;
  cards;
1  AA  2
3  BB  4
;
```
image

Zhirong2022:
No syntax help for the function char()
```
proc sql;
  create table unique proclib.paylist
  (IdNum char(4),
   Jobcode char(3)
  );
```

Zhirong2022:
Type code below, python code highlighting works as expected. However, if trying to comment line 3 (print('first statement after for loop')) through Ctrl+/,  it would be better to comment as Python format.
```
proc python;
interactive;
print('first statement after for loop') 
endinteractive;
run;
```
image

Zhirong2022:
No autocomplete if trying to type some global statements.
Please check the attached file to find the expected behavior.
autocomplete

jingjiajie:
@Zhirong2022 All reported bugs fixed.

themes/sas-light-color-theme.json
```diff
@@ -7,7 +7,7 @@
     "keyword": { "foreground": "#3578c5" },
     "sec-keyword": { "foreground": "#224c7c", "bold": true },
     "proc-name": { "foreground": "#224c7c", "bold": true },
-    "comment": { "foreground": "#65802a" },
+    "comment": { "foreground": "#647f29" },
```
scnwwu, line 10: @jingjiajie, please rebase with latest main branch. Thanks. This line is already on main branch and should not be listed as a change in this PR.

Zhirong2022:
There is no syntax help for 'python' and the autocomplete list data is not correct.
*
'''
my proc python
'''
;
proc python;
interactive;
print('first statement after for loop')
endinteractive;
run;
image

Zhirong2022:
'local' will be highlighted unexpectedly, it is a defined libname having no specific function
image

Zhirong2022:
It would be better to support customized region  as VSC does for the included python code.
'''
proc python infile='/tmp/my_script.py';
submit;
#region
my_function()
print("y = " + y)
#endregion
endsubmit;
run;
'''
image

Zhirong2022:
It has no autocomplete if trying to type submit/endsubmit, interactive/endinteractive
image

Zhirong2022:
Type 'ctrl+/' to comment code in 'proc sql', it should be commented as SAS language format.
image

jingjiajie:
> submit;

@scnwwu I found the root cause of this is an existing code, I'm not sure what's the purpose of this:
CodeZoneManager.ts  _getPrev():
a

When the first token is a comment, it returns the previous one, and there's not a previous one, so it returns null token. It caused _skipToStmtStart() error (should move to proc, but got a null token). I want to confirm is this behavior expected?

jingjiajie:
> There is no syntax help for 'python' and the autocomplete list data is not correct. * ''' my proc python ''' ; proc python; interactive; print('first statement after for loop') endinteractive; run; image

The quote is this

scnwwu:
> > submit;
> 
> @scnwwu I found the root cause of this is an existing code, I'm not sure what's the purpose of this: CodeZoneManager.ts _getPrev():
> When the first token is a comment, it returns the previous one, and there's not a previous one, so it returns null token. It caused _skipToStmtStart() error (should move to proc, but got a null token). I want to confirm is this behavior expected?

As it is reproducible on main branch and not specific to proc python, please report a separate issue. It will be addressed separately. Thanks.

jingjiajie:
> It would be better to support customized region as VSC does for the included python code. ''' proc python infile='/tmp/my_script.py'; submit; #region my_function() print("y = " + y) #endregion endsubmit; run; ''' image


@Zhirong2022 This is a heavier feature provided by language server. After we integrate the python language server, this feature would be naturally supported.

Others are fixed, excepted the separate issue Wei mentioned.

package-lock.json
```diff
@@ -594,1176 +428,479 @@
         "node": ">=10.0.0"
       }
     },
-    "node_modules/@esbuild/aix-ppc64": {
+    "node_modules/@esbuild/win32-x64": {
```
scnwwu, line 431: @jingjiajie, the changes in this file look far beyond your change in the `package.json` file. Would you please install with existing lock file first, then only apply the corresponding change in `package.json` file?

server/package.json
```diff
@@ -11,4 +11,4 @@
     "vscode-languageserver": "^9.0.1",
     "vscode-languageserver-textdocument": "^1.0.11"
   }
-}
+}
```
scnwwu, line 14: please do not change this file if not needed. Thanks.

server/test/index.ts
scnwwu, line 1: please remove empty files. thanks

syntaxes/sassql.tmLanguage.json
```diff
@@ -0,0 +1,682 @@
...
+      "end": "\\)|(?=\\n)",
+      "name": "sas.string.quoted.macro",
+      "patterns": [
+        {
+          "include": "#sas-macro-quotes"
+        }
+      ]
+    },
+    "sas-global-stmt": {
+      "begin": "(?i)(libname|((title|footnote|axis|legend|pattern|synbol)\\d{0,}))\\b",
```
scnwwu, line 672: did you mean `symbol`?

server/src/sas/SyntaxDataProvider.ts
```diff
@@ -1981,7 +1979,11 @@ export class SyntaxDataProvider {
       return data;
     });
   }
-  getProcedureStatements(procName: string, cb?: (data: string[]) => void) {
+  getProcedureStatements(
+    procName: string,
+    noGlobal: boolean,
```
scnwwu, line 1984: nit:
```suggestion
    noGlobal?: boolean,
```

Zhirong2022:
No syntax help for interactive/endinteractive, submit/endsubmit with typing and mouse hovering. It will show syntax help by typing Ctrl+Space.
```
proc python;
  interactive;
  endinteractive;
proc lua;
  submit;
  endsubmit;
```
image
image

Zhirong2022:
All issues have fixed.

Zhirong2022 (approved):

PR #815, opened 2024-02-08
fix: prevent viya from generating guid based html files

scottdover:
**Summary**
This prevents viya connections from generating uuid based files

**Testing**
 - [x] Made sure viya results worked as expected
 - [x] Made sure SAS 9 worked as expected via iom


smorrisj:
Tested this out locally this afternoon. Looks good on viya connections.

smorrisj (approved):

scottdover:
To test this, we'll just need to make sure ods html results work as expected for viya and non-viya connections.

Zhirong2022:
It works well with Viya and SAS9 IOM connection.

Zhirong2022 (approved):

PR #842, opened 2024-02-21
feat: add better error reporting for COM/IOM connections

scottdover:
**Summary**
This adds better error support for IOM & COM connections. Specifically, it provides _better_ errors for:
 - being on the wrong plaform (i.e. Mac)
 - using an incorrect username / password
 - using an incorrect host
 - using an incorrect port (or having a connection where a sas server isn't present)
 - trying to run code that will trigger transcoding errors

**Testing**
 - [x] test being on the wrong plaform (i.e. Mac)
 - [x] test using an incorrect username / password
 - [x] test using an incorrect host
 - [x] test using an incorrect port (or having a connection where a sas server isn't present)
 - [x] test trying to run code that will trigger transcoding errors

### DIFF
--- a/client/src/connection/itc/LineParser.ts
+++ b/client/src/connection/itc/LineParser.ts
@@ -2,34 +2,39 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export class LineParser {
-  protected errorLines: string[] = [];
-  protected capturingError: boolean = false;
+  protected processedLines: string[] = [];
+  protected capturingLine: boolean = false;
 
   public constructor(
     protected startTag: string,
     protected endTag: string,
+    protected returnNonProcessedLines: boolean,
   ) {}
 
   public processLine(line: string): string | undefined {
-    if (line.includes(this.startTag) || this.capturingError) {
-      this.errorLines.push(line);
-      this.capturingError = true;
+    if (line.includes(this.startTag) || this.capturingLine) {
+      this.processedLines.push(line);
+      this.capturingLine = true;
       if (line.includes(this.endTag)) {
-        return this.processedError();
+        return this.processedLine();
       }
       return;
     }
 
-    return line;
+    return this.returnNonProcessedLines ? line : undefined;
   }
 
-  protected processedError(): string {
-    this.capturingError = false;
-    const fullError = this.errorLines
+  protected processedLine(): string {
+    this.capturingLine = false;
+    const fullError = this.processedLines
       .join("")
       .replace(this.startTag, "")
       .replace(this.endTag, "");
-    this.errorLines = [];
+    this.processedLines = [];
     return fullError;
+  }
+
+  public isCapturingLine(): boolean {
+    return this.capturingLine;
   }
 }
--- a/client/src/connection/itc/LineParser.ts
+++ b/client/src/connection/itc/LineParser.ts
@@ -1,0 +1,31 @@
+export class LineParser {
+  protected errorLines: string[] = [];
+  protected capturingError: boolean = false;
+  protected startTag: string;
+  protected endTag: string;
+
+  public constructor(startTag: string, endTag: string) {
+    this.startTag = startTag;
+    this.endTag = endTag;
+  }
+
+  public processLine(line: string): string | undefined {
+    if (line.includes(this.startTag) || this.capturingError) {
+      this.errorLines.push(line);
+      this.capturingError = true;
+      if (line.includes(this.endTag)) {
+        return this.processedError();
+      }
+      return;
+    }
+
+    return;
+  }
+
+  protected processedError(): string {
+    this.capturingError = false;
+    const fullError = this.errorLines.join('').replace(this.startTag, '').replace(this.endTag,'');
+    this.errorLines = [];
+    return fullError;
+  }
+}
--- a/client/src/connection/itc/LineParser.ts
+++ b/client/src/connection/itc/LineParser.ts
@@ -4,13 +4,11 @@
 export class LineParser {
   protected errorLines: string[] = [];
   protected capturingError: boolean = false;
-  protected startTag: string;
-  protected endTag: string;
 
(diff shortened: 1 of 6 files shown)


client/src/connection/itc/const.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright © 2024, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { l10n } from "vscode";
+
+export const humanReadableMessages = [
```
scottdover, line 5: Hey @clangsmith . Do we think it's worth having a resource to expand on what these error messages mean? Or perhaps we can tweak these messages to elaborate? 

Really, the only messages that look like they _could_ require more explanation are the errors on line 18 and 28 (imo).

client/src/connection/itc/index.ts
```diff
@@ -13,10 +20,13 @@ import {
...
 import { extractOutputHtmlFileName } from "../../components/utils/sasCode";
 import { Session } from "../session";
+import { LineParser } from "./LineParser";
+import { ERROR_END_TAG, ERROR_START_TAG } from "./const";
 import { scriptContent } from "./script";
 import { LineCodes } from "./types";
+import { decodeEntities } from "./util";
 
-const SECRET_STORAGE_NAMESPACE = "ITC_SECRET_STORAGE";
+const SECRET_STORAGE_NAMESPACE = "ITC_SECRET_STORAGE-1";
```
scnwwu, line 29: is this change necessary?

client/src/connection/itc/index.ts
```diff
@@ -280,22 +300,80 @@ export class ITCSession extends Session {
...
-        ),
-      ),
+      new Error(this.fetchHumanReadableErrorMessage(errorMessage)),
     );
+
     // If we encountered an error in setup, we need to go through everything again
-    if (/Setup error/.test(msg)) {
+    const fatalErrors = [/Setup error/, /powershell\.exe: command not found/];
+    if (fatalErrors.find((regex) => regex.test(errorMessage))) {
+      // If we can't even run the shell script (i.e. powershell.exe not found),
```
scnwwu, line 316: if the system language is not English, it may get translated "not found", isn't it? not sure if there's any better way to check this

client/src/connection/itc/LineParser.ts
```diff
@@ -0,0 +1,37 @@
...
+export class LineParser {
+  protected errorLines: string[] = [];
+  protected capturingError: boolean = false;
+  protected startTag: string;
+  protected endTag: string;
+
+  public constructor(startTag: string, endTag: string) {
+    this.startTag = startTag;
+    this.endTag = endTag;
+  }
```
scnwwu, line 13: nit: just a note that this can be simplified to
`public constructor(protected startTag: string, protected endTag: string) {}`
https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties
But I'm fine if you prefer this

client/src/connection/itc/index.ts
```diff
@@ -280,22 +300,80 @@ export class ITCSession extends Session {
...
       this._workDirectory = undefined;
     }
   };
 
+  private fetchHumanReadableErrorMessage = (msg: string): string => {
+    const atLineIndex = msg.indexOf("At line");
+    const errorMessage = atLineIndex ? msg.slice(0, atLineIndex) : msg;
+
+    // Dump error to console
+    console.warn("shellProcess stderr: " + errorMessage);
```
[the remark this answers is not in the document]
scottdover: Hmm. Perhaps. Fwiw, this was just copied from another location. That said, I think there was _some_ duplication in the console. Let me clean that up and see if we still see all the necessary information

client/src/connection/itc/index.ts
scottdover: d'oh. Updated in https://github.com/sassoftware/vscode-sas-extension/pull/842/commits/d7f28f15d5c4ec6925932a8f3ac546b28e816ecd

client/src/connection/itc/LineParser.ts
scottdover: I updated it to this format in https://github.com/sassoftware/vscode-sas-extension/pull/842/commits/d7f28f15d5c4ec6925932a8f3ac546b28e816ecd . That said, I'm not entirely sure how I feel about it. 

I think my reservation with this approach is that it requires removing the class attributes from the top of the class. For me, this makes it a little less clear what attributes are defined. That said, it's super minor and as long as the constructor is close enough to the rest of the attributes, it seems fine

client/src/connection/itc/index.ts
scottdover: That's a good point. We _could_ just isolate this to check `powershell\.exe`. I wonder if that'd be problematic. Fwiw, the _only_ error I've seen with `powershell\.exe` as part of the message is this one 🤷

scottdover: Okay, I simplified this in https://github.com/sassoftware/vscode-sas-extension/pull/842/commits/58babb4f1a20be6cc3590605cf35b0bfa2a5bbe4. It _seems_ to work well for running code successfully and w/ errors. 

That said, if we think it's _too_ generic, I can roll it back and just let this error fall through to the "default" error

client/src/connection/itc/index.ts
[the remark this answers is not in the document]
scottdover: So, there is some duplication here. However, this line prints the raw error from powershell (which could be useful for debugging), while the other only prints the formatted error. I think there's value in leaving this here.

scnwwu (approved):

ShermanWang0324:
testing_results

ShermanWang0324:
I have tested all the scenarios listed, all fixed messages are captured as above screenshot.

ShermanWang0324 (approved):

Issue #506, opened 2023-09-13
SAS Log color-coding in Output Window for SAS 9 IOM connections


PaulShannon-KatalyzeData:
SAS program logs are written to the VSCode output window in plain text, with no colour-coding for Errors, Notes and Warnings. Colouring of the Output channel is supported in VSCode from  [v1.66 (March 2022)](https://code.visualstudio.com/updates/v1_66#_output-channel-with-custom-language-id).

> In this milestone, we have finalized the [createOutputChannel API](https://github.com/microsoft/vscode/blob/dc2f5d8dd1790ac4fc6054e11b44e36884caa4be/src/vscode-dts/vscode.d.ts#L9415) to create an output channel with custom language ID. This will allow you to contribute token or syntax coloring and CodeLens features to your Output channel by passing a language ID

Suggest development for standard SAS colouring based on the theme applied (i.e.: dark/light theme etc.).

Further, the ability to configure custom highlights for customer-specific log entries perhaps using RegEx would be advantageous. For example: search for "variable is uninitialized" note and highlight with a user-defined colour.

scnwwu:
Hi @PaulShannon-KatalyzeData, the SAS Log is colored for Viya connection and SAS themes.
Are you requesting for non-SAS themes or non-Viya connections?

PaulShannon-KatalyzeData:
Hi,

I Haven’t seen the extension with Viya yet, I’m using it with SAS 9.4.

Many thanks,
Paul

scnwwu:
OK. I updated the issue title.
For SSH connection it is challenging. Only plain text is exchanged through SSH. RegEx may not work well as error/warning/info messages can across multiple lines and there's no explicit ending.
For COM/IOM it should be feasible.

PaulShannon-KatalyzeData:
Hi,

I’m not overly up to speed on VSCode extensions, but I believe they are written using NodeJS, something I’m not massively familiar with.

My thought RE: the ability to use RegEx would be that a user could provide some patterns in a JSON file, and these would be parsed locally within VSCode, meaning plain text read from SAS would be suitable.

However, if that’s not a realistic approach, I think standard colouring of logs would help encourage many of our clients to drop the use of Enterprise Guide in favour of VSCode.

Many thanks,
Paul

QuentinMcMullen:
+1 for needing log coloring when connected to local SAS 9.

[...]

Issue #841, opened 2024-02-21
Add matrix documenting capabilities supported for each server connection type


snlwih:
**Is your feature request related to a problem? Please describe.**
We have a number of capabilities that are not available across all connection types and this is not always clear from the documentation. 

**Describe the solution you'd like**
A matrix listing all the capabilities across one dimension and the connection types across the other dimension with a checkmark in the cell if that capability is supported for the specific connection type. 

**Describe alternatives you've considered**
For connection types we have information about what options (autoexec statement, configuration options) are supported for ech connection type. (see https://github.com/sassoftware/vscode-sas-extension/blob/main/connect-and-run.md#additional-settings-in-a-profile). We could add info there about whether capabilities like Table Viewer, SAS log color-coding, ... are supported. But it is not easy to find and many users would not look there.

A dedicated documentation file containing a matrix of features supported for each connection type is much easier to read and understand. That doc could then have links to dedicated doc with more details.

scnwwu:
I just took a try. Please see
https://scnwwu.github.io/vscode-sas-extension/matrix

jennifert-sas:
@scnwwu That looks very nice! I made a draft table, as well, with links to more details. Can we include links once we have settled on all the content that needs to be included? I'll email you a link to my document.
cc: @clangsmith @snlwih

clangsmith:
@scnwwu, I love the decision tree diagram! @jennifert-sas, I reviewed your draft doc and added some comments. Thank you both.

snlwih:
@scnwwu , can the columns showing the checkmarks be centered? Have added comments for the tree diagram in #865.

PR #843, opened 2024-02-22
feat: SAS log colors for Local 94 and IOM

scnwwu:
**Summary**
Support #506

**Testing**
See Error/Warning/Note colors for SAS log with Local 94 and IOM connections.

### DIFF
--- a/client/src/connection/itc/index.ts
+++ b/client/src/connection/itc/index.ts
@@ -5,7 +5,7 @@ import { Uri, env, l10n, window, workspace } from "vscode";
 import { ChildProcessWithoutNullStreams, spawn } from "child_process";
 import { resolve } from "path";
 
-import { BaseConfig, RunResult } from "..";
+import { BaseConfig, LogLineTypeEnum, RunResult } from "..";
 import {
   getGlobalStorageUri,
   getSecretStorage,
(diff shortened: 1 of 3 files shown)


scottdover (approved):

Zhirong2022:
Log type with Note, Error and Warning have appropriate colors if running SAS code with Local 94 and IOM connections.

Zhirong2022 (approved):

snlwih:
@Zhirong2022 , it seems this only works when the user has selected one of the SAS provided themes. I noticed that this capablity is not mentioned in the readme at all. Would it make sense to add this capability in the **_SAS log_** section of the readme (https://github.com/sassoftware/vscode-sas-extension/blob/main/README.md#sas-log) and then also mention for which connection types this is supported? 

Btw. when listing the connection types make sure to use the exact same phrases as shown in the UI.
image

scnwwu:
> @Zhirong2022 , it seems this only works when the user has selected one of the SAS provided themes. I noticed that this capablity is not mentioned in the readme at all. Would it make sense to add this capability in the **_SAS log_** section of the readme (https://github.com/sassoftware/vscode-sas-extension/blob/main/README.md#sas-log) and then also mention for which connection types this is supported? I think all connection types except for **SAS 9.4 (remote - SSH)**, correct?
> 
> Btw. when listing the connection types make sure to use the exact same phrases as shown in the UI. image

Just a note that the SAS theme requirement is not specific to this PR (same as Viya). There's a FAQ entry from the beginning.
https://github.com/sassoftware/vscode-sas-extension/wiki/FAQ#i-dont-see-errornote-colors-on-sas-log
We can definitely improve the doc via #841 and #865

Issue #699, opened 2023-12-14
Cannot run PROC IMPORT on remote Unix machine


rhstanton:
**Describe the bug**
Running VS Code on a Mac, connecting to SAS via SSH on a remote Unix machine, I can get some code to run (so the connection works), but I get an error message if I try to run "PROC IMPORT":

NOTE: Writing HTML5 Body file: sashtml.htm
NOTE: This SAS session is using a registry in WORK.  All changes will be lost 
      at the end of this session.
ERROR: The connection to the X display server could not be made. Verify that 
       the X display name is correct, and that you have access authorization. 
       See the online Help for more information about connecting to an X 
       display server.
ERROR: Device does not support full-screen.

**Steps to reproduce**
Run the line

proc import datafile=x out=_csv dbms=csv replace; run;

**Environment (please complete the following information):**
Client OS: MacOS 12.7.2
Extension version: 1.5.0
SAS version: 9.4

sasutils:
It looks like the remote SAS session is trying to open an X window on the remote machine.  Did you include the -noterminal option on the command you used to launch SAS?

PR #845, opened 2024-02-23
fix: remote unix machine x window issue when running specific sas cod…

scnnaw:
Cannot run PROC IMPORT on remote Unix machine (#699)

**Summary**
When using the SSH to login the Unix machine, we supplied a wrong SAS options to connect the session. 
The "-terminal" should be changed to "-noterminal".
(https://go.documentation.sas.com/doc/en/pgmsascdc/v_045/lesysoptsref/p1db6atd7mw5fnn1lvcra2s1i7tq.htm)

**Testing**
1. Add a new connection profile which support the SAS9.4 (remote -SSH); 
2. Try to run the line
proc import datafile=x out=_csv dbms=csv replace; run;
3. The output should *not* be:
ERROR: The connection to the X display server could not be made. Verify that
the X display name is correct, and that you have access authorization.
See the online Help for more information about connecting to an X
display server.
ERROR: Device does not support full-screen.

It should be: ERROR: Physical file does not exist.....

### DIFF
--- a/client/src/connection/ssh/index.ts
+++ b/client/src/connection/ssh/index.ts
@@ -219,7 +219,11 @@ export class SSHSession extends Session {
     const resolvedEnv: string[] = ['_JAVA_OPTIONS="-Djava.awt.headless=true"'];
     const execArgs: string = resolvedEnv.join(" ");
 
-    const resolvedSasOpts: string[] = ["-nodms", "-terminal", "-nosyntaxcheck"];
+    const resolvedSasOpts: string[] = [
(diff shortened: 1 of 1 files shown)


scnwwu:
adding @smorrisj who originally authored it for review

smorrisj (approved):

Zhirong2022:
We have no proper environment to reproduce issue (https://github.com/sassoftware/vscode-sas-extension/issues/699).  Approve the fix and the consumer can verify it after the release.

PR #880, opened 2024-03-08
feat: add libraries & tables for IOM/COM

scottdover:
**Summary**
This adds support for library/table viewing under IOM & COM connections. The general approach to getting the information we need is to:
 - Temporarily re-route session log output to a variable
 - Run some sql or a data step to get library & table info (similar to what EG does)
 - Parse data between a starting tag and end tag (since we don't care about _all_ input).

One somewhat hacky thing is parsing the SAS system line (On the en-us configuration I used, this said "The SAS System"). This line comes at the start of each new page of data, causing issues with data parsing. This is subject to change with any feedback in code review that may provide a better approach to solving this.

**Testing**
 - [x] Tested new functionality by starting with an empty set of profiles
 - [x] Tested viewing/refreshing libraries
 - [x] Tested viewing/refreshing tables
 - [x] Tested viewing table data (column types, row data, etc)
 - [x] Tested downloading data as a CSV
 - [x] Tested deleting tables
 - [x] Make sure log works as expected for regular sas program


client/src/connection/itc/index.ts
```diff
@@ -310,8 +300,9 @@ export class ITCSession extends Session {
         `
   do {
     $chunkSize = 32768
-    $count = $runner.FlushLogLines($chunkSize)
-  } while ($count -gt 0)\n
+    $log = $runner.FlushLog($chunkSize)
```
scottdover, line 303: Fix me! ...and re-test :)

client/src/connection/itc/index.ts
```diff
@@ -421,7 +411,12 @@ export class ITCSession extends Session {
    * @param data a buffer of stdout output from the child process.
    */
   private onShellStdOut = (data: Buffer): void => {
-    const output = data.toString();
```
scnwwu, line 424: It was intended to NOT trimEnd. At line 423, it will ignore empty lines. If it was trimed, empty lines will never be displayed. But we do need some empty lines to make the log look pretty.

package.json
```diff
@@ -975,30 +975,28 @@
...
           "name": "%views.SAS.signIn%",
-          "when": "!SAS.authorized && SAS.connectionType != rest"
+          "when": "!SAS.authorized && !SAS.canSignIn"
         }
       ]
     },
     "viewsWelcome": [
       {
         "view": "sas-content-get-started",
-        "when": "(!SAS.authorized && SAS.connectionType == rest) && !SAS.connection.direct",
```
scnwwu, line 995: im a little lost, why totally delete this? direct and SSH connection still don't need this, right?

package.json
```diff
@@ -975,30 +975,28 @@
...
         "contents": "%views.SAS.welcome%",
         "enablement": "!SAS.authorizing"
       },
       {
         "view": "sas-content-invalid-connection",
-        "when": "SAS.connectionType != rest",
```
scnwwu, line 1001: what about ssh?

scnwwu:
> Tested downloading data as a PDF

Did you mean CSV?

DmitryMK:
Hi @scottdover, thanks for this feature! I'm was testing this branch and noticed that for somedatasets, it creates the layout, but does not load any data.

I have 2 types of errors in devtools:

1) Unexpected end of JSON input. The test dataset is the following:

```
data test;
    length col1-col128 $12;
    array col [*] $ col:;
    do i = 1 to dim(col);
        col[i] = byte(i);
    end;
    output;
run;
```
The message is:

> mainThreadExtensionService.ts:81 [SAS.sas-lsp]Unexpected end of JSON input
> $onExtensionRuntimeError @ mainThreadExtensionService.ts:81
> S @ rpcProtocol.ts:456
> Q @ rpcProtocol.ts:441
> M @ rpcProtocol.ts:371
> L @ rpcProtocol.ts:297
> (anonymous) @ rpcProtocol.ts:161
> y @ event.ts:1148
> fire @ event.ts:1179
> fire @ ipc.net.ts:650
> x.onmessage @ localProcessExtensionHost.ts:374
> mainThreadExtensionService.ts:82 SyntaxError: Unexpected end of JSON input
>     at JSON.parse (<anonymous>)
>     at i_.getDatasetInformation (d:\Users\pic0dxk\.vscode\extensions\sas.sas-lsp-1.7.1\client\dist\node\extension.js:360:205)
>     at processTicksAndRejections (node:internal/process/task_queues:95:5)
>     at i_.getRows (d:\Users\pic0dxk\.vscode\extensions\sas.sas-lsp-1.7.1\client\dist\node\extension.js:326:181)
>     at f_.queryForData (d:\Users\pic0dxk\.vscode\extensions\sas.sas-lsp-1.7.1\client\dist\node\extension.js:360:5830)
>     at f_.getData (d:\Users\pic0dxk\.vscode\extensions\sas.sas-lsp-1.7.1\client\dist\node\extension.js:360:5571)
>     at e_.processMessage (d:\Users\pic0dxk\.vscode\extensions\sas.sas-lsp-1.7.1\client\dist\node\extension.js:308:152)


2. Unfortunately I cannot share the test data (it is 90 rows with ~20 variables). I tried to isolate the issue, but strangely it looks like it does not depend on a certain value (when I split the dataset into 2 parts, it loaded both parts without issues, but when trying to load the whole dataset it failed).

The message I was getting is

> mainThreadExtensionService.ts:81 [SAS.sas-lsp]Unterminated string in JSON at position 32758
> ... The rest is same as above

And sometimes the message is for the same dataset:

> mainThreadExtensionService.ts:81 [SAS.sas-lsp]Expected double-quoted property name in JSON at position 32758
> ... The rest is same as above

DmitryMK:
Here is the test dataset for the second type of issues.

```
data test;
    length longText $32758;
    longText = repeat("_",32757);
run;
```

Looks like when data for 100 rows (as I understand this is the hardcoded limit) is > 32758, it fails with this error.

julien-weinstein:
What's the status of this update? Would really like to see my libraries and tables with COM connection.

scottdover:
Hi @julien-weinstein. Thanks for checking in :). This task is currently on hold for the next day or so while I take care of other priorities.

If you're looking for a rough estimate of when this will be available, I think the most optimal timeline would be around 3 weeks (1 week to wrap up dev work, 1 week for testing, 1 week to gather translations/deploy)

julien-weinstein:
> Hi @julien-weinstein. Thanks for checking in :). This task is currently on hold for the next day or so while I take care of other priorities.
> 
> If you're looking for a rough estimate of when this will be available, I think the most optimal timeline would be around 3 weeks (1 week to wrap up dev work, 1 week for testing, 1 week to gather translations/deploy)

@scottdover Thank you so much for letting me know! That all sounds good, looking forward to the update! I'm just getting started with SAS and the SAS extension in VS Code, so do those updates happen automatically or will I have to manually update in VS Code?

scottdover:
> > Hi @julien-weinstein. Thanks for checking in :). This task is currently on hold for the next day or so while I take care of other priorities.
> > If you're looking for a rough estimate of when this will be available, I think the most optimal timeline would be around 3 weeks (1 week to wrap up dev work, 1 week for testing, 1 week to gather translations/deploy)
> 
> @scottdover Thank you so much for letting me know! That all sounds good, looking forward to the update! I'm just getting started with SAS and the SAS extension in VS Code, so do those updates happen automatically or will I have to manually update in VS Code?

You're very welcome 😊. Happy to have your feedback! I believe that extension updates happen automatically, unless you have them disabled (see https://code.visualstudio.com/docs/supporting/FAQ#_opt-out-of-extension-updates). 

If automatic updates are disabled, you can always manually update by clicking the update button on the extension page (https://code.visualstudio.com/docs/editor/extension-marketplace#_update-an-extension-manually)

julien-weinstein:
> > > Hi @julien-weinstein. Thanks for checking in :). This task is currently on hold for the next day or so while I take care of other priorities.
> > > If you're looking for a rough estimate of when this will be available, I think the most optimal timeline would be around 3 weeks (1 week to wrap up dev work, 1 week for testing, 1 week to gather translations/deploy)
> > 
> > 
> > @scottdover Thank you so much for letting me know! That all sounds good, looking forward to the update! I'm just getting started with SAS and the SAS extension in VS Code, so do those updates happen automatically or will I have to manually update in VS Code?
> 
> You're very welcome 😊. Happy to have your feedback! I believe that extension updates happen automatically, unless you have them disabled (see https://code.visualstudio.com/docs/supporting/FAQ#_opt-out-of-extension-updates).
> 
> If automatic updates are disabled, you can always manually update by clicking the update button on the extension page (https://code.visualstudio.com/docs/editor/extension-marketplace#_update-an-extension-manually)

Got it. Keep up the good work, you guys rock!

client/src/connection/itc/CodeRunner.ts
```diff
@@ -0,0 +1,92 @@
...
+    const outputLines = [];
+
+    const addLine = (logLines: LogLine[]) =>
+      outputLines.push(...logLines.map(({ line }) => line));
+
+    try {
+      await session.setup(true);
+
+      // Lets capture output to use it later
+      session.onSessionLogFn = addLine;
```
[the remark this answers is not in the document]
scottdover: I honestly wasn't super familiar w/ that. I just figured we'd capture all output to prevent it from going to the user. That said, you're right...startup log should be fine to be there. Will adjust.

client/src/connection/itc/CodeRunner.ts
```diff
@@ -0,0 +1,92 @@
...
+      session.onExecutionLogFn = addLine;
```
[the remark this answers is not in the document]
scottdover: Ah, yeah...what I saw of EG (from what @clangsmith showed me) is that we're executing the queries I have to pull in columns/libraries/tables (table rows being paginated is a different story, though). What I don't know is how EG maintains that line number 🤷 

I'm assuming the answer is that although we use those queries, we don't necessarily run things as proc sql and parse the results.

client/src/connection/index.ts
```diff
@@ -29,6 +30,7 @@ export type OnLogFn = (logs: LogLine[]) => void;
 export interface RunResult {
   html5?: string;
   title?: string;
+  logOutput?: string;
```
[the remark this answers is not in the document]
scottdover: Failed to clean this up. Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/3af76e75d6c6fab6104671d44f522330e3049db0

client/src/connection/itc/index.ts
scottdover: I don't remember adding trimEnd there. I've removed it in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/cfbd34ab90a037e8df31905535cfb89d8e0d886e and will make sure to re-test

package.json
scottdover: It looks like these are unneeded because determing whether or not things are displayed is handled in `views.sas-view-container`.

My guess is that the when clauses here can be used to further differentiate what view to show, but I don't think we need it.

package.json
scottdover: ssh would be an "invalid connection" (not the right wording) because it doesn't have access to sas content _or_ libraries/tables. Also, see the above :)

scottdover:
Hey @DmitryMK. From what I could tell, your issues came down to:
 - JSON parsing errors
 - JSON parsing errors specifically due to the max line length cutoff

I've made a couple adjustments in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/e30da68a262ad149842924e9e199f33adebbf29e.
 - I've updated the json output to be more compact
 - I've adjusted line length to allow for more characters (instead of the 32768 limit, it's not been adjusted to 100000)
 - I've added better error reporting for where things went awry

When you get a moment, will you retest with your first scenario and let me know if things have improved? Without the data, I'm not sure how we can effectively debug (I'm guessing there is a piece of data that is not working well with json parsing).

If you're able to share that data, or the full error message that shows up with that test case, that would be helpful. You are welcome to send that information to my email at scott.dover@sas.com if that helps :)

client/src/connection/itc/CodeRunner.ts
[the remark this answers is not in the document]
scottdover: Addressed in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/7a0e0ad0cbbff2e003227dc7ae40f01c0eb14fc2

client/src/connection/itc/index.ts
```diff
@@ -421,7 +411,12 @@ export class ITCSession extends Session {
...
+    const output = data.toString().trimEnd();
+    const sasSystemRegex = /1\s{4,}(.*)\s{4,}/;
```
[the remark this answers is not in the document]
scottdover: Brilliant idea! I updated things in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/7a0e0ad0cbbff2e003227dc7ae40f01c0eb14fc2 to use the carriage controls. This looks a _lot_ less fragile than what I had before :)

client/src/connection/itc/CodeRunner.ts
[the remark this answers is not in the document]
scottdover: This feels like something we can iterate on in a future PR. Thoughts?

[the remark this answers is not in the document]
scnwwu: Defer to @clangsmith

client/src/connection/itc/CodeRunner.ts
```diff
@@ -0,0 +1,80 @@
...
+  ): Promise<string> {
+    // If we're already executing code, lets wait for it
+    // to finish up.
+    if (useRunStore.getState().isExecutingCode) {
+      await new Promise((resolve) => {
+        if (this.executionInterval) {
+          clearInterval(this.executionInterval);
+        }
+
+        this.executionInterval = setInterval(() => {
```
scnwwu, line 25: oh just noticed this, how about subscribe the state change event, like
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/logViewer/index.ts#L91

client/src/connection/itc/index.ts
```diff
@@ -515,6 +508,8 @@ export class ITCSession extends Session {
    */
   private fetchResults = async () => {
     if (!this._html5FileName) {
+      this._pollingForLogResults = false;
+      this._skipPageHeaders = false;
```
scnwwu, line 512: nit: it feels a little risky to set to false here. For example some exception before entering here. Or some other route may be added in fututre.
How about set false from CodeRunner finally? Or, the `fetchLog` and `run` function accepts a `skipPageHeader` parameter, so that it only affect a specific run and not session level?

scottdover: > Or, the fetchLog and run function accepts a skipPageHeader parameter, so that it only affect a specific run and not session level?
I don't know why I didn't just do this. My thoughts were that I didn't want to muck with the session interface, but this seems like the most reasonable approach.

client/src/connection/itc/CodeRunner.ts
scottdover: That works! Updated in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/bfce3044abf0021f961b85cd12cc21406209e729

client/src/connection/itc/index.ts
scottdover: Updated in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/bfce3044abf0021f961b85cd12cc21406209e729

client/src/connection/session.ts
```diff
@@ -30,7 +30,7 @@ export abstract class Session {
   }
 
   protected abstract establishConnection(): Promise<void>;
-  abstract run(code: string): Promise<RunResult>;
+  abstract run(code: string, skipPageHeaders?: boolean): Promise<RunResult>;
```
scnwwu, line 33: I'm not sure do we really want to change the base signature? `skipPageHeaders` isn't implemented in other connection methods.
how about just `await (session as ITCSession).run(code, true);` in CodeRunner?

client/src/connection/itc/CodeRunner.ts
```diff
@@ -18,16 +15,10 @@ class CodeRunner {
...
-        this.executionInterval = setInterval(() => {
-          if (!useRunStore.getState().isExecutingCode) {
-            clearInterval(this.executionInterval);
-            return resolve(true);
-          }
-        }, 200);
+        useRunStore.subscribe(
+          (state) => state.isExecutingCode,
+          (isExecutingCode) => !isExecutingCode && resolve(true),
+        );
```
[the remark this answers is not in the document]
scottdover: Done in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/c9ca53cf966b6544f4769897964db5d4bf12fe76

client/src/connection/session.ts
scottdover: Done in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/c9ca53cf966b6544f4769897964db5d4bf12fe76

scnwwu (approved):
Good to me

ShermanWang0324:
Testing results: 
 Tested new functionality by starting with an empty set of profiles - PASS
 Tested viewing/refreshing libraries - PASS
 Tested viewing/refreshing tables -PASS
 Tested viewing table data (column types, row data, etc) - PASS
 Tested downloading data as a CSV - PASS
 Tested deleting tables -PASS
 Make sure log works as expected for regular sas program - PASS
Both IOM and COM connections are tested.

ShermanWang0324 (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 
 - Update command palette entry `SAS: Sign in` so that it's only visible for Viya profiles. Non-viya users will still be able to connect and execute code by using the running man icon ([#862](https://github.com/sassoftware/vscode-sas-extension/pull/862))
+- Add data viewer support for IOM/COM connections ([#680](https://github.com/sassoftware/vscode-sas-extension/issues/680))
```
scnwwu, line 18: please move it to [Unreleased]

scottdover: Done in https://github.com/sassoftware/vscode-sas-extension/pull/880/commits/ab08988169b1e2ec2ca33c650e5ea18072ead689

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Changed
```
scnwwu, line 9: nit: well, it would be better to be a `### Added` section in my opinion.
Please feel free to go ahead and merge it after resolved, thanks

scottdover: 🤦 resolved

Issue #878, opened 2024-03-06
Create files in SAS Content with proper content-type


brunomueller:
**Is your feature request related to a problem? Please describe.**
For a SAS Viya connection we can create files in SAS Content. However they are not created with the proper content-type.
Example:
Create a .html file in SAS Content works just fine, but will be created as 
{
  "creationTimeStamp": "2024-03-05T15:37:40.994Z",
  "modifiedTimeStamp": "2024-03-05T15:38:20.809Z",
  "createdBy": "...",
  "modifiedBy": "...",
  "id": "e55727e5-f7c5-40fc-b585-abe40bf34af9",
  "properties": {  },
  "contentDisposition": "filename*=UTF-8''simple.html",
  **"contentType": "text/plain",**
  "encoding": "UTF-8",

This means the file can not be used as a source for lets say a WebContent or DataDrivenObject in Visual Analytics. See this https://communities.sas.com/t5/SAS-Communities-Library/Deploy-DDC-Implementation-Files-in-SAS-Content-Server-via-SAS/ta-p/825559 for more details on storing .js, .css, .html in SAS Content

**Describe the solution you'd like**
Files create in SAS Content should have the proper content-type, so that they can be accessed using the SASJobExecution application using "/SASJobExecution/?_file=</full_path/file_name>" as described in the communities article above. Maybe using https://server/files/files/.../content will work as well.

**Additional context**
Since VSCode has good support to create .html, .js, .css files it would be great if files with these extensions are created appropriately in SAS Content.

**Environment**
Extension: 1.7.1, SAS Viya 2024.01

scottdover:
https://github.com/broofa/mime looks to have what we need to implement this. I'll take a look at this one soon :)

scottdover:
Thanks for bringing this one to our attention, @brunomueller. I believe we've found a solution in the linked pull request. This should be resolved in the 1.9.0. release :)

scottdover:
Interesting. Both things go through the same code path as far as I know. I'll take another look. Thanks for catching that :)

scottdover:
Hey @bruno-mueller-sas. I took a look into this. I'm wondering if I'm doing something differently than you. Here's what steps I'm taking:
 - Click the sas icon in the left activity bar, sign in
 - Create a new folder under my folder
 - Create a new file using the context menu (right click)
 - Name the file test.html
 - Get the file uri. Visit `<domain>/files/files/<uuid>`
[...]

PR #914, opened 2024-03-26
fix: create sas content files with proper content-type

scottdover:
**Summary**
This finds the content-type for files that are created or dragged into sas content

**Testing**
 - [x] Tested creating a new file in sas content
 - [x] Tested dragging a file into sas content
 - [x] Made sure renaming file didn't affect content-type


client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -8,6 +8,7 @@ import axios, {
   AxiosRequestConfig,
   AxiosResponse,
 } from "axios";
+import mime from "mime";
```
scottdover, line 11: This is giving me issues when running tests:

> > Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/vscode-sas-extension/vscode-sas-extension/client/node_modules/mime/dist/src/index.js from /home/runner/work/vscode-sas-extension/vscode-sas-extension/client/out/src/components/ContentNavigator/ContentModel.js not supported.

@smorrisj @scnwwu I know this has come up before. I can't recall if we had a solution/what it might be? 

As an alternative, we can always just use a subset of this list, or the list in its entirety - https://github.com/micnic/mime.json/blob/master/index.json

scottdover: it seems that vscode doesn't allow the consumption of esm modules, but that's on its way to being resolved - https://github.com/microsoft/vscode/issues/130367. 

Since mime types are pretty static, I just grabbed the json file above and used that (note: mime.json isn't in the npm registry) in https://github.com/sassoftware/vscode-sas-extension/pull/914/commits/03229bd86e4f471e8f5cf7f4da5b882a34e414ef

scnwwu (approved):

tianliang657:
Hi @scottdover, since I need to verify the content-type when creating a new .js/.css/.html files in sas content. but I cannot find the content-type. Do you know where I can find it? Like #878 's JSON information, or I can verify the content type in some other way. Thanks.

For example:
{
"creationTimeStamp": "2024-03-05T15:37:40.994Z",
"modifiedTimeStamp": "2024-03-05T15:38:20.809Z",
"createdBy": "...",
"modifiedBy": "...",
"id": "e55727e5-f7c5-40fc-b585-abe40bf34af9",
"properties": { },
"contentDisposition": "filename*=UTF-8''simple.html",
"contentType": "text/plain",
"encoding": "UTF-8",

tianliang657:
> Hey @tianliang657. Here's how I checked this:
> 
> * Uploaded a file with the extension
> * Opened chrome and visited `<domain>/folders/folders/@myFolder`
> * Clicked the `href` for the link identified by `"rel": "members"`
> * This takes me to a listing of all files in the folder
> * For each file, there is a link on the `uri` attribute
> * If you click the `uri` link, you'll be taken to more details about the file. That should include the `contentType`/`contentDisposition`/etc
> 
> Let me know if you need any further clarification, and thank you for testing :)

**[Qingbo]:** Thanks @scottdover for your help!

tianliang657:
This issue is fixed in the playpen package, no other issues were found, so approving.

 
image

image

image

PR #935, opened 2024-04-09
fix: rehydrate result panels on restarts

smorrisj:
**Summary**
Currently, Results from executed programs are not re-rendered if there are Result tabs open when vscode exits/restarts. On a restart, the editor group is left behind without any content. This leads to an undesirable user experience. The extension api supports [persisting](https://code.visualstudio.com/api/extension-guides/webview#persistence) state for webviews that can be leveraged in a such a way that we can rehydrate the result html for these panels in between vscode exits/restarts.

The following changes are made:

- Add deserialize function responsible for rehydrating panel state on restarts of vscode.
- Inject a small bit of javascript into the ODS results html to set the webviewpanel's unique ID using the vscode api's `setState` api. We need to call `setState` so that the deserialize logic knows which webview html to retrieve on rehydration. A unique id is needed due to the fact that we do not have access to vscode's internal representation of a unique identifier for webview panels. If a user elects to download the results html, this script is removed so that the ods html retains its fidelity.
- Extract html augmentation into its own function, since we need to ensure that the panel id stays up to date in the injected javascript either when a new panel is created or an existing one is updated with new html.
- Use a static name for ResultPanel view types. This is required so that the serialize/deserialize handlers will work correctly. A unique panel id is still calculated and embedded in the vscode object context and on the injected script for the above item.

**Testing**
See Issue details. Also re-verified that result files can be downloaded with using the new approach.


scottdover (commented):
Honestly, this isn't too dissimilar from what we do already. I think this is a fine approach

client/src/components/ResultPanel/ResultPanelSubscriptionProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+  webview: string;
+  resultPanelId: string;
+}
+export class ResultPanelSubscriptionProvider implements SubscriptionProvider {
+  getSubscriptions(): Disposable[] {
+    return [
+      commands.registerCommand(
+        "SAS.saveHTML",
+        async (context: ResultPanelContext) => {
+          const panel =
```
scottdover, line 20: Since we really only care about the panel html, how about making a method in `resultPanelManager` to give us back the html? That way, the logic removing the `<script>` tags can be encapsulated in the same spot where they're added

client/src/components/ResultPanel/ResultPanelManager.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    if (!singlePanel || !this._currentPanel) {
+      const resultPanel = window.createWebviewPanel(
+        SAS_RESULT_PANEL, // Identifies the type of the webview. Used internally
+        title, // Title of the panel displayed to the user
+        {
+          preserveFocus: true,
+          viewColumn: sideResult ? ViewColumn.Beside : ViewColumn.Active,
+        }, // Editor column to show the new webview panel in.
+        { enableScripts: true }, // Webview options. More on these later.
```
scottdover, line 99: Probably don't need the `More on these later.` here

smorrisj: Removed in https://github.com/sassoftware/vscode-sas-extension/pull/935/commits/fcfa9ddc0204a6fde9e630c22203f1e99f85e357.

client/src/components/ResultPanel/ResultPanelSubscriptionProvider.ts
smorrisj: Refactored in https://github.com/sassoftware/vscode-sas-extension/pull/935/commits/fcfa9ddc0204a6fde9e630c22203f1e99f85e357. Note that I dont strip the html on panel hydration, because I still want the setstate script in there so that we can support downloading of rehydrated webview panels.

smorrisj:
Thanks for the reviews. Since we already have a module that has access to the extension context, having a separate class for managing the result panels seemed redundant. I've refactored the code to get rid of this and to extend the ResultPanel module that we already have.

I also needed to move the html replacing further down so that the different display modes would work corectly (single panel true, false, etc).

client/src/components/ResultPanel/ResultPanel.ts
```diff
@@ -4,49 +4,38 @@ import { Uri, ViewColumn, WebviewPanel, l10n, window } from "vscode";
...
 import { v4 } from "uuid";
 
+import { getContextValue, setContextValue } from "../ExtensionContext";
 import { isSideResultEnabled, isSinglePanelEnabled } from "../utils/settings";
 
+const SCRIPT_REGEX = /<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi;
+export const SAS_RESULT_PANEL = "SASResultPanel";
+
 let resultPanel: WebviewPanel | undefined;
 export const resultPanels: Record<string, WebviewPanel> = {};
```
scnwwu, line 14: There's a problem. Steps to reproduce:

1. Set `"SAS.results.singlePanel": true` in settings
2. Run code to generate a result
3. Run another code to generate another result. It replaces the previous result content.
4. Right click the result content and select "Download" from the context menu.
5. Nothing happens, which is unexpected.

The reason is that the new `resultPanelId` is not recorded in `resultPanels`. Actually, I think now this `resultPanels` can be removed. The html can be always retrieved from context.

client/src/components/ResultPanel/ResultPanel.ts
```diff
@@ -62,5 +51,56 @@ export const showResult = (html: string, uri?: Uri, title?: string) => {
...
+          if(acquireVsCodeApi){
+            const vscode = acquireVsCodeApi();
+            const panelId = '${resultPanelId}'
+            vscode.setState({panelId});
+          }
+         </script>
+         <style>html,body { height: 100% !important; }</style></head>`,
+    );
   resultPanel.webview.html = html;
+  setContextValue(resultPanelId, html);
```
scnwwu, line 78: There's a problem. When "SAS.results.singlePanel" set to `true`.
First time say you have `resultPanelId === '001'`, it will create a result panel with
`resultPanel.onDidDispose(() => disposePanel('001'));`
Next time you have `resultPanelId === '002'`. So in your context, you have
```
{
    '001': 'html content 1',
    '002': 'html content 2',
}
```
But when the panel closes, it will only call `disposePanel('001')`. Then you have stale content left in the context.
I would suggest the `resultPanelId` should really match the "panel" instead of "content". When a panel is re-used, we should use the original panel id instead of generating a new id.

client/src/components/ResultPanel/ResultPanel.ts
```diff
@@ -4,63 +4,116 @@ import { Uri, ViewColumn, WebviewPanel, l10n, window } from "vscode";
...
     );
   }
-  resultPanel.webview.html = html;
+
+  const panelHtml = wrapPanelHtml(html, panelId);
+  resultPanel.webviewPanel.webview.html = panelHtml;
+  setContextValue(resultPanel.panelId, panelHtml);
+};
+
+/**
```
scnwwu, line 67: Would you either complete the comment or remove the comment?

client/src/components/ResultPanel/ResultPanel.ts
```diff
@@ -4,63 +4,116 @@ import { Uri, ViewColumn, WebviewPanel, l10n, window } from "vscode";
...
+export const fetchHtmlFor = async (panelId: string) => {
+  let panelHtml: string = "";
+  panelHtml = await getContextValue(panelId);
+  panelHtml = panelHtml.replace(SCRIPT_REGEX, "");
+
+  return panelHtml;
+};
+
+const disposePanel = (id: string) => {
+  console.log(`disposing panel ${id}`);
```
scnwwu, line 116: nit: is it intended to go in production?

scnwwu (commented):
Hi, it would be great if you can update the PR description to reflect the latest change about `ResultPanelManager`, and update the Testing section for better future reference.

scottdover (commented):
This looks okay to me, as long as @scnwwu's comments are addressed

client/src/components/ResultPanel/ResultPanel.ts
smorrisj: Removed in https://github.com/sassoftware/vscode-sas-extension/pull/935/commits/1708088bbdaee17dfac035c7896e2b7f1fdaf35a

client/src/components/ResultPanel/ResultPanel.ts
smorrisj: Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/935/commits/1708088bbdaee17dfac035c7896e2b7f1fdaf35a

scnwwu (approved):

tianliang657:
This issue is fixed in the playpen package, no other issues were found, see the attached video "Fixed.gif" for details, so approving.

PR #938, opened 2024-04-12
chore: using pretty for json output to avoid long line

ShuguangSun:
**Summary**
Now in the library for the ITC session, json is print to log in a long line which might exceed the limit of SAS if the data set is too large. Now we add the `pretty` option to `proc json` which makes each line of the json file short and it help SAS print full json file to log.

**Testing**
Prepare a large dataset, and open it in library.


scottdover (approved):

ShermanWang0324:
Tested with all types of connections, open large dataset in library is fine.

ShermanWang0324 (approved):

PR #965, opened 2024-04-28
fix: Fileref OUTFILE should be allowed for COM/IOM

### LINKED ISSUE #868: Fileref OUTFILE should be allowed for COM/IOM
### Discussed in https://github.com/sassoftware/vscode-sas-extension/discussions/858

Run below code against COM/IOM, error occurs.
```
filename outfile '/path';
proc print data=sashelp.class;
run;
```

scnwwu:
**Summary**
Fix #868
Pass an empty string to the `AssignFileref` call and it will create a temporary fileref that will never conflict with user code.

**Testing**
Test case in #868

### DIFF
--- a/client/src/connection/itc/script.ts
+++ b/client/src/connection/itc/script.ts
@@ -152,7 +152,7 @@ class SASRunner{
 
   [void]FetchResultsFile([string]$filePath, [string]$outputFile) {
     $fileRef = ""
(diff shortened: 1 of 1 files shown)


scottdover (approved):
Nice!

ShermanWang0324 (approved):
Testing completed, run code on both COM/IOM connections, no error anymore.

PR #978, opened 2024-05-09
fix: IOM connection hang when work dir is long

### LINKED ISSUE #964: Unable to complete IOM connection
**Describe the bug**
I cannot complete connection to SAS 9.4 on Unix from VSCODE on PC.  It was working before (2 to 3 weeks ago) and now it connects and I can see start of SAS log in VSCODE. but then it just waits (for what?) and never seems to actually submit my SAS code.  I can submit to "local" connection to Windows SAS 9.4 session.  Instead it just has message window saying connection to SAS session.

**Steps to reproduce**
Open or create SAS program.  For my example I just use "proc setinit;run;" as the code. Select SAS connection defined using IOM connection.  Submit code to run.

**Expected behavior**
SAS log should show results of running the code.

**Screenshots**

VSCODE Version: 1.88.1 (user setup)
Commit: e170252f762678dec6ca2cc69aba1570769a5d39
Date: 2024-04-10T17:41:02.734Z
Electron: 28.2.8
ElectronBuildId: 27744544
Chromium: 120.0.6099.291
Node.js: 18.18.2
V8: 12.0.267.19-electron.0
OS: Windows_NT x64 10.0.19045

SAS version: 9.4 (TS1M7 MBCS3170)
Host version: Linux LIN X64 3.10.0-1160.105.1.el7.x86_64 #1 SMP Mon Nov 6 06:58:51 EST 2023 x86_64 Red Hat Enterprise Linux Server release 7.9 
      (Maipo) 

**Additional context**
Worked before.  Works when connecting to LOCAL.  Now cannot get past the line in the SAS log of the code generated by the VSCode extension:  

%let workDir = %sysfunc(pathname(work));

scnwwu:
**Summary**
Fix #964
When the default working directory name is long, it may wrap into multiple lines in SAS log. The `WORK_DIR_END_TAG` may be break into 2 lines thus not captured.
We're injecting log type between each log line, so when parsing across multiple log lines, it has to process LogLineType first, otherwise the log line type will be captured as part of working directory name, which is wrong.

**Testing**
Make default working directory be a pretty long name and try run code.

### DIFF
--- a/client/src/connection/itc/index.ts
+++ b/client/src/connection/itc/index.ts
@@ -345,7 +345,9 @@ export class ITCSession extends Session {
   private fetchWorkDirectory = (line: string): string | undefined => {
     let foundWorkDirectory = "";
     if (
-      !line.includes(`%put ${WORK_DIR_START_TAG}&workDir${WORK_DIR_END_TAG};`)
+      !line.includes(`%put ${WORK_DIR_START_TAG};`) &&
+      !line.includes(`%put &workDir;`) &&
+      !line.includes(`%put ${WORK_DIR_END_TAG};`)
     ) {
       foundWorkDirectory = this._workDirectoryParser.processLine(line);
     } else {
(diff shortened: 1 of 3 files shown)


smorrisj (approved):
LGTM

scottdover (approved):

ShermanWang0324 (approved):
Running code with report's connection information, the code can be executed successful and the 'path' value in log is correct and no truncation.

sasutils:
I see issue 964 was merged into this.  It does make some sense that problems with long log lines might be what is preventing my previously working connections from failing.  The only system change was adding SAS monitoring. Which resulted in this wrapped line that I can see in the SAS log right before where it appears to hang.

NOTE: The autoexec file, ('/Volumes/app/sas/clinsas/esm/esm_8286/esm-agent/sasautos/esmtags.sas'
      '/Volumes/app/sas/prod3/grid_control_11374/Lev1/SASAppAWS/WorkspaceServer/autoexec.sas'), was executed at server
      initialization.

jinmei98:
how to change different workspace server like from SASAppAWS to SASAppTest

PR #980, opened 2024-05-10
fix: fix content type when saving files

scottdover:
**Summary**
Previously, saving files would always convert the content type to text/plain. This changes things such that we maintain the content type on save.

Additionally, this makes a few small refactoring changes:
 - This moves a couple of helper functions from ContentModel to utils
 - It removes some unnecessary writes to file metadata

**Testing**
 - [x] Test creating file
 - [x] Test saving file
 - [x] Test renaming file

### DIFF
--- a/client/src/components/ContentNavigator/ContentModel.ts
+++ b/client/src/components/ContentNavigator/ContentModel.ts
@@ -15,19 +15,19 @@ import {
 } from "../../connection/studio";
 import { SASAuthProvider } from "../AuthProvider";
 import {
-  DATAFLOW_TYPE,
-  FAVORITES_FOLDER_TYPE,
+  DEFAULT_FILE_CONTENT_TYPE,
   FILE_TYPES,
-  FOLDER_TYPE,
   FOLDER_TYPES,
   Messages,
   ROOT_FOLDER,
   TRASH_FOLDER_TYPE,
 } from "./const";
-import mimeTypes from "./mime-types";
-import { ContentItem, Link, Permission } from "./types";
+import { ContentItem, Link } from "./types";
 import {
+  fetchFileContentType,
+  fetchItemContentType,
   getLink,
+  getPermission,
   getResourceId,
   getResourceIdFromItem,
   getTypeName,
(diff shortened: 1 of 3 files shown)


client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -362,20 +336,21 @@ export class ContentModel {
...
   } {
-    if (resourceId in this.fileTokenMaps) {
-      return this.fileTokenMaps[resourceId];
+    if (resourceId in this.fileMetadataMap) {
+      return this.fileMetadataMap[resourceId];
     }
     const now = new Date();
     const timestamp = now.toUTCString();
-    return { etag: "", lastModified: timestamp };
+    return { etag: "", lastModified: timestamp, contentType: "text/plain" };
```
scnwwu, line 346: nit: DEFAULT_FILE_CONTENT_TYPE

client/src/components/ContentNavigator/ContentModel.ts
```diff
@@ -244,16 +227,13 @@ export class ContentModel {
   ): Promise<ContentItem | undefined> {
     const contentType = await this.getFileContentType(fileName);
```
scnwwu, line 228: nit: it looks a little confusing with `fetchFileContentType`. It's hard to tell the difference from the name "fetch" vs. "get". The implementation look unrelated to each other.

scottdover: I updated the method to better match what is being returned, and updating both "fetch" methods to use a prefix of "get" since that seems to be more consistent with what we use elsewhere https://github.com/sassoftware/vscode-sas-extension/pull/980/commits/1ee77cc65681082fc3b202e55e265821d23a11db

client/src/components/ContentNavigator/ContentModel.ts
scottdover: Done in https://github.com/sassoftware/vscode-sas-extension/pull/980/commits/1ee77cc65681082fc3b202e55e265821d23a11db

scnwwu (approved):

ShermanWang0324 (approved):

PR #991, opened 2024-05-19
feat: python language features

jingjiajie:
New libs are installed, please run **_npm install_**

**Summary**
"Pytight" language server is integrated, which is the Microsoft's official python language server.

Thanks to the Pyright, the following features **are supported**:
_Code completion
Hover hint
Outline for python code_
and functions shown in the picture:
a

Due to reasons, the following features are **not yet included**:
**Python code formatting:** It's not included in the Pyright language server. It has to be a feature we separately implement in the future.
**Folding range for Python code:** Because we set to use the Language Server side Folding range parsing rather than using the configuration file, it only recognizes the Folding ranges we returned from language server. However, the Python folding range is naturally implemented with configuration file, so it's ignored. We have to treat it as a separate feature and manually implement it with our language server in the future. 

a simple test code:
```
proc python;
submit;
        class Jonas:
            def name(self) -> str:
                return "Jonas"

        def salute() -> None:
            jonas = Jonas()
            print("This man is %s!" % jonas.name())
endsubmit;
run;

proc python;
    submit;
salute()
```


jingjiajie:
@scnwwu  In **main** branch, _formatting_ the above test code, the code disappearing problem occurs again just like before. Would you mind to have a check again? :)

Zhirong2022:
> @scnwwu In **main** branch, _formatting_ the above test code, the code disappearing problem occurs again just like before. Would you mind to have a check again? :)

@jingjiajie,  I have raised the issue https://github.com/sassoftware/vscode-sas-extension/issues/992 to track the changes. Thanks.

scnwwu:
@jingjiajie, I failed to build your PR code on my PC.
ENOENT: no such file or directory
vscode-sas-extension\server\node_modules\pyright-internal-lsp\dist\packages\pyright-internal\typeshed-fallback
I did `npm install` but still no such directory

jingjiajie:
> @jingjiajie, I failed to build your PR code on my PC. ENOENT: no such file or directory vscode-sas-extension\server\node_modules\pyright-internal-lsp\dist\packages\pyright-internal\typeshed-fallback I did `npm install` but still no such directory

Fixed

scnwwu:
There's a UX problem with this PR. The context menu item "Go to Definition", "Go to Declaration" etc. now always display. For most SAS code except PROC PYTHON, they have no functionality. It may be confusing for some user who rarely use Python.

scnwwu:
Got unexpected `"p" is not defined` error with below code in the editor
```
proc python terminate;
submit;
#Reference to variable defined in previous PROC PYTHON call
print("x = " + x)
def my_function():
    print("Inside the proc step")
endsubmit;
run;
```

scnwwu:
It does NOT work in browser build.

jingjiajie:
> There's a UX problem with this PR. The context menu item "Go to Definition", "Go to Declaration" etc. now always display. For most SAS code except PROC PYTHON, they have no functionality. It may be confusing for some user who rarely use Python.

@scnwwu  It's controlled by the InitializeResult returned by onInitialize() at server side. I'm not sure if there's a way to dynamically control this. Could have a future investigation. Please let me know if you have more ideas about this design.



> It does NOT work in browser build.

I found Microsoft maintained two language servers for Python, one is for desktop(Pyright https://github.com/microsoft/pyright), another is for browser(https://github.com/microsoft/vscode-python). In this PR we integrated the Pyright, so the browser version will become another work after this. 
Currently I haven't had a future look at the differecies between them, but it seems that the file system is the biggest difference. Desktop version reads local python executables, libs, etc. while browser version should have totally different mechanisms. Multi-threading is another big difference. There are also many node features required by desktop version for specific features, which browser version implements differently. So that's why there have to be two version.

scnwwu:
> > There's a UX problem with this PR. The context menu item "Go to Definition", "Go to Declaration" etc. now always display. For most SAS code except PROC PYTHON, they have no functionality. It may be confusing for some user who rarely use Python.
> 
> @scnwwu It's controlled by the InitializeResult returned by onInitialize() at server side. I'm not sure if there's a way to dynamically control this. Could have a future investigation. Please let me know if you have more ideas about this design.

Would you take a look at https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#client_registerCapability
So only register `textDocument/definition` when found proc python. I don't know if it will work.

jingjiajie:
> > > There's a UX problem with this PR. The context menu item "Go to Definition", "Go to Declaration" etc. now always display. For most SAS code except PROC PYTHON, they have no functionality. It may be confusing for some user who rarely use Python.
> > 
> > 
> > @scnwwu It's controlled by the InitializeResult returned by onInitialize() at server side. I'm not sure if there's a way to dynamically control this. Could have a future investigation. Please let me know if you have more ideas about this design.
> 
> Would you take a look at https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#client_registerCapability So only register `textDocument/definition` when found proc python. I don't know if it will work.

Good. It seems to work!

jingjiajie:
> Got unexpected `"p" is not defined` error with below code in the editor
> 
> ```
> proc python terminate;
> submit;
> #Reference to variable defined in previous PROC PYTHON call
> print("x = " + x)
> def my_function():
>     print("Inside the proc step")
> endsubmit;
> run;
> ```
> 
> Got unexpected `String literal is unterminated` error with below code in the editor
> 
> ```
> proc python;
>   interactive;
>   x = '''
>     endinteractive
>   '''
>   print('hello')
> endinteractive;
> run;
> ```

@scnwwu I found they are caused by Lexer.startFrom(). The CodeZoneManager reads token by calling the Lexer.startFrom() in the middle of embedded code. Without having read "proc python" in previous lines, the Lexer can't regard the token as a part of embedded code, instead it regards it as a regular proc stmt, then gives wrong result.

For example, the correct token should be 'print("x = " + x)' because it's in a proc python block, but if use Lexer.startFrom() from the start position of this line, the token returned is "print", that's wrong.

Is it possible to remove all the occurance of Lexer.startFrom() function, instead use SyntaxProvider.getSyntax()? It returns correct tokens. However it seems to be a big complexity, I'll need your confirmation.

scnwwu:
> > Got unexpected `"p" is not defined` error with below code in the editor
> > ```
> > proc python terminate;
> > submit;
> > #Reference to variable defined in previous PROC PYTHON call
> > print("x = " + x)
> > def my_function():
> >     print("Inside the proc step")
> > endsubmit;
> > run;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Got unexpected `String literal is unterminated` error with below code in the editor
> > ```
> > proc python;
> >   interactive;
> >   x = '''
> >     endinteractive
> >   '''
> >   print('hello')
> > endinteractive;
> > run;
> > ```
> 
> @scnwwu I found they are caused by Lexer.startFrom(). The CodeZoneManager reads token by calling the Lexer.startFrom() in the middle of embedded code. Without having read "proc python" in previous lines, the Lexer can't regard the token as a part of embedded code, instead it regards it as a regular proc stmt, then gives wrong result.
> 
> For example, the correct token should be 'print("x = " + x)' because it's in a proc python block, but if use Lexer.startFrom() from the start position of this line, the token returned is "print", that's wrong.
> 
> Is it possible to remove all the occurance of Lexer.startFrom() function, instead use SyntaxProvider.getSyntax()? It returns correct tokens. However it seems to be a big complexity, I'll need your confirmation.

@jingjiajie, please reset context in `Lexer.startFrom()`
Add `this.context.embeddedLangState = EmbeddedLangState.NONE;` under
https://github.com/sassoftware/vscode-sas-extension/blob/main/server/src/sas/Lexer.ts#L1027
Then the code zone looks correct. And the 1st snippet looks good.
I don't know what's the problem for the 2nd snippet. Would you provide more details?

server/src/sas/CodeZoneManager.ts
```diff
@@ -2457,8 +2461,11 @@ export class CodeZoneManager {
     }
   }
   private _currentZone(line: number, col: number) {
-    const newToken = this._token(line, col)!,
-      type = newToken.type; //self.type(line,col),
+    const newToken = this._token(line, col)!;
+    const type = newToken.type;
+    if (type === "embedded-code") {
```
jingjiajie, line 2466: @scnwwu 
I found a simple solution. I add this check here, so that if it's embedded, stop here, and not reach the later startFrom() logic.
Actually the line 2464: _"const newToken = this._token(line, col)!;"_ also calls the SyntaxProvider to get the correct "embedded-code" token. If any concern, please raise a comment.

jingjiajie:
All reported issues so far have been fixed.

scnwwu:
@jingjiajie, the build is hang. Would you please address it?

server/src/python/browser/PyrightLanguageProviderBrowser.ts
```diff
@@ -0,0 +1,268 @@
...
+import { TextDocument } from "vscode-languageserver-textdocument";
+
+import { IPythonMode } from "pyright-internal-browser/dist/packages/pyright-internal/src/analyzer/sourceFile";
+import { FileSystem } from "pyright-internal-browser/dist/packages/pyright-internal/src/common/fileSystem";
+import { PyrightServer } from "pyright-internal-browser/dist/packages/pyright-internal/src/server";
+
+import { LanguageServiceProvider } from "../../sas/LanguageServiceProvider";
+import { extractPythonCodes } from "../utils";
+
+export class PyrightLanguageProviderBrowser extends PyrightServer {
```
scnwwu, line 52: looks like the majority part of code in PyrightLanguageProviderBrowser is same to PyrightLanguageProviderNode
any chance to make a platform-independent PyrightLanguageProvider class (instead of only an interface)?

client/src/commands/run.ts
```diff
@@ -49,8 +49,7 @@ async function getSelectedRegions(
       "sas/getFoldingBlock",
       {
         textDocument: { uri: window.activeTextEditor.document.uri.toString() },
-        line,
-        col,
+        position: { line, col },
```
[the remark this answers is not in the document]
jingjiajie: I changed this because all the onXXX() events at server side have a structure like param{position{line: xx, col: xx}}, so I encapsulated a common function "dispatch" which resolves that structure and handle common logics. Changing this is just to make it the same like all the other interfaces, and adapt the dispatch() function. Considering this is a sas only call, this shouldn't have other side effects that I can know.

It's a bug at the line 228-229, I'll fix it. For any other concerns pls let me know.

tools/build.mjs
```diff
@@ -1,10 +1,28 @@
+import esbuildCopyPlugin from "@sprout2000/esbuild-copy-plugin";
+
 import concurrently from "concurrently";
 import esbuild from "esbuild";
 
 console.log("start");
 const dev = process.argv[2];
 
 const plugins = [
```
scnwwu, line 9: this plugin is only needed for `nodeBuildOptions`, not `browserBuildOptions`. The 2 run concurrently (see line 88 below) thus will produce errors.
Actually you can simply do `fs.copyFile` with native Nodejs after build done (for example at line 100 below)

server/src/server.ts
```diff
@@ -1,172 +1,628 @@
...
+  connection.onDidCloseTextDocument(async (params) => {
+    const uri = params.textDocument.uri;
+    delete documentPool[uri];
+    await _pyrightLanguageProvider.onDidCloseTextDocument(params);
+  });
+
+  connection.onDidChangeTextDocument((params) => {
+    const uri = params.textDocument.uri;
+    const docInfo = documentPool[uri];
+    TextDocument.update(
```
scnwwu, line 305: why update document by yourself? why not use the `documents.onDidChangeContent` event as before?

jingjiajie: Because the PyrightServer needs onDidOpenTextDocument() and onDidCloseTextDocument() to be handled.
If use documents.onDidChangeContent() as before, it requires to also use documents.onDidOpen(), documents.onDidClose() to handle open/close event, otherwise it doesn't work. However, it's parameters are different from what Pyright accepts, so can't easily pass it to Pyright.

server/src/python/PyrightLanguageProvider.ts
```diff
@@ -0,0 +1,162 @@
...
+import type { ReadOnlyFileSystem } from "pyright-internal-node/dist/packages/pyright-internal/src/common/fileSystem";
+import type { ClientCapabilities } from "pyright-internal-node/dist/packages/pyright-internal/src/common/languageServerInterface";
+import type { DocumentRange } from "pyright-internal-node/dist/packages/pyright-internal/src/common/textRange";
+import type { Uri } from "pyright-internal-node/dist/packages/pyright-internal/src/common/uri/uri";
+import type { CollectionResult } from "pyright-internal-node/dist/packages/pyright-internal/src/languageService/documentSymbolCollector";
+import type { ParseFileResults } from "pyright-internal-node/dist/packages/pyright-internal/src/parser/parser";
+
+import { LanguageServiceProvider } from "../sas/LanguageServiceProvider";
+
+export interface PyrightLanguageProvider {
```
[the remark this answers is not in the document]
jingjiajie: Cool, it seems to work well. I learned a good paradigm in TypeScript👍

server/src/sas/CodeZoneManager.ts
```diff
@@ -256,7 +256,7 @@ export class CodeZoneManager {
       type: Token["type"] = "text",
       currLine = line;
     for (; i < len; i++) {
-      if (syntax[i].start >= col) {
+      if (syntax[i].start > col) {
```
scnwwu, line 259: @jingjiajie, would you please provide a detailed sample about why need to change this?
As it is fundamental logic, I want to verify it carefully. Thank you.

server/src/sas/CodeZoneManager.ts
```diff
@@ -394,7 +394,7 @@ export class CodeZoneManager {
     } while (/^\s*$/.test(text));
 
     if (Lexer.isComment[type] || Lexer.isLiteral[type]) {
-      token = this._token(context.line, col + 1)!;
+      token = this._token(context.line, col - 1)!;
```
scnwwu, line 397: @jingjiajie, would you please provide a detailed sample about why need to change this?
As it is fundamental logic, I want to verify it carefully. Thank you.

client/package-lock.json
```diff
@@ -895,39 +901,43 @@
...
-        "minimatch": "^5.1.0",
-        "semver": "^7.3.7",
-        "vscode-languageserver-protocol": "3.17.5"
+        "minimatch": "^9.0.3",
+        "semver": "^7.6.0",
+        "vscode-languageserver-protocol": "3.17.6-next.6"
       },
       "engines": {
-        "vscode": "^1.82.0"
+        "vscode": "^1.89.0"
```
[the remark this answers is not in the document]
scnwwu: A PR can add `[Unreleased]` entry, like
https://github.com/sassoftware/vscode-sas-extension/pull/552/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR7
It matches our standard https://keepachangelog.com/en/1.0.0/
If multiple PRs going at same time, it may cause conflicts. So it's also OK to not change CHANGELOG in PR. I'll finalize it at release. I just put a comment here for awareness. Thanks.

server/src/sas/CodeZoneManager.ts
jingjiajie: Here is an example. The "e" of "endinteractive;" is misidentified if not change the logic of line 259, 397. Due to code zone is misidentified.

```
proc python;
  interactive;
  x = '''
    endinteractive
  '''
  print('hello') 
endinteractive;
run;
```

a

client/package-lock.json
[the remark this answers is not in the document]
jingjiajie: OK, I added an [unreleased] entry.

server/src/server.ts
```diff
@@ -1,172 +1,631 @@
...
 
-  connection.onCompletion((params) => {
-    const languageService = getLanguageService(params.textDocument.uri);
-    completionProvider = languageService.completionProvider;
-
-    return completionProvider.getCompleteItems(params.position);
+  connection.onCompletion(async (params, token) => {
+    return await dispatch(params, {
+      async sas(languageService) {
+        if (params.context?.triggerCharacter !== " ") {
```
scnwwu, line 135: There's a situation that the `triggerCharacter` is `undefined` (triggered by typing matched keyword), in that case we should allow autocomplete.
```suggestion
        if (params.context?.triggerCharacter && params.context.triggerCharacter !== " ") {
```

scnwwu (approved):

Zhirong2022:
Put mouse indicator in the python code, try to 'Run Region' from the context menu, it will run proc python code block and others together.
```
proc print data=sashelp.cars(obs=1);
run;

proc python;
submit;
import random

if __name__ == '__main__':
    # print
    print('Hello World!')

    # 4 random number
    i = 0
    rand_list = []
    while i < 4:
        rand_num = random.randint(0, 499)
        if rand_num not in rand_list:
            rand_list.append(rand_num)
            i += 1
    # ouput
    print(rand_list)
endsubmit;
run;

```

Zhirong2022:
Python language features are disabled if the 'proc python' code block locates in a customized code block.
```
*region;
proc print data=sashelp.cars;
run;

proc python;
submit;
my_function()
print("y = " + y)
endsubmit;
run;
*endregion;
```

Zhirong2022:
Error on 'run;' unexpectedly if only having comments in python code
```
proc python;
    interactive;
    #comment
    endinteractive;
run;
```
image

Zhirong2022:
Code highlighting behave improperly if having some indentation issue in python code block.
```
proc python;
submit;
import requests
def get_poins(url):
request = requests.get(url)
poins = request.json()
return poins
endsubmit;
run;

proc print data=sashelp.class(obs=1);
run;

proc python;
submit;
    def f(x):
        x = str(x)
        mid = int(len(x) / 2)
        i = 0
        while i < mid:
            if x[i] != x[-(i + 1)]:
                return False
            i += 1
        return True
    endsubmit;
run;
```
image

jingjiajie:
> Code highlighting behave improperly if having some indentation issue in python code block.
> 
> ```
> proc python;
> submit;
> import requests
> def get_poins(url):
> request = requests.get(url)
> poins = request.json()
> return poins
> endsubmit;
> run;
> 
> proc print data=sashelp.class(obs=1);
> run;
> 
> proc python;
> submit;
>     def f(x):
>         x = str(x)
>         mid = int(len(x) / 2)
>         i = 0
>         while i < mid:
>             if x[i] != x[-(i + 1)]:
>                 return False
>             i += 1
>         return True
>     endsubmit;
> run;
> ```
> 
> image

@Zhirong2022 It's a hard problem. There isn't a simple way to disable this warning cross the proc python boundary. Since in Python, the global 'return' statement is an illegal statement, I suggest no fix. 

There is another case similar to this. If user write a global 'exit()', the same issue would happen. When this happen, we could suggest user refactor there code like:
```
def my_exit():
  exit()
my_exit()
```
So that the issue would be avoided.

tianliang657:
Hi @jingjiajie , I found an issue in the playpen package, first, I created 2 ".py" files (test111.py and test222.py) and a ".sas" file (test111.sas), the background is that the "test222.py" calls a method in "test111.py", and the "test111.sas" calls the same method in "test111.py", rename the method name of "test111.py" in "test111.sas", the new method name cannot be displayed in "test111.sas" after press Enter key, but the new method name can be displayed in "test111.py", because the "test222.py" also call the same method in "test111.py", but the new method name cannot be displayed in "test222.py". please see the attached video for the details.
Video

Zhirong2022:
Please check the code below, type 'submit' at the second 'proc python' code block, it will prompt the variable 'submit' information instead of the SAS Keyword 'submit' syntax help. 
```
proc python;
    interactive;
    print("Hello World")
    submit = "Hi!"
    endinteractive;
run;

proc python;
    
    
data test;
    x=6;
run;
```
image

tianliang657:
When a method is called behind the return keyword of another method in the ".sas" file, although the code is unreachable, but the "Go to Definition" feature should be supported, currently, the "no definition found for xxx.sas" is displayed after click on the "Go to Definition" menu item. please see the attached video for the details.
Video

image

jingjiajie:
> Hi @jingjiajie , I found an issue in the playpen package, first, I created 2 ".py" files (test111.py and test222.py) and a ".sas" file (test111.sas), the background is that the "test222.py" calls a method in "test111.py", and the "test111.sas" calls the same method in "test111.py", rename the method name of "test111.py" in "test111.sas", the new method name cannot be displayed in "test111.sas" after press Enter key, but the new method name can be displayed in "test111.py", because the "test222.py" also call the same method in "test111.py", but the new method name cannot be displayed in "test222.py". please see the attached video for the details. Video [ Video ](https://private-user-images.githubusercontent.com/37020387/358161473-8290edb7-b22d-4eed-8cbd-67085b92d9af.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjM3OTI1MjQsIm5iZiI6MTcyMzc5MjIyNCwicGF0aCI6Ii8zNzAyMDM4Ny8zNTgxNjE0NzMtODI5MGVkYjctYjIyZC00ZWVkLThjYmQtNjcwODViOTJkOWFmLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MTYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODE2VDA3MTAyNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWU0MDliMjA3ZDIwYTE0YTJmMTU2ZWVmZGU0NTU0NzhlMzZmZGM1OTk4MzYyZTRmZmM1NWQwN2YxNGEyODg5NWEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.Ip_b56oZ-cFa1yfx5DpThycedKAe5IgpEJe7mCuKU4k) [ ](https://private-user-images.githubusercontent.com/37020387/358161473-8290edb7-b22d-4eed-8cbd-67085b92d9af.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjM3OTI1MjQsIm5iZiI6MTcyMzc5MjIyNCwicGF0aCI6Ii8zNzAyMDM4Ny8zNTgxNjE0NzMtODI5MGVkYjctYjIyZC00ZWVkLThjYmQtNjcwODViOTJkOWFmLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MTYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODE2VDA3MTAyNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWU0MDliMjA3ZDIwYTE0YTJmMTU2ZWVmZGU0NTU0NzhlMzZmZGM1OTk4MzYyZTRmZmM1NWQwN2YxNGEyODg5NWEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.Ip_b56oZ-cFa1yfx5DpThycedKAe5IgpEJe7mCuKU4k)

@tianliang657 This seems to be an internal problem of Pyright. I posted a discussion for this so we can wait for their response: https://github.com/microsoft/pyright/discussions/8772

jingjiajie:
> When a method is called behind the return keyword of another method in the ".sas" file, although the code is unreachable, but the "Go to Definition" feature should be supported, currently, the "no definition found for xxx.sas" is displayed after click on the "Go to Definition" menu item. please see the attached video for the details. Video [ Video ](https://private-user-images.githubusercontent.com/37020387/358175513-8bd058eb-5294-49a9-a81c-e737b372067d.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjM3OTI1MjQsIm5iZiI6MTcyMzc5MjIyNCwicGF0aCI6Ii8zNzAyMDM4Ny8zNTgxNzU1MTMtOGJkMDU4ZWItNTI5NC00OWE5LWE4MWMtZTczN2IzNzIwNjdkLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MTYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODE2VDA3MTAyNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTQyMWFhOTVlM2UwNTAxMDY3Y2U2MTYwODA2NjI1YTE2NTZmNzNlYjE5MjY4ZjIxYTBiOTBmOTJhOTU4YTU4ZDgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.htwmt2GEaYhY7ix74j3-BTZRHVuwYtzZiRIDINab49I) [ ](https://private-user-images.githubusercontent.com/37020387/358175513-8bd058eb-5294-49a9-a81c-e737b372067d.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjM3OTI1MjQsIm5iZiI6MTcyMzc5MjIyNCwicGF0aCI6Ii8zNzAyMDM4Ny8zNTgxNzU1MTMtOGJkMDU4ZWItNTI5NC00OWE5LWE4MWMtZTczN2IzNzIwNjdkLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MTYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODE2VDA3MTAyNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTQyMWFhOTVlM2UwNTAxMDY3Y2U2MTYwODA2NjI1YTE2NTZmNzNlYjE5MjY4ZjIxYTBiOTBmOTJhOTU4YTU4ZDgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.htwmt2GEaYhY7ix74j3-BTZRHVuwYtzZiRIDINab49I)
> 
> image

I tried this in raw .py file, where no SAS involved. It behaves the same. So it's just Pyright's bug, out of our scope, I suggest nofix.

jingjiajie:
> Please check the code below, type 'submit' at the second 'proc python' code block, it will prompt the variable 'submit' information instead of the SAS Keyword 'submit' syntax help.
> 
> ```
> proc python;
>     interactive;
>     print("Hello World")
>     submit = "Hi!"
>     endinteractive;
> run;
> 
> proc python;
>     
>     
> data test;
>     x=6;
> run;
> ```
> 
> image

@Zhirong2022 fixed.

Zhirong2022:
The autocomplete with item 'endsubmit' will be shown if trying to specify a value to a variable.
image

Zhirong2022:
It has error validation for a valid code typing.
```
proc python;

    interactive;
    endinteractive;
run;
```
image

tianliang657:
> The autocomplete with item 'endsubmit' will be shown if trying to specify a value to a variable. image

I can met the same issue with the latest extension, but it cannot be reproduced in previous extensions, so it should be a regression issue.

tianliang657:
The autocomplete cannot appear until the 'endinteractive' keyword is fully entered when entering 'endinteractive' keyword. 
Video

tianliang657:
> It has error validation for a valid code typing.
> 
> ```
> proc python;
> 
>     interactive;
>     endinteractive;
> run;
> ```
> 
> image

If there is a blank line between the “proc python;” and “interactive”, the “interactive” keyword will be thrown error " 'interactive' is not defined", if the blank line does not exist, the error will not appear, this issue cannot be reproduced in previous extensions, so it should be a regression issue.
Video

Zhirong2022:
Proper syntax highlighting for the embedded python if using the alias name of  'interactive'
image

jingjiajie:
New reported bugs are all fixed. @Zhirong2022 @tianliang657

Zhirong2022:
There is no autocomplete if the first line is a comment. Try to type 'python' or 'print' next to 'proc'
```
/*  */
proc 
```
image

Zhirong2022:
There is no syntax help for 'endsubmit' and 'endinteractive'.
image

Zhirong2022:
Suppose the proc python code written on a line intentionally as below. Delete the code and retype properly, no syntax highlighting for the SAS keyword and no outline update.
```
proc python;submit;print('test') endsubmit; run;
```
image

Zhirong2022:
The typed python code will be removed in the code below after format code
```
proc python;
    i;
    print('test')
    endinteractive;
run;

```

jingjiajie:
> There is no autocomplete if the first line is a comment. Try to type 'python' or 'print' next to 'proc'
> 
> ```
> /*  */
> proc 
> ```
> 
> image

Fixed.

jingjiajie:
> Suppose the proc python code written on a line intentionally as below. Delete the code and retype properly, no syntax highlighting for the SAS keyword and no outline update.
> 
> ```
> proc python;submit;print('test') endsubmit; run;
> ```
> 
> image

Fixed.

scnwwu:
> > The typed python code will be removed in the code below after format code
> > ```
> > proc python;
> >     i;
> >     print('test')
> >     endinteractive;
> > run;
> > ```
> 
> @scnwwu I haven't read through the formatter codes. But just found a logic that is suspicious: image It seems that it discards any embedded codes. It could cause this bug.

@jingjiajie, the embedded code is handled separately. Would you please try add `i` beside `interactive`?
https://github.com/sassoftware/vscode-sas-extension/blob/main/server/src/sas/formatter/parser.ts#L131

server/src/sas/CodeZoneManager.ts
```diff
@@ -251,57 +252,56 @@ export class CodeZoneManager {
...
+    let foundSyntaxIdx = -1,
       type: Token["type"] = "text",
       currLine = line;
-    for (; i < len; i++) {
-      if (syntax[i].start > col) {
-        if (syntax[i - 1].start <= col) {
-          j = i - 1;
-          type = syntax[j].style;
+
+    for (let i = 0; i < syntax.length; i++) {
```
scnwwu, line 259: @jingjiajie, would you please provide a detailed explanation for this change? It looks like a fundamental logic. I'd like to verify it carefully. Thank you.

Zhirong2022:
No autocomplete if trying to type proc python or some others under comment line.
```
data _null_;
    set sashelp.cars;
run;

/*  */
```
image

Zhirong2022:
Error 'Request textDocument/foldingRange failed.' if trying to delete 'endsubmit' at line 6.
```
proc python terminate;
submit;
#Reference to variable defined in previous PROC PYTHON call
x = "Updating the value for x"
print("x = " + x)
endsubmit;

#Redefining a function called my_function
# def my_function():
#     print("Inside the proc step")
run;

proc python;
submit;
print("x = " + x)
endsubmit;
run;
```

Zhirong2022:
Some code will be always automatically inserted in the first python code block.  Type run to load the autocomplete, select 'runtime' from the list.
```
proc python;
    submit;
    print('test')
    endsubmit;
run;

proc python;
    interactive;
```
image

jingjiajie:
> > > The typed python code will be removed in the code below after format code
> > > ```
> > > proc python;
> > >     i;
> > >     print('test')
> > >     endinteractive;
> > > run;
> > > ```
> > 
> > 
> > @scnwwu I haven't read through the formatter codes. But just found a logic that is suspicious: image It seems that it discards any embedded codes. It could cause this bug.
> 
> @jingjiajie, the embedded code is handled separately. Would you please try add `i` beside `interactive`? https://github.com/sassoftware/vscode-sas-extension/blob/main/server/src/sas/formatter/parser.ts#L131

@scnwwu Sorry for missed message, I just see your reply. I tried replace i with interactive, then it worked. I found the root cause in the code, and added support for "i" keyword in addition to interactive; Thanks.

server/src/sas/CodeZoneManager.ts
jingjiajie: I remember this change is to fix the bug:

_There is no autocomplete if the first line is a comment. Try to type 'python' or 'print' next to 'proc'_

```
/*  */
proc 
```

Because if the starting one token of the whole document is a comment, __token()_ returned a null value. This change is to fix this bug.

The intention is only to fix that issue, and the remaining logics should keep the same behavior as before.

jingjiajie:
> No autocomplete if trying to type proc python or some others under comment line.
> 
> ```
> data _null_;
>     set sashelp.cars;
> run;
> 
> /*  */
> ```
> 
> image

Fixed

jingjiajie:
> Error 'Request textDocument/foldingRange failed.' if trying to delete 'endsubmit' at line 6.
> 
> ```
> proc python terminate;
> submit;
> #Reference to variable defined in previous PROC PYTHON call
> x = "Updating the value for x"
> print("x = " + x)
> endsubmit;
> 
> #Redefining a function called my_function
> # def my_function():
> #     print("Inside the proc step")
> run;
> 
> proc python;
> submit;
> print("x = " + x)
> endsubmit;
> run;
> ```

Fixed.

jingjiajie:
> Some code will be always automatically inserted in the first python code block. Type run to load the autocomplete, select 'runtime' from the list.
> 
> ```
> proc python;
>     submit;
>     print('test')
>     endsubmit;
> run;
> 
> proc python;
>     interactive;
> ```
> 
> image

@Zhirong2022 
This problem is hard to fix because this behavior provided by Pyright is not configurable for us.
Pyright's rule is: 
It always automatically insert _import_ clause at the first non-empty and non-comment python code line. In this case, the first python code line is _print('test')_, but actually we will insert a hidden _if True:_ line before this line, that is, the same line of _submit;_ clause. That's why you see it insert the _import_ clause at the line of _submit_ clause.

If you insert a empty line before _print('test')_ , it will insert it at the _print('test');_ line.
This is the only solution so far I could come up with.

Zhirong2022:
Suppose define a variable named 'endsubmit' or 'endinteractive', it will show SAS syntax help for the variable.
Type 'end' at the second python code block.
```
proc python;
submit;
endsubmit = 'hello'
endinteractive ='hi'
print('test')
endsubmit;
run;

proc python;
interactive;

endinteractive;
run;
```
image

Zhirong2022:
Create a new .sas file and type code below. The code has syntax error as expected since it has no colon next to submit. Perform 'Format document' through ContextMenu.

1. It will keep processing and no further response.
2. Stop the process and correct the code, there is no syntax highlighting anymore if trying to type something new.

```
proc python;
    submit
    endsubmit;
    run;
```

image

jingjiajie:
> Suppose define a variable named 'endsubmit' or 'endinteractive', it will show SAS syntax help for the variable. Type 'end' at the second python code block.
> 
> ```
> proc python;
> submit;
> endsubmit = 'hello'
> endinteractive ='hi'
> print('test')
> endsubmit;
> run;
> 
> proc python;
> interactive;
> 
> endinteractive;
> run;
> ```
> 
> image

**Fixed.**

jingjiajie:
> Create a new .sas file and type code below. The code has syntax error as expected since it has no colon next to submit. Perform 'Format document' through ContextMenu.
> 
> 1. It will keep processing and no further response.
> 2. Stop the process and correct the code, there is no syntax highlighting anymore if trying to type something new.
> 
> ```
> proc python;
>     submit
>     endsubmit;
>     run;
> ```
> 
> image

@Zhirong2022 **Fixed**.

Zhirong2022:
The syntax help and the keyword is not consistent, hover mouse over 'endinteractive'.
```
proc python;
interactive;

endinteractive;
run;
```
image

jingjiajie:
> The syntax help and the keyword is not consistent, hover mouse over 'endinteractive'.
> 
> ```
> proc python;
> interactive;
> 
> endinteractive;
> run;
> ```
> 
> image

@Zhirong2022 **Fixed.**

Zhirong2022:
All raised issues have fixed.

Zhirong2022 (approved):

PR #1006, opened 2024-05-23
Revert "fix: pin node version to avoid compatibility issues"

scnwwu:
Reverts sassoftware/vscode-sas-extension#786

Looks like the issue was resolved now


scnwwu:
Manually triggered it and it reported missing translations as expected.
https://github.com/sassoftware/vscode-sas-extension/actions/runs/9203806295

Issue #1017, opened 2024-05-29
Problems are not cleared on problem panel after correcting the code


Zhirong2022:
**Describe the bug**
Open VSC SAS Extension base on main branch. Type some code and run, problems will be shown on problem panel as expected. Correct the code and run again, problems are not cleared.

**Steps to reproduce**
1. Open VSC SAS Extension
2. Type code below
```
libname newname "c:\temp" access=readonly;

```
3. Run code, it has no error or warning. 
4. Type code below
```
libname longername "c:\temp" access=readonly;

```
4. Run code, it will show errors on problem panel as expected
5. Correct the code to 
```
libname newname "c:\temp" access=readonly;

```
6. Run the corrected code, errors are not cleared on problem panel

**Expected behavior**
Problems are cleared if having no errors or warning after correcting the code 

**Screenshots**
![image](https://github.com/sassoftware/vscode-sas-extension/assets/121005702/ba23d303-d4d7-44e5-aa3e-ce8c4a40f647)

**Environment (please complete the following information):**
Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.9.0]
SAS version: [e.g. Viya 2022.11]

PR #1018, opened 2024-05-29
fix: problems are not clear when there is no problem in log (#1017)

scnjdl:
**Summary**
The problem related with a sas file should be clear if there is no problem in its log.

**Testing**
see issue #1017 description.


scnwwu (approved):

Zhirong2022:
The issue has been fixed.  Error or warning is cleared on problem panel after correcting the code.

Zhirong2022 (approved):

PR #1060, opened 2024-06-20
fix: output log in running tasks

### LINKED ISSUE #1058: Since v1.10.0, Terminal -> Run Task -> 'sas: Run sas file' does not produce any log in the OUTPUT pane
**Describe the bug**
Since v1.10.0, submitting SAS code using the menu Terminal -> Run Task -> 'sas: Run sas file' does not produce any visible log in the OUTPUT pane anymore (observed using local profile).  This was working in the previous version.

Only when submitting with F3 or F8 do I see the log in the output pane.

**Expected behavior**
Log should be visible even when submitting with "Run Task"

**Screenshots**
NA

Client OS:  Windows 10
Extension version: v1.10.0
SAS version: SAS 9.4 TS1M8

scnjdl:
**Summary**
Log will be shown in output panel
Fix #1058 

**Testing**
Run task
see log is shown in output panel

### DIFF
--- a/client/src/commands/run.ts
+++ b/client/src/commands/run.ts
@@ -21,7 +21,7 @@ import { sasDiagnostic } from "../components/logViewer/sasDiagnostics";
 import { SASCodeDocument } from "../components/utils/SASCodeDocument";
 import { getCodeDocumentConstructionParameters } from "../components/utils/SASCodeDocumentHelper";
 import { isOutputHtmlEnabled } from "../components/utils/settings";
-import { ErrorRepresentation, RunResult, getSession } from "../connection";
+import { ErrorRepresentation, getSession } from "../connection";
 import { useRunStore } from "../store";
 import { profileConfig, switchProfile } from "./profile";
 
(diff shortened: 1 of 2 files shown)


client/src/commands/run.ts
```diff
@@ -204,14 +215,39 @@ export async function runTask(
...
   messageEmitter.fire(`${l10n.t("SAS code running...")}\r\n`);
-  return cancelled ? undefined : session.run(code);
+  return cancelled
+    ? undefined
+    : session.run(codeDoc.getWrappedCode()).then((results) => {
+        showRunResult(results, messageEmitter, taskLabel);
+      });
+}
+
+function showRunResult(
```
scnwwu, line 236: I would suggest to keep `showRunResult` in `SasTasks.ts` as before. It's task-specific.

scnwwu (approved):

Zhirong2022:
Log contents are shown on output panel as expected through 'Run task'.

Zhirong2022 (approved):

Issue #1059, opened 2024-06-20
Fallback auth when SSH_AUTH_SOCK unavailable


chunky:
## Is your feature request related to a problem? Please describe.

I'm in a Windows-heavy environment with Linux servers running 9.4, and IOM is unavailable. My userbase are highly skilled statisticians/economists/researchers, not really interested in messing around with environment variables or "arcane ssh stuff".

On Windows in particular, ssh-agent and setting environment variables are high-friction. I am reasonably expecting my users to have git [including git-bash and ssh/ssh-agent/etc] installed, but not any other SSH client. Some of them have PuTTY and PuTTY Keys configured, but that's much less common.

## Current (Disliked) Workflow

Currently the process looks like this:
1. One-time setup:
      1. Use the windows control panel for environment variables to set ```SSH_AUTH_SOCK``` to ```c:/Users/$USER/ssh.sock```
      2. Create a 9.4 SSH profile for SAS, in VS Code.
2. Every time they log in:
      1. Open git-bash
      2. Run ```ssh-agent -a /c/Users/$USER/ssh.sock```
      3. Run ```ssh-add```

Which is a little heavy on magick, highly error-prone, and a bunch more steps than most analysts enjoy repeating. One option is to try and have VS code execute the one-off stuff on startup using a task, but that seems brittle and hard to debug.

## Describe the solution you'd like

The ssh2 npm module that this plugin uses can take a "password" item in the configuration. The function getSession in client/src/connection/ssh/index.ts:24 could be edited to ask for a password, if the SSH_AUTH_SOCK environment variable isn't set. Notionally:

```typescript
export function getSession(c: Config): Session {
  if(!sessionInstance) {
      if (!process.env.SSH_AUTH_SOCK) {
        c.password = await createInputTextBox(ProfilePromptType.ClientSecret, c.password, true);
      }
      sessionInstance = new SSHSession();
      sessionInstance.config = c;
  }
  return sessionInstance;
}
```

Then, around line 71 in the same file, add ```password: this._config.password``` to the relevant configuration creation.

### Use npm ssh2's authhandler to ask for a password as a fallback

According to this documentation, you can stuff some additional items into the authHandler when connecting: https://www.npmjs.com/package/ssh2#client

The last example there provides a callback for use as a ```keyboard-interactive``` authhandler. 

[...]

Issue #182, opened 2023-03-20
Remote SAS 9.4 not usable with encrypted private keys.


cstotzer:
I'm currently testing the vscode-sas-extension at a clients site. This client has automated processes for generating SSH key pairs and the private keys come with a passphrase by default. The users are not permitted to remove the passphrase from their private keys by a user agreement. However, the new support for remote SAS 9.4 does not allow to use private keys and or a SSH-agent. The use of unprotected SSH private keys is strongly discouraged and companies usually don't permit the use of passphrase-less keys.

Solutions:
- Prompt for the passphrase upon detection of an encrypted private key file (and set the passphrase property of SSH2 Client.connect() method)
- Allow the user to specify the agent socket in the configuration options (and set the agent property of SSH2 Client.connect() method)

The latter being the preferred solution.

smorrisj:
@cstotzer thanks for logging this issue. See attached PR for a proposal on how I think we can solve this. Essentially we can look at the ssh-agent's SSH_AUTH_SOCK env variable and pass this value to the ClientConnect agent property as you mentioned. Typically SSH_AUTH_SOCK gets set on agent startup for linux/macOS environments. Additionally the user would specify the IdentityFile in the global ssh config for the host entry. Based on this we could remove it from the connection profile altogether and just let the global ssh config drive the auth pieces.

The reason I propose to look at SSH_AUTH_SOCK is is that depending on how ssh-agent is set to start we could get different socket values on restarts unless they specify -a to bind to a static values for linux, macOS clients. I noticed on windows, the OpenSSH client optional feature uses a named pipe instead of a file descriptor in /tmp. The named pipe value did not change in between restarts. A side effect on windows is that the user might have to define this user env variable once, set to this static named pipe value( //./pipe/openssh-ssh-agent in my testing with a post build 1809 windows install). Thoughts?

Becky-Williamson:
Leaving open so the reporter can test the fix in his environment (per Joe)

PR #1072, opened 2024-06-29
Feat: Add support for SSH and private key without requiring SSH-agent

chunky:
**Summary**
Patch adds an "identityFile" parameter to the SSH profile, one of the suggested alternate fixes to Issue #1059 .

Using SSH_AUTH_SOCK in our environment is too high friction to be useful, so this provides a fallback option.

**Testing**
I have not yet added formal tests to this. If this request is likely to be accepted, I would appreciate guidance on what an appropriate set of tests would look like.

Signed-off-by: Gary Briggs <chunky@icculus.org>


chunky:
@smorrisj What is the correct way to put in i18n placeholders? Should I just edit each translation file and put the English verbiage in as the translation, and let the translators deal with it?

smorrisj:
> @smorrisj What is the correct way to put in i18n placeholders? Should I just edit each translation file and put the English verbiage in as the translation, and let the translators deal with it?

The typical process is to just fill in the english bundle with strings. We have automation that we run near the end of a release cycle that will propagate the deltas out to the other language bundle files. We have a group of both internal SAS employees and community members that help with translation of the non english files. We typically bundle these up in a PR for merge as we prepare to release.

client/src/components/profile.ts
```diff
@@ -88,6 +88,7 @@ export interface SSHProfile extends BaseProfile {
   saspath: string;
   port: number;
   username: string;
+  identityFile: string;
```
smorrisj, line 91: Can this be made as an optional field? I think having it as optional is more on brand for the overall approaches we recommend for this in the markdown changes. It should also clean up the tests a bit so that the "original" tests dont have to be burdened with keeping up with default values. In my mind we would write a separate set of tests that isolates how this new member gets passed around etc.

smorrisj:
Some context around the introduction to use SSH Agent. We actually used Identify file in the beginning:
https://github.com/sassoftware/vscode-sas-extension/issues/182

client/src/components/profile.ts
chunky: With apologies, this is the first time I've ever used either typescript, or VS Code. In my head this is already "_optional_", in that when the profile is created, you can leave it blank.

I intuit you mean "_optional_" like "_the TS type system affords making it explicit that this field is optional_", but it's not obvious to me how to implement that in a way that aligns with the rest of the codebase. I would appreciate pointers to the right path.

chunky:
I read that, and I do agree that in some grander sense it makes more sense to use agents to handle keys. Unfortunately it just isn't a friendly enough option in our environment.

I think this configuration approach is a reasonable fallback that still uses SSH keys, which users prefer post-initial-setup, to having to enter their password each time.

I have also put in a pull request, #1081, for a mechanism by which users are asked for their password each time a connection is made. Personally I would like to see both of these PR's accepted, to afford compatibility with a wider array of servers, and to give users more choices where they want them.

snlwih:
@chunky , could you explain to me (using Sesamestreet language 😃 ) what the exact use case for this is? 

Now that we're planning for user/pass authentication for SSH connections by storing them locally using the VS Code Secrets API (#1126 ), is there still a need for this? Reason for asking is that you mention this PR being a workaround for #1059, which is also a workaround for being able to use user/pass authentication.

chunky:
@snlwih Sure. There are a few issues with password as a notional magic bullet:
* Some places don't allow password at all [as I understand, that was even in SAS internal Security Group's recommendations].
* Some places require 2-factor, that I believe usually implemented as a keyboard-interactive setup, so could appear to be password but isn't.
* Passwords rotate, possibly even as single-use items.

At the other end of the spectrum is ssh-agent that you support right now. It's great when it works, but the setup is potentially complex and, in the case of our systems, impossible to complete as described on Windows because of some long-standing security policy. Even on systems where it's possible, it requires administrative privileges on the laptop, which most orgs no longer give people. I'm imagining that the typical SAS user doing analytics isn't in the category of people specially granted admin rights, at orgs where that's on the table.

Loading private keys from a file gives a middle-ground. There's the additional convenience of private keys and avoidance of password-related problems, but without any complexity or logistical-impossibility for the user to do initial setup. On my windows laptop, this would be my own personal preference. I also see it as the one easiest to handhold my users through, if they need to go one notch beyond password.

snlwih:
@chunky, thanks for your patience and education 😃 
  - when  understand things correctly this PR is about using private key file without needing SSH-agent. Right? If so, I will rename the title of this PR.
  - as part of PR #1126 there will be support for using a private key file without needing SSH-agent, so are you okay with closing this PR?

chunky:
Yes, please, feel free to close. Same as the other pr, I just want stuff that works!

And yes, this was for using pk without agent.

Issue #889, opened 2024-03-14
[Local/Remote IOM connection] - Support SAS 9 server file system


snlwih:
**Is your feature request related to a problem? Please describe.**
I need to be able to work with files that are stored on the file system of the (remote) SAS Server 

**Describe the solution you'd like**
Similar to connections to Viya 4, I would like to see a SAS Explorer panel that allows me to navigate the SAS server file system, create/edit/delete/rename files and folders, and Upload Files and Folders from the local file system where VSCode is running. 

**Environment**
SAS version: SAS 9.4

clangsmith:
This should likely be done in coordination with the same feature request for Viya (#417), since they should have a largely consistent experience.

scottdover:
Hey  @smorrisj and @clangsmith. I'm wondering if I can get your thoughts on design approach for this (assuming this will include implementing SAS server for ITC/COM & Viya)

Here's some background of how things work currently. Right now, we have the following concepts:

- `ContentNavigator`: This is the outer most layer for sas content. It registers our commands and delegates those to our `ContentDataProvider`
- Our `ContentDataProvider` houses the tree structure for all of sas content. It's kind of like our application service for sas content in that it talks to our `ContentModel` and prepares the data for the ui
- Our `ContentModel` interacts with our rest api to call endpoints and turn raw data into `ContentItem`s for the outer layers to consume
- `ContentItem` is a representation of a file/folder in the tree

Right now, these are _all_ viya specific and SAS content specific. Here's how I'm thinking of changing things to support SAS server. As an aside, right now I'm going with the assumption that the "content" references above are not about "SAS Content" specifically, but more about "content" (as in the contents of a file).

 - Update `ContentNavigator` to accept a "config" which will contain: `treeIdentifier` (string), `mimeType` (string), and `sourceType` (either "SASContent" or "SASServer")
 - Update `ContentDataProvider` to accept a `ContentModel`, a `treeIdentifier`, and a `mimeType` (`treeIdentifier` and `mimeType` will need to be distinct between sas content/sas server)
 - In our extension, I'll update things to construct two content navigators. One with things relevant to sas content; one with things relevant to sas server. This will basically mean seeing "SAS Content" and "SAS Server" as two of the accordions in the extension
 - I'll create a `ContentAdapterFactory` that returns a `ContentAdapter` based on connection type _and_ `sourceType`. Here are the adapters I think we'll need:
[...]

PR #1079, opened 2024-07-01
feat: make content models connection agnostic

scottdover:
**Summary**
This PR is a prerequisite  for integrating sas server into the extension (see overall design, here: https://github.com/sassoftware/vscode-sas-extension/issues/889). This makes a few additional refactors outside of those mentioned in https://github.com/sassoftware/vscode-sas-extension/issues/889.

Favorites. Previously, we were storing a favorite id w/n `ContentItem.flags`. ~This change delays finding that favorite id _until_ we're actually doing something with that favorite.~ Now, we are storing the uri to simplify the deletion process. 

Notebook -> flow conversion. When we first implemented this change, there was some question around whether this belong in content data provider or not. In this PR, functionality is moved to a `NotebookToFlowConverter` which makes use of a `ContentModel` directly.

It's also worth mentioning that this moves some additional properties into `ContentItem` that were previously utils functions, and updates _every_ content adapter method to return a content item decorated with those new properties. Additionally, this splits utils such that `ContentNavigator/utils` has things that are connection agnostic, while `rest/utils` contains things that are connection specific.

**Testing**
 - [x] Test creating file
 - [x] Test creating folder
 - [x] Test moving files & folders
 - [x] Test recycling files & folders
 - [x] Test deleting files & folders
 - [x] Test add to favorites for files and folders
 - [x] Test remove from favorites for files and folders
 - [x] Test editing files (regular files & favorites & favorites w/n folders)
 - [x] Test downloading files and folders
 - [x] Test converting sas notebooks to flows (test in explorer as well as sas content)
 - [x] Test renaming files & folders
 - [x] Test emptying recyling bin
 - [x] Test refreshing content
 - [x] Test filename statement
 - [x] Test fetching viya cadence
 - [x] Test uploading files (context menu, dragging from file system, dragging from explorer)
 - [x] Test collapsing all files
 - [x] Make sure multi-select is working the same on this branch and main
 - [x] Test removing a favorited item from 'my folder'

**Todos**
 - [x] Address test coverage/build issues

**Observations**
 - Unable to download flows. This is because flows are stored at a different endpoint than regular files


client/src/connection/rest/SASContentAdapter.ts
```diff
@@ -0,0 +1,707 @@
...
+
+    async function getTypeDefinition(fileName: string): Promise<string> {
+      const defaultContentType = "file";
+      const ext = fileName.split(".").pop().toLowerCase();
+      if (ext === "sas") {
+        return "programFile";
+      }
+
+      try {
+        const typeResponse = await this.connection.get(
```
scottdover, line 458: Make sure this works

client/src/connection/rest/SASContentAdapter.ts
```diff
@@ -0,0 +1,707 @@
...
+      return false;
+    }
+    return true;
+
+    async function getDeleteMemberUri(): Promise<string> {
+      if (item.flags?.isInMyFavorites) {
+        return getLink(item.links, "DELETE", "delete")?.uri;
+      }
+
+      const myFavoritesFolder = this.getDelegateFolder("@myFavorites");
```
scottdover, line 565: Perhaps this does work?

scottdover:
Hey @smorrisj. I made some changes matching the general design of https://github.com/sassoftware/vscode-sas-extension/issues/889.

That said, after these changes, `ContentModel` looks pretty slim and serves no _real_ purpose of its own other than:
 - Presenting some error messaging back to content data provider
 - providing a different data format (i.e. string vs buffer)

After working through this, it almost feels like `ContentModel` should be an abstract class that connection-specific things extend. Thoughts?

scottdover:
@scnwwu wondering if you have any thoughts about the above?

scnwwu:
> @scnwwu wondering if you have any thoughts about the above?

defer to @boyce-w who is more familiar with the Content part.

boyce-w (approved):

Zhirong2022:
ContextMenu name changed from 'Remove from My Favorites' to 'Add to My Favorites' if the file already has been added to My Favorites.
1.Create a file under My Folder
2.Add the file to My Favorites through ContextMenu
3.Open ContextMenu against the file again
image

Zhirong2022:
Multi-select is not working the same on this branch and main.
Multi-selection through pressing Shift or Ctrl works well on this branch. It does not work well on main.

Zhirong2022:
Multi-select some files and perform 'Add to My Favorites' through context menu, it does not move all files 
image

scottdover:
> ContextMenu name changed from 'Remove from My Favorites' to 'Add to My Favorites' if the file already has been added to My Favorites. 1.Create a file under My Folder 2.Add the file to My Favorites through ContextMenu 3.Open ContextMenu against the file again image

> Multi-select some files and perform 'Add to My Favorites' through context menu, it does not move all files image

It looks like this second one is an issue on main as well. Both of these have been resolved in https://github.com/sassoftware/vscode-sas-extension/pull/1079/commits/da0417b33d5994c55590fc5b73c36259c884ce65 .

> Multi-select is not working the same on this branch and main. Multi-selection through pressing Shift or Ctrl works well on this branch. It does not work well on main.

Is this still an issue for you, @Zhirong2022 . I noticed this problem on main and my branch several weeks ago when implementing this, but it since seems to have resolved itself (I'm guessing it may've been a vscode bug)

Zhirong2022:
@scottdover , thank you for your help.  I upgraded my VSC to the latest version, multi-selection works well through pressing Shift or Control.  And other issues have been fixed in the branch package.

Zhirong2022 (approved):

PR #1098, opened 2024-07-15
fix: library pane crash with option symbolgen

### LINKED ISSUE #1012: SAS: Libraries crash --> Element with id .SASHELP is already registered
**Describe the bug**
After running various programs SAS libraries disappear and an error with the text "Element with id .SASHELP is already registered" appears in the bottom right corner. 

**Steps to reproduce**
SAS 9.4 and visual studio code on windows 10

**Expected behavior**
Library visible and possible to update with data steps

Client OS: Windows 10
Extension version: v1.9.0
SAS version: 9.4

scnwwu:
**Summary**
Fix #1012
Put the start tag in the same statement with macro resolution so that it will be after the extra log produced by `options symbolgen` 

**Testing**
Test case in #1012

### DIFF
--- a/client/src/connection/itc/ItcLibraryAdapter.ts
+++ b/client/src/connection/itc/ItcLibraryAdapter.ts
@@ -54,7 +54,7 @@ class ItcLibraryAdapter implements LibraryAdapter {
         where libname='${item.library}' and memname='${item.name}'
         order by varnum;
       quit;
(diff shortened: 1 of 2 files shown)


scottdover (approved):

Zhirong2022:
There is no error "Element with id .SASHELP is already registered" and the library pane is shown as expected without crash.

Zhirong2022 (approved):

Issue #1091, opened 2024-07-09
VS Code Connection To SAS Server Using Remote SSH Fails


jandewitte:
Hi

I use a locally installed VS Code editor installed on a Windows 11 machine and try to connect to a remote SAS 9.4 M7 engine on a Windows 2019 server in AWS. I have v1.10.0 of the SAS Extension for VS Code installed. I set up an SSH connection with public and private key. I can connect to the SAS server from my workstation over SSH. I have configured a specific SSH connection profile as well. When I run my simple program I get an 'All configured authentication methods failed' message (see screenshot 2). I have the impression I am signed in (see screenshot 1), but I am not really sure. I would expect a pop up to provide username and password when I run the program as well.

Could you help to troubleshoot this issue?

Thanks
Jan

![Screenshot VS Code SAS sign in](https://github.com/sassoftware/vscode-sas-extension/assets/72504950/72c6785c-2d9e-41b6-a2da-770b664c67a7)
![Screenshot VS Code auth fails](https://github.com/sassoftware/vscode-sas-extension/assets/72504950/41ad7f0f-2f93-4422-a3fc-d56f025dfc0b)

scnwwu:
are you following the steps in https://sassoftware.github.io/vscode-sas-extension/Configurations/Profiles/sas9ssh?

jandewitte:
Yes, I followed these steps. Is there anything I can do to get more insight in what is going wrong? Thanks.

scnwwu:
Is there any error/warning in the console? (Click `Help > Toggle Developer Tools` from the top menu bar.)

jandewitte:
Yes, I enclosed them.
![image](https://github.com/sassoftware/vscode-sas-extension/assets/72504950/87207f74-04d0-463c-9f90-99fedb7e4d43)

jandewitte:
Sorry, the screenshot enclosed below came before the previous one in the console and the first error matches the popup in the GUI.
![image](https://github.com/sassoftware/vscode-sas-extension/assets/72504950/8f5e8faf-f3ce-473b-9319-022ed0a003e6)

scnwwu:
Would you also post a screen of the console above the 'All configured authentication methods failed' error? Thanks.

jandewitte:
Yes, but it is a repetition of the same sequence.

![image](https://github.com/sassoftware/vscode-sas-extension/assets/72504950/85b6fe27-505d-435c-8702-f02c4d83980c)

[...]

PR #1126, opened 2024-08-01
feat: username and password support for SSH connection type

smorrisj:
**Summary**
- Adds support for username and password to the 9.4 Remote (SSH) connection type.
- Refactor SSH auth process to use a handler that can support use of agent, manually specifying a private keyfile (with or without a passphrase) in the connection profile,  using a traditional username password, or using keyboard-interactive mode, depending on how the ssh server is configured.
- Run initialization code to set working dir to the WORK lib path to autocleanup ODS results on session close
- Add keepalive configuration to ssh client to prevent idle disconnects for long-running programs

Todos:
- [x] Use authHandler to handle authentication requests. See [ssh2](https://github.com/mscdex/ssh2) docs for authHandler support. See prototype by @chunky [here](https://github.com/sassoftware/vscode-sas-extension/commit/d276e0c27026fa18afb9ebbb1fb61824016af10d). We need to give some thought on how we define what supported auth methods should be attempted. I think it would be great to let the server tell us what the supported methods are, and then react to them. Kerberos methods are going to be a problem because the ssh2 library cant support those auth methods currently. For each supported method, the auth should be attempted, moving onto the next one during a failure to avoid issues like [this](https://github.com/sassoftware/vscode-sas-extension/issues/1091#issuecomment-2225596787). 
- [x] Username/Password Support. Prompt for password "Just in time" on session establishment. ~~Store password in Secrets API so that the user is not prompted for a password on each session establishment after initial profile creation.~~ Use masked input field like we do for IOM on the text input prompt.
- [x] Keyboard Interactive support. Similar to, but not the same as password. MFA scenarios will request this auth method. We need to detect the incoming requests for input and prompt the user for answers, before sending on the answers to the auth process.
- [x] Maintain support for agent based authentication for users that want passwordless or agent based setup. Users should still be able to specify an identify file for keys that have a passphrase.
- [x] The `setTimeout` logic will need to be adjusted so that we dont errantly timeout while a user is typing in a password or passphrase.
- [x] Add username to password prompt text 
- [x] Add vscode metadata for new privateKeyFilePath profile field
- [x] Add support for work directory detection for #1176 
- [x] Add keepalives and max session time config options for #1156
- [x] Unit tests
- [x] Doc updates

**Testing**

Connection Profile Prompts:
- Create new connection profile should offer private key path as an optional prompt and create the profile with the value that the user specifies

Auth Refactor:
Users can successfully authenticate to the server to run sas programs using the following auth methods:
- SSH Agent (See current doc)
- Private key specified  in privateKeyFilePath field in connection profile (no passphrase)
- Public key specified in privateKeyFilePath field in connection profile (with a passphrase)
- Using a traditional username/password
- Using Keyboard Interactive (RHEL disables this by default in modern versions, will need to enable it in sshd config on the server and set the AuthenticationMethods field such that this method is used instead of password)

Work Directory Detection
- ODS SAS result html files are no longer written out to the users home directory on the SSH Server
- Verify that the SAS Log Output contains an output entry for the working dir getting set to the WORK dir generated.
- User can change the work directory by passing -WORK [path] option as a sasOption on the connection profile options array

Keepalive Changes:
- Verify that long running program (using sleep function) does not result in the ssh server disconnecting the client


smorrisj:
Pushed a super rough draft of auth handler logic this morning. Very rough around the edges but wanted to get thoughts down in some (semi) organized manner. Will further test and refine as time allows.

chunky:
First up, thank you for working on this. It'll be a big improvement for us.

A few random thoughts:
1. You can find a conversation I had with a maintainer of ssh2 about authhandlers here: https://github.com/mscdex/ssh2/issues/1400 . Things that might be relevant from your comments and code so far:
    1. The authhandler is only called with a list of methods that would work for that server
    2. "agent" and "private key" are different, in the authhandler callback.
    3. Note this from the ssh2 doc you link: "_Valid method names are: ''none', 'password', 'publickey', 'agent', 'keyboard-interactive', 'hostbased'_"
2. I feel pretty strongly that attempting all non-interactive options on the table first, and only then trying the interactive ones is the right choice. If agent would be accepted, and I have a running agent, then _obviously_ that should be tried before asking users for a password, for example.
4. Don't forget keyboard-interactive, as separate from password. Some server configurations ask for "keyboard-interactive" authentication and ask you for a password in a message, rather than asking for "password" authentication. [set sshd config "ChallengeResponseAuthentication yes" to see this]
5. Affording use of an unshielded private key as a separate config item, and proffering it if "private key" method is available, would really help
6. I have no strong feelings about password storage in secrets. The main concern is, users have to rotate passwords regularly; what happens if the password in storage is no longer the same one that works for the server? Honestly for the first pass, asking for it each time doesn't seem unreasonable.
    1. users can set up public key auth if they don't like it

From your top message: "Users should still be able to specify an identify file for keys that have a passphrase." ; do you mean "for keys that _don't_ have a passphrase"? For keys that do, you would need to develop the ability to de-shield them after loading them from disk, to pass to the private key method. Probably safer to just expect users to set up ssh-agent, if that's where they're at. [although it's true that that is unfortunate on windows]

client/src/connection/ssh/index.ts
```diff
@@ -33,20 +46,26 @@ export function getSession(c: Config): Session {
...
+  private _cancellationSource: CancellationTokenSource | undefined;
+  private _secretStorage;
+  private _passwordKey: string;
 
   constructor(c?: Config) {
     super();
     this._config = c;
-    this.conn = new Client();
+    this._conn = new Client();
+    this._authMethods = ["publickey", "password", "keyboard-interactive"];
```
smorrisj, line 67: Attempt non-interactive methods first, falling back to interactive ones.

client/src/connection/ssh/index.ts
```diff
@@ -233,6 +256,155 @@ export class SSHSession extends Session {
...
+          if (this._config.privateKeyFilePath) {
+            //TODO: handle "file not found" errors
+            const keyContents = readFileSync(this._config.privateKeyFilePath);
+            const parsedKeyResult = utils.parseKey(keyContents);
+            // key is encrypted, prompt for passphrase
+            if (
+              parsedKeyResult instanceof Error &&
+              parsedKeyResult.message ===
+                "Encrypted OpenSSH private key detected, but no passphrase given"
+            ) {
```
smorrisj, line 336: The ssh2 library has a function that can assist with detecting that the key file specified on the config property is encrypted, which is a nice help to allow users to specify shielded and unshielded keys on the session object (if they dont want to use agent). Right now with the code below if we detect a passphrase we'd prompt for it.

client/src/connection/ssh/index.ts
```diff
@@ -233,6 +256,155 @@ export class SSHSession extends Session {
...
+              //TODO: refactor typechecking into one place, maybe a utility function
+              if (!(parsedKeyResult instanceof Error)) {
+                cb({
+                  type: "publickey",
+                  key: parsedKeyResult,
+                  username: this._config.username,
+                });
+              }
+            }
+          } else if (process.env.SSH_AUTH_SOCK) {
```
smorrisj, line 364: Pretty straightforward, this is the agent method that we supported in the before state. Users can still elect not to specify a key file on the connection profile and have the agent do all of the work (if they have set it up).

client/src/connection/ssh/index.ts
```diff
@@ -233,6 +256,155 @@ export class SSHSession extends Session {
...
+            //attempt to auth using ssh-agent
+            cb({
+              type: "agent",
+              agent: process.env.SSH_AUTH_SOCK,
+              username: this._config.username,
+            });
+          }
+          break;
+        }
```
smorrisj, line 373: Support both password and keyboard-interactive. Prompt user "just in time".

smorrisj:
> First up, thank you for working on this. It'll be a big improvement for us.
> 
> A few random thoughts:
> 
> 1. You can find a conversation I had with a maintainer of ssh2 about authhandlers here: [Request: Delay client password entry until password actually needed mscdex/ssh2#1400](https://github.com/mscdex/ssh2/issues/1400) . Things that might be relevant from your comments and code so far:
>    
>    1. The authhandler is only called with a list of methods that would work for that server
>    2. "agent" and "private key" are different, in the authhandler callback.
>    3. Note this from the ssh2 doc you link: "_Valid method names are: ''none', 'password', 'publickey', 'agent', 'keyboard-interactive', 'hostbased'_"
> 2. I feel pretty strongly that attempting all non-interactive options on the table first, and only then trying the interactive ones is the right choice. If agent would be accepted, and I have a running agent, then _obviously_ that should be tried before asking users for a password, for example.
> 3. Don't forget keyboard-interactive, as separate from password. Some server configurations ask for "keyboard-interactive" authentication and ask you for a password in a message, rather than asking for "password" authentication. [set sshd config "ChallengeResponseAuthentication yes" to see this]
> 4. Affording use of an unshielded private key as a separate config item, and proffering it if "private key" method is available, would really help
> 5. I have no strong feelings about password storage in secrets. The main concern is, users have to rotate passwords regularly; what happens if the password in storage is no longer the same one that works for the server? Honestly for the first pass, asking for it each time doesn't seem unreasonable.
>    
>    1. users can set up public key auth if they don't like it
> 
> From your top message: "Users should still be able to specify an identify file for keys that have a passphrase." ; do you mean "for keys that _don't_ have a passphrase"? For keys that do, you would need to develop the ability to de-shield them after loading them from disk, to pass to the private key method. Probably safer to just expect users to set up ssh-agent, if that's where they're at. [although it's true that that is unfortunate on windows]

All good feedback. I generally agree with thoughts above. I've put comments around changes that I think support these items. On the issue of whether or not to store the passwords in secrets, we have some precedent for doing this in other connection types. I do agree though that generally for ssh interaction with other solutions, anytime passwordless or passphraseless interactions are intended, that I've seen docs push users towards agent or public key.

jbreitman:
I would like to see the authentication method gssapi-with-mic added.  This will be beneficial to those in corporate environments where Kerberos Tickets enable you to access other resources such as file systems, databases, APIs, etc ... while eliminating the need for the person to enter their credentials.

chunky:
If you can figure it out, yes, the kerberos SSO stuff is great where available. That was another thing I couldn't figure out. It works on some of our servers. Another of those "if it works it's magical and absolutely preferred".

I think the distinction between "agent" and "public key" is a manifestation from the ssh2 library, not the server. In practice it might mean taking two bites at the key-based apple, which is explicitly afforded by the Auth handler mechanism

smorrisj:
> I think the distinction between "agent" and "public key" is a manifestation from the ssh2 library, not the server. In practice it might mean taking two bites at the key-based apple, which is explicitly afforded by the Auth handler mechanism

Yea that's what the current changeset has, inside of "publickey" we'd look for whether or not the user has set their keyfile on the connection profile, if so we'd use that file, otherwise we'd defer to agent.

> I would like to see the authentication method gssapi-with-mic added. This will be beneficial to those in corporate environments where Kerberos Tickets enable you to access other resources such as file systems, databases, APIs, etc ... while eliminating the need for the person to enter their credentials.

One challenge is that the ssh2 library doesnt support GSSAPI methods yet: 
https://github.com/mscdex/ssh2/issues/333

chunky:
@smorrisj I missed this comment from earlier: "_I'm seeing the auth method for agent and public key coming back from the server simply as "publickey". I'm wondering if the library is adding in agent based on set config options?_"

From the server's perspective, these are the same thing; it just wants to do a key-based handshake. But on the client, they're different beasts. With ssh2's 'agent' they go get it from a daemon running on a socket, while 'publickey' is just a thing where you hand ssh2 the ready-to-use bytes.

website/docs/Configurations/Profiles/sas9ssh.md
scnwwu, line 1: cc @jennifert-sas for review

website/docs/Configurations/Profiles/sas9ssh.md
```diff
@@ -118,3 +129,26 @@ Host host.machine.name
 ```
 
 6. Add the public part of the keypair to the SAS server. Add the contents of the key file to the ~/.ssh/authorized_keys file.
+
+### Connection Profile
```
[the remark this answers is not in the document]
smorrisj: Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/ea17b7554c9090c05c893abe452a74becde45891

client/src/connection/LineParser.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright © 2024, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export class LineParser {
```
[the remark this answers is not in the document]
smorrisj: Thanks for catching. I had originally promoted LineParser up higher. A merge must have errantly added it back. Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/76a271f3e526a476e63d776d5cecb539100ab04b.

client/src/components/profile.ts
```diff
@@ -794,6 +804,13 @@ const input: ProfilePromptInput = {
     placeholder: l10n.t("Enter your username"),
     description: l10n.t("Enter your SAS server username."),
   },
+  [ProfilePromptType.PrivateKeyFilePath]: {
+    title: l10n.t("Private Key File Path (optional)"),
+    placeholder: l10n.t("Enter the private key file path"),
+    description: l10n.t(
+      "Enter the local path of the private key file. Leave empty to use SSH Agent.",
```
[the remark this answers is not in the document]
smorrisj: I was already unsure about having such a long description. I think we could shorten it a bit. See https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/7a2c0b4d2721451508128fc8e05d67a5e15250a0. Will that work?

client/src/connection/ssh/index.ts
```diff
@@ -157,28 +173,62 @@ export class SSHSession extends Session {
...
+    if (this._workDirectoryParser.isCapturingLine()) {
+      return;
+    }
+
+    return foundWorkDirectory || "";
+  };
+
+  private resolveSystemVars = (): void => {
+    const code = `%let workDir = %sysfunc(pathname(work));
+    %put ${WORK_DIR_START_TAG};
```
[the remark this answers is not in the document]
smorrisj: I noticed this in testing. One "chunk" of data written to the stream actually contained multiple long lines with breaks on my test environment (I used VA). In testing, this code was able to parse out the actual unix path from the input strings I was seeing come across. Take a look and let me know what you think?

```javascript
const match = foundWorkDir.match(/\/[^\s\r]+/);
this._workDirectory = match ? match[0] : "";
```

client/src/connection/ssh/types.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export interface Config extends BaseConfig {
+  host: string;
+  username: string;
+  saspath: string;
+  port: number;
+  privateKeyFilePath: string;
+}
+
+export const LogLineTypes: LogLineTypeEnum[] = [
```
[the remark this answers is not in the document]
smorrisj: Thanks. Relic from the initial prototype / approach. Removed in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/3e1f267322b20378f756f30d9132a82f4a10f652.

scnwwu (approved):
Looks good to me. Thanks.

client/src/connection/ssh/const.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright © 2024, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const SECOND = 1000;
+export const KEEPALIVE_INTERVAL = 60 * SECOND; //How often (in milliseconds) to send SSH-level keepalive packets to the server. Set to 0 to disable.
```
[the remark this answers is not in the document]
chunky: @smorrisj  60 seconds is a reasonable figure - I think I usually set it to 30, but I have no strong feelings as long as it's well inside 90.
The unanswered_threshold, I confess to be unsure about appropriate behaviour. If the comment is right, it's measured in "pings" not in "time". So if you're aiming for 12 hours, it should be (12 * HOUR / KEEPALIVE_INTERVAL). Either way, because this is happening at the SSH/networking level, then 12 hours is probably way too long. Maybe 15 minutes?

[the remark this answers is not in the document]
smorrisj: Thanks @chunky. Must have forgot my morning coffee when making the first pass on that value =) See https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/04444a735da125568a1bfba2e2cbd5fdedba0875.

website/docs/Configurations/Profiles/sas9ssh.md
jennifert-sas: @scnwwu  For "Leave empty to use SSH Agent or password."  I would suggest:
"To use the ssh-agent or a password, leave blank."
Not entirely sure if it should be "SSH Agent" or "ssh-agent" - i see both in the GitHub doc.

smorrisj: `ssh-agent` typically refers to the running process in the context of running in the operating system, whereas we generally have been using `SSH Agent` for a more general reference, regardless of operating system. I updated the prompt to use "SSH Agent" in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/5f9ec571cab6b1675c81821a773ee47bbdc85589, since in this case we were referring to it more as a general reference?

jennifert-sas: Sounds good - thanks!

Zhirong2022:
1.Create a SSH connection profile and specify an invalid  privateKeyFilePath, run some code, it will prompt error message as expected.
2.Update the profile to correct the file path, run code, it has the same error as before.
The only way to restart VSC to use the updated profile.

Zhirong2022:
1.Create a SSH connection profile, specify privateKeyFilePath field in connection profile (with a passphrase). 
2.Run some code, it will ask the user to enter the passphrase for the private key
3.Cancel the dialog or type an invalid password in such scenario
Result: It keeps showing 'Connecting to SAS session' until it gets error message 'Could not connect to the SAS server.'
Suggestion: If the user cancel or type with an invalid password, it can stop the process right now without having to wait for some time.

Zhirong2022:
It cannot create the profile successfully if leave the 'Enter the local private key file path' as blank. For others fields have been filled correctly.

smorrisj:
> It cannot create the profile successfully if leave the 'Enter the local private key file path' as blank. For others fields have been filled correctly.

Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/d3c9cd20a7ed32da89318826b0ede0239c067780

smorrisj:
> It cannot create the profile successfully if leave the 'Enter the local private key file path' as blank. For others fields have been filled correctly.

Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/d3c9cd20a7ed32da89318826b0ede0239c067780

smorrisj:
> 1.Create a SSH connection profile and specify an invalid privateKeyFilePath, run some code, it will prompt error message as expected. 2.Update the profile to correct the file path, run code, it has the same error as before. The only way to restart VSC to use the updated profile.

Fixed in [6d28a43](https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/6d28a434f55d7535083c89846984f14e7253d591)

smorrisj:
> 1.Create a SSH connection profile, specify privateKeyFilePath field in connection profile (with a passphrase). 2.Run some code, it will ask the user to enter the passphrase for the private key 3.Cancel the dialog or type an invalid password in such scenario Result: It keeps showing 'Connecting to SAS session' until it gets error message 'Could not connect to the SAS server.' Suggestion: If the user cancel or type with an invalid password, it can stop the process right now without having to wait for some time.

In this setup, we're trying as much as possible to let the SSH client and server negotiate auth based on the server configuration. If you're seeing that message, it means that the SSH client received a disconnect from the server side to close the connection before auth could finish. This can happen in a number of auth setups, for example, if max auth tries are exceeded etc. I think to support the largest number of customer configured auth setups, we should keep this the way it is.

Zhirong2022:
Private key specified in privateKeyFilePath field in connection profile (with a passphrase)
1.Remove the public key information from ~/.ssh/authorized_keys file for the ssh server. 
2.Run some code
Results: 

1. It keeps asking for passphrase (3 times at least)
2. It keeps showing 'Connecting to SAS session...' without further response

Zhirong2022:
There are two SSH connection profiles and both do not work properly.  One was configured as an invalid private key file path, the other one was not reachable with error like 'getaddrinfo ENOTFOUND...',  switch between the two profiles, run some code, it did not use the specified one.

smorrisj:
> There are two SSH connection profiles and both do not work properly. One was configured as an invalid private key file path, the other one was not reachable with error like 'getaddrinfo ENOTFOUND...', switch between the two profiles, run some code, it did not use the specified one.

Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/a0b7f88e5137f05462e2c7e7290d69bf8cd4d0aa

smorrisj:
> Private key specified in privateKeyFilePath field in connection profile (with a passphrase) 1.Remove the public key information from ~/.ssh/authorized_keys file for the ssh server. 2.Run some code Results:
> 
> 1. It keeps asking for passphrase (3 times at least)
> 2. It keeps showing 'Connecting to SAS session...' without further response

Refactored error handling to fix issues like this in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/a0b7f88e5137f05462e2c7e7290d69bf8cd4d0aa

Zhirong2022:
1. Input an invalid password or cancel password typing, it will show message 'Encrypted private OpenSSH key detected, but no passphrase given'. 
2. Run the code again it will show message 'Could not connect to the SAS server.' and passphrase input dialog will be opened at the same time. 
3. Now input the correct passphrase, but, there is no corresponding result. Run the code again and type the passphrase, it will work as normal.
```
proc print data=sashelp.cars(obs=1);run;
```

Zhirong2022:
Private key specified in privateKeyFilePath field in connection profile (with a passphrase) 
1.Remove the public key information from ~/.ssh/authorized_keys file for the ssh server.
2.Run some code 
Result:
It kept asking for passphrase if the user typed the correct password until the user got message 'Too many authentication failures'.

smorrisj:
> Private key specified in privateKeyFilePath field in connection profile (with a passphrase) 1.Remove the public key information from ~/.ssh/authorized_keys file for the ssh server. 2.Run some code Result: It kept asking for passphrase if the user typed the correct password until the user got message 'Too many authentication failures'.

There was a lifecycle issue where the ssh socket was dangling open after the run errored out. Fixed in https://github.com/sassoftware/vscode-sas-extension/commit/6e464402422436e98f8ead38581cfa991f6cbd72

smorrisj:
> 1. Input an invalid password or cancel password typing, it will show message 'Encrypted private OpenSSH key detected, but no passphrase given'.
> 2. Run the code again it will show message 'Could not connect to the SAS server.' and passphrase input dialog will be opened at the same time.
> 3. Now input the correct passphrase, but, there is no corresponding result. Run the code again and type the passphrase, it will work as normal.
> 
> ```
> proc print data=sashelp.cars(obs=1);run;
> ```

Fixed in https://github.com/sassoftware/vscode-sas-extension/commit/6e464402422436e98f8ead38581cfa991f6cbd72. We now properly close the ssh connection when the user cancels out of the passphrase dialog. Entering blank on password will be treated as an invalid password attempt. Retries allowed are governed by the ssh server settings. So for that it's possible that the user will continue to be prompted until the server responds with "Too many authentication failures", where it will then close the connection.

Zhirong2022:
There is no result output if specify a long workdir and a specific pagesize in sasOptions.
```
proc print data=sashelp.cars(obs=1);run;
```

smorrisj:
> There is no result output if specify a long workdir and a specific pagesize in sasOptions.
> 
> ```
> proc print data=sashelp.cars(obs=1);run;
> ```

The work dir code was setting up the value over multiple lines. In this approach, for long directory values we were getting log output intermixed in between the start element, end element, and directory value. Fixed in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/2195eb967441272ae57c83a553f52dcf15a77c50

Zhirong2022:
> > There is no result output if specify a long workdir and a specific pagesize in sasOptions.
> > ```
> > proc print data=sashelp.cars(obs=1);run;
> > ```
> 
> The work dir code was setting up the value over multiple lines. In this approach, for long directory values we were getting log output intermixed in between the start element, end element, and directory value. Fixed in [2195eb9](https://github.com/sassoftware/vscode-sas-extension/commit/2195eb967441272ae57c83a553f52dcf15a77c50)

@smorrisj  I checked the issue in the latest PR build. I can still reproduce it if specify the page size as new value '23'.

smorrisj:
Thanks @Zhirong2022. The log parsing for detecting work dir still was not robust enough. Refactored the code that sets work dir to use a more structured logging output. See fix in https://github.com/sassoftware/vscode-sas-extension/pull/1126/commits/cde43f065bd1ddeda4281fee5577014d5439222e.

Zhirong2022:
All raised issues have been fixed.
Error validation works as following:

**Feature: Public key specified in privateKeyFilePath field in connection profile (with a passphrase)**
1.Type Enter and close the passphrase dialog
Error message: Failed to generate information to decrypt
2.Type ESC to cancel the passphrase typing
Error message: Encrypted private OpenSSH key detected, but no passphrase given key
3.Input invalid passphrase
Error message: OpenSSH key integrity check failed -- bad passphrase?
4.Remove the public key from the SSH sever with passphrase and type the correct passphrase several times
Error message: Too many authentication failures
5.Leave the passphrase dialog there and waiting
Error: Could not connect to the SAS server.

**Feature: Traditional username/password**
1.Type Enter to close the password typing or Input invalid password
It attempts to ask the user to type the password for several times
Error message: Too many authentication failures
2.Type ESC to cancel the passphrase typing
Error message: All configured authentication methods failed
3.Leave the password dialog there and waiting
Error: Could not connect to the SAS server.

Zhirong2022 (approved):

Issue #963, opened 2024-04-26
Log information are missing in certain case with Viya connection


Zhirong2022:
**Describe the bug**
Open VSC SAS Extension base on main branch. Try to run some code, log contents are not fully shown on output log panel.

**Steps to reproduce**
1. Open VSC SAS Extension
2. Log in with a valid Viya connection
3. Type code below
```
data hernio;
   input patient age gender$ OKstatus leave los;
   datalines;
 1   78  m   1   0   9
 2   60  m   1   0   4
 3   68  m   1   1   7
 4   62  m   0   1  35
 5   76  m   0   0   9
 6   76  m   1   1   7
 7   64  m   1   1   5
 8   74  f   1   1  16
 9   68  m   0   1   7
10   79  f   1   0  11
11   80  f   0   1   4
12   48  m   1   1   9
13   35  f   1   1   2
14   58  m   1   1   4
15   40  m   1   1   3
16   19  m   1   1   4
17   79  m   0   0   3
18   51  m   1   1   5
19   57  m   1   1   8
20   51  m   0   1   8
21   48  m   1   1   3
22   48  m   1   1   5
23   66  m   1   1   8
24   71  m   1   0   2
25   75  f   0   0   7
26    2  f   1   1   0
27   65  f   1   0  16
28   42  f   1   0   3
29   54  m   1   0   2
30   43  m   1   1   3
31    4  m   1   1   3
32   52  m   1   1   8
;

data hernio_uv;
   length dist $7;
   set hernio;
   response = (leave=1);
   dist     = "Binary";
   output;
   response = los;
   dist     = "Poisson";
   output;
   keep patient age OKstatus response dist;
run;

proc glimmix data=hernio_uv(where=(dist="Binary));
   model response(event='1') = age OKStatus / s dist=binary;
run;

```
4. Run the code 

**Expected behavior**
Log contents can be shown completely as other connections (Local, remote-IOM, remote-SSH)

**Screenshots**
![image](https://github.com/sassoftware/vscode-sas-extension/assets/121005702/ce720df9-af6b-41f1-a2bc-b1c37c76976b)

**Environment (please complete the following information):**
Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.8.0]

sasutils:
You have unbalanced quotes in the dataset option here:

data=hernio_uv(where=(dist="Binary));

PR #1151, opened 2024-08-13
fix: workaround compute log display issues

smorrisj:
**Summary**
Simple workaround to introduce a newline at the end of the postamble to ensure that the last bit of log displays properly in the given error scenario. We'll be following up internally with the compute api team to understand why this is required.

**Testing**
Sample program from #963 for 9x and Viya connections.


scnwwu (approved):

Zhirong2022:
The issue was fixed. Log contents can be fully shown on output log panel with Viya connection.

Issue #867, opened 2024-03-01
Better organization of log output for code executions


smorrisj:
(Creating on behalf of internal user feedback)

**Is your feature request related to a problem? Please describe.**
Currently, all code executions from various sas files (not sas notebook) are appended to a single output channel called SAS Log. This can make it confusing to follow over time for many executions. The ask is that we  offer a way to better organize output log generation from various sas program runs. One such option might be to have one output channel per run.

smorrisj:
Just brainstorming, another option might be to keep one output channel, but switch out its contents when the user changes active document / tab focus.

snlwih:
Users seem to like what SAS Studio does in this context. SAS Studio allows for:
 * Having multiple SAS programs open concurrently, and each maintaining their own Log, Results, Output Data tab

But it also allows for submitting multiple SAS programs concurrently. This then results in them being queued (FIFO) and run sequentially in the same SAS Compute session. Queued Program tabs are displayed with a green circle inside the tab.
![image](https://github.com/user-attachments/assets/7e1f9db1-c3c9-4753-b44d-f5db8f1392e1)

PR #1168, opened 2024-08-23
feat: clear log on execution start

### LINKED ISSUE #581: Can I view the SAS log as a file?
I struggle to find the relevant information in the SAS log as displayed in the output pane in VS Code.  I think there are actually two interlinked challenges:
First, there is limited structure available to help the user parse the log file, and what structure is available is difficult to exploit. Second, I have a physical disability and struggle with the physical act of scrolling through so much log text without other navigation methods (e.g. searching) .

I would like to be able to view the SAS log as a text file within a normal editor tab in VS code.  This would allow me to make use of all the extensions and customizations that I already have in place in my VS code environment to more easily navigate the log and find the information I need.

[Issue 159](https://github.com/sassoftware/vscode-sas-extension/issues/159) describes another approach that I think is valid but probably more difficult to implement.

I've taken a look at the [documentation on VS code tasks](https://github.com/sassoftware/vscode-sas-extension/blob/main/connect-and-run.md#run-sas-code-by-task), but that does not seem to give me a way to redirect the log output to a file.

I've also tried to address some of my navigation challenges by maximizing the output pane in VS code. This is somewhat helpful as it does fit more text on screen, but I still do not have the ability to search within the log.

**Additional context**
N/A

**Environment**
SAS version: 	20231012.1697126034654

scnwwu:
**Summary**
Hopefully will resolve #581 and #867

Added a new settings `SAS.log.clearOnExecutionStart`. When set to `true`, clear SAS Log output channel each time code execution starts. As the SAS Log is dedicated for one execution, display the executed file name on the channel for easier tracking.

User can click the `Open Output in Editor` menu item on the Output channel toolbar to open the log as a text file. The file content contains only the SAS Log for one execution. New execution will produce a new file and the previous file can be preserved.

**Testing**

1. Verify that when `SAS.log.clearOnExecutionStart` set to `false`, the behavior is not changed from previous version.
2. Verify that when `SAS.log.clearOnExecutionStart` set to `true`, previous log was cleared on next run.


smorrisj:
We've also had some internal feedback on how to best match log output with the program or file that produced it on #867. Any thoughts on ideas that I mentioned on that one?

smorrisj:
Some users are going to structure their programs to run selections / blocks of code in a given file. One potential drawback is that if a user were to set clearOnExecutionStart in that scenario, then the log would be cleared as each incremental execution happens, correct? 

What if we allocated an output channel to each program that gets run on a session? If we did that, we'd need to be sure to clean it up on session close. This approach would allow those users to incrementally run sections of code, or an entire program, and have the log output show up accordingly in that program / code document's output channel. Maybe we could also set the active output channel based on the users active document tab. Not sure what this means for users that would have multiple programs in a layout however. We would need to investigate that scenario for sure.

scnwwu:
I guess the tab layout like Studio would be preferred - log and result grouped by program. The output channels can't be split into separate tabs. And user can't manually remove a channel if we added too many channels. I'm thinking the log file (opened by `Open Output in Editor`) may work better. Even we disposed the channel, the log file remains. User can drag it into a new window with corresponding program. Switch or close whenever they want.
The only thing is that the log file is plaintext and don't have error/warning colors.

client/src/commands/run.ts
```diff
@@ -222,6 +226,9 @@ async function _runTask(
   );
   session.onSessionLogFn = appendSessionLogFn;
 
+  const fileName = codeDoc.getUri().split("/").pop().split(".")[0];
```
scottdover, line 229:
```suggestion
  const fileName = basename(codeDoc.getUri(), extname(codeDoc.getUri()));
```

scottdover: We can rely on `basename` and `extname` from `path` to handle this for us

scnwwu: Cool. Done.

scottdover (approved):

tianliang657 (approved):
The new feature is working fine in the playpen extension, no issues were found, so approving. please see the attached video for the details.
Fixed

PR #1297, opened 2024-11-26
fix: use vscode.open to open file

### LINKED ISSUE #1289: The format of the notebook file is not correct after doing renaming action
**Describe the bug**
Open VSC SAS Extension base on Main branch. Sign in with a valid Viya connection, create a SASNotebook file on SAS server, rename the file. The format of the opened file is not correct.  

**Steps to reproduce**
1. Sign in with a valid viya connection
2. Open SAS Server
3. Create a new SASNotebook file under an authorized directory
4. Rename the file

**Expected behavior**
The file is in correct format after reopening it

**Screenshots**
image

Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.12.0]

scnwwu:
**Summary**
Fix #1289
Need to use `vscode.open` to open file with correct UI instead of always opening with the default code editor.
It looks like the reason using `vscode.openWith` was to set `preview` to `false`. In my opinion we should not always set `preview` to `false`. It's possible that the file was in preview mode before closing, in which case it doesn't make sense to re-open with `preview: false`. The tab closing is noticeable to end user. We even ask user to save or not save the file before closing. So I think it's OK to re-open with preview mode, same as if user close and re-open.

**Testing**
Test case in #1289

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -301,18 +301,14 @@ class ContentDataProvider
     const oldUriToNewUriMap = [[item.vscUri, newUri]];
     const newItemIsContainer = getIsContainer(newItem);
     if (closing !== true && !newItemIsContainer) {
-      await commands.executeCommand("vscode.openWith", newUri, "default", {
-        preview: false,
-      });
+      await commands.executeCommand("vscode.open", newUri);
     }
     if (closing !== true && newItemIsContainer) {
       const urisToOpen = getPreviouslyOpenedChildItems(
         await this.getChildren(newItem),
       );
       for (const [, newUri] of urisToOpen) {
-        await commands.executeCommand("vscode.openWith", newUri, "default", {
-          preview: false,
-        });
+        await commands.executeCommand("vscode.open", newUri);
       }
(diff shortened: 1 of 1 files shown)


scottdover:
I switched to using openWith in my latest changes because renaming a folder requires closing _all_ files within the folder (for sas server, where the uris change as part of renaming). This change would effectively make it such that we close N files (where N > 1) but only re-open 1 file (unless preview mode is disabled)

scnwwu:
> I switched to using openWith in my latest changes because renaming a folder requires closing _all_ files within the folder (for sas server, where the uris change as part of renaming). This change would effectively make it such that we close N files (where N > 1) but only re-open 1 file (unless preview mode is disabled)

Hi @scottdover, I tried with this PR, it successfully re-opened all closed files. When calling `vscode.open` the second time, the first opened tab was made non-preview automatically. Would you double check?

scottdover (approved):

scottdover:
You're right 🤔 . I was pretty certain I tested re-opening multiple files after renaming the folder and it didn't work with preview mode...must've been something else causing my issue. Thanks for taking care of this :)

ShermanWang0324:
Tested the rename sas notebook file with unsaved sas code, re-render will successful as expected.
fixed

ShermanWang0324 (approved):

PR #1310, opened 2024-12-02
feat: inherit VS Code file icons in SAS Content and Server

### LINKED ISSUE #1299: Use VSCode Icons for other file types in SAS Server Files
Right now if you are looking at files on the SAS Server filesystem, i.e. Linux it will not provide icons to files that are not `.sas` files. It would be nice if other files types had icons so it was easier to identify what the file is without needing to look at the extension.

I would like to see icons that are provided from the SAS extension to be pulled from VSCode. VSCode supports a lot of different file extension icons. As well this should support alternative icons sets, but I imagine if setup correctly it wouldn't need to be managed.

There is none.

**Additional context**
Below is an example of how non `.sas` files do not have an icon.
image

**Environment**
SAS version: Viya 4

scnwwu:
**Summary**
Resolves #1299
We should not explicitly specify a general file icon so that VS Code can go with its default.
It honors user's [File Icon Themes](https://code.visualstudio.com/docs/getstarted/themes#_file-icon-themes)

**Testing**
See file icons in SAS Content and SAS Server same to VS Code Explorer view.

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -18,7 +18,6 @@ import {
   TabInputText,
   TextDocument,
   TextDocumentContentProvider,
-  ThemeIcon,
   TreeDataProvider,
   TreeDragAndDropController,
   TreeItem,
@@ -230,6 +229,7 @@ class ContentDataProvider
       iconPath: this.iconPathForItem(item),
       id: item.uid,
       label: item.name,
+      resourceUri: uri,
     };
   }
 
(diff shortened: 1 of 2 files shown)


client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -723,11 +723,6 @@ class ContentDataProvider
           icon = "folder";
           break;
       }
-    } else {
-      const extension = item.name.split(".").pop().toLowerCase();
```
scottdover, line 727: Am I understanding correctly? By removing this case, we return undefined, and `package.json` basically dictates the icon for `.sas` extension files, right? (looking at https://github.com/sassoftware/vscode-sas-extension/blob/95794c63f4cd8ffa23f3dd14583a71f1f867ce77/package.json#L1110-L1113)

If this is accurate, this LGTM

scnwwu: yes correct

tianliang657 (approved):
This issue is fixed in the playpen sas extension, no other issues were found, so approving.
image

PR #1326, opened 2024-12-16
feat: display macro name in outline

### LINKED ISSUE #1323: It would be useful if the Outline pane showed the name of the macro associated with the 'MACRO' items in the tree.
When using the SAS Extension for VS Code to view SAS code containing many macros, the `Outline` pane shows many items in the tree with the label `%MACRO`. Clicking on one of those items takes you to the associated `%MACRO` statement within the code, but we have no indication in the pane of exactly which macro is represented by which entry.
image

It would be extremely useful if we could also see the name of the associated entry on each item.
image

scnwwu:
**Summary**
Resolves #1323

**Testing**
See macro name and the first dataset name in data step displayed in Outline pane.

### DIFF
--- a/server/src/sas/LanguageServiceProvider.ts
+++ b/server/src/sas/LanguageServiceProvider.ts
@@ -170,7 +170,7 @@ export class LanguageServiceProvider {
       end: { line: block.endFoldingLine, character: block.endFoldingCol },
     };
     const docSymbol: DocumentSymbol = {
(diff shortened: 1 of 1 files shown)


scnjdl (approved):

ShermanWang0324:
Tested the fix is expected, %MACRO has show its name on OUTLINE panel.

ShermanWang0324 (approved):

PR #1389, opened 2025-02-06
fix: bubble server error up

### LINKED ISSUE #1367: Better Error Display for Rest Session Errors
**Describe the bug**
We should show better error messages when Compute jobs return errors, specifically on Session creation. Currently, only the "top level" error message is being show to the user, which is not helpful to understand why requests might have failed.

**Steps to reproduce**

1. Create an autoexec file that has an include statement. Make sure the include file exists:
```
%include cowboy.sas
```
2. Specify the autoexec in the connection profile autoxec config:
```
        "viyavsc": {
            "connectionType": "rest",
            "endpoint": <omitted>,
           
            "autoExec": [
                {
                    "type": "file",
                    "filePath": <absolute path to autoexec sas file>
                }
            ]
        
       }
```
3. Create a sas file to run, for example:
```
proc sgplot data=sashelp.class;
    histogram height;
    density height;
    density height / type=kernel;
run;
```
4. Run the test file in step 3. Observe the error reported as "Request failed with status code 400".
5. If you observe the actual server log, the real error message is noted:
```
[00000019] << {"details":["%INCLUDE statements can only be set by an administrator, and not directly in autoExecLines of the session request."],"errorCode":5101,"errors":[],"httpStatusCode":400,"id":"","links":[],"message":"%INCLUDE statements are not allowed in the autoExecLines of the session request.","remediation":"Contact your system administrator for changes to the autoexec content for the context.","version":2} 
```

**Expected behavior**
Show the contents of the rest error representation if we receive a response payload that looks like an Error Representation from compute. We have existing examples of how this is done in the run commands. See run.ts.

**Additional context**
Reported by a WFL user. This issue will affect viya and Workbench connection types

scnwwu:
**Summary**
Fix #1367
Remove the try block so that error will bubble up to the existing error handling.

**Testing**
Tested with Workbench and it can show error details same to Viya

### DIFF
--- a/client/src/connection/rest/server.ts
+++ b/client/src/connection/rest/server.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import { l10n } from "vscode";
 
-import axios, { AxiosResponse } from "axios";
+import { AxiosResponse } from "axios";
 
 import { Link, Server, ServersApi, SessionRequest } from "./api/compute";
 import { BaseCompute, Compute, getApiConfig, stateOptions } from "./common";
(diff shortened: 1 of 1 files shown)


scottdover (approved):

Zhirong2022:
Check the issue in workbench, it can show error details the same as Viya. No regression issues in VSC SAS Extension.
image

Zhirong2022 (approved):

Issue #1212, opened 2024-09-17
[SAS Notebook] Support Refresh Kernel action in SAS Session


fabiocerutisas:
**Is your feature request related to a problem? Please describe.**
When coding in .sasnb notebooks in SAS Viya Workbench, it can happen that a cell takes too long to execute and it can't be stopped with the interrupt button. Specifically, after pressing the interrupt button, there is a "cancelling SAS jobs" message that shows up but the cell does not stop from running. 
This is something that also happens in Python programming: when stopping a cell that is taking too long, often times the Interrupt button is not enough to stop the process and the .ipynb notebook forces you to refresh the kernel to resume coding. 

**Describe the solution you'd like**
It would be nice to have a Refresh Kernel button for sas notebooks similar to the one already available for .ipynb notebooks. The button should clear the session from all on-going processes and reset all the macro variables/functions that have been defined. Additionally, when trying to stop a cell from running is taking too long, a message stating to refresh the kernel should pop-up (as it is already the case for .ipynb notebooks).

**Describe alternatives you've considered**
Currently, when a SAS cell is taking too long to execute and is not stopping after pressing the Interrupt button, I need to shut down my entire workbench instance to make sure that the SAS Code stops from running.

**Additional context**
N/A

**Environment**
SAS Viya Workbench

scnwwu:
Does `Close Session` (hovering the profile at the bottom) work in this case?

fabiocerutisas:
I have tried but the code still runs, it is not killed by the Close Session button. 
Once I click on the Close Session button I correctly see the "The SAS session has closed" banner in the bottom right part of screen yet, as seen from the picture below, even after 10+ minutes the cell is still executing
![image](https://github.com/user-attachments/assets/e5da8f48-4216-49f5-8438-7f7d76d3dc7d)

scnwwu:
With #1390 shipped in v1.13.1, now that the `Close Session` command should reset everything correctly.
Will re-consider if a `Refresh Kernel` command is needed when #403 implemented.

Nurquhart:
@fabiocerutisas - I have been revisiting this bug.  With @scnwwu changes, **Close Session** should reset the session completely and allow you to re-run your code block. Does this accomplish what you were wanting or is there anything else you suggest we add?

Nurquhart:
[...]

PR #1390, opened 2025-02-10
fix: reject run promise when close session

scnwwu:
**Summary**
Fix the problem described [here](https://github.com/sassoftware/vscode-sas-extension/issues/1212#issuecomment-2360386822)
Need to reject the current run promise when close session, otherwise it will never resolve or reject and user can't run again.

**Testing**
Close session in a long running, user can run again.

### DIFF
--- a/client/src/components/notebook/Controller.ts
+++ b/client/src/components/notebook/Controller.ts
@@ -98,10 +98,12 @@ export class NotebookController {
         ]),
       ]);
       execution.end(false, Date.now());
+      if (!this._interrupted) {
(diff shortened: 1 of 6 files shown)


smorrisj (approved):

scottdover (approved):

Zhirong2022:
Suppose it has a code cell in SASNotebook which takes a long time to run.  The user can click 'Close Session' to stop the running directly. 
image

Zhirong2022 (approved):

PR #1409, opened 2025-02-28
feat: load interop dlls for iom/com, supply appname

scottdover:
**Summary**
This change updates both IOM & COM connections to load SAS interop dlls so that we can specify an application name when starting a connection. We attempt to load the DLLs in the following order...
 - A user specified location (specified in a new profile option, `interopLibraryFolderPath`). If this location can't be found, we proceed checking other locations
 - If no `interopLibraryFolderPath` is specified, we attempt to lookup the path in the registry
 - Then, in the Integration Technologies folder

If all of these fail, we provide the user with an error message.

**Testing**
 - [x] Tested each different way of loading interop dlls, and made sure connections worked as expected


clangsmith:
I pulled these changes, connected to a V9 server (uitsrv05) via IOM, checked the Object Spawner log on that server (ex. C:\SAS\Config\Lev1\ObjectSpawner\Logs\ObjectSpawner_<latest>.log), and see the APPNAME property value is now being passed for connections from VS Code (looks good!):
 
2025-03-03T16:14:24,205 INFO  [00000052] :sasdemo@UITSRV05 - New client connection (23) accepted from server port 8591 for user sasdemo@UITSRV05. Encryption level is Credentials using encryption algorithm SASPROPRIETARY. Peer IP address and port are [::ffff:10.26.13.150]:27666 for **APPNAME=SAS Extension for Visual Studio Code**.

For comparison, here is one in the same Object Spawner log for a connection made from EG:

2025-03-03T16:32:09,736 INFO  [00000085] :cassmi@CARYNT - New client connection (25) accepted from server port 8591 for user cassmi@CARYNT. Encryption level is Credentials using encryption algorithm SASProprietary. Peer IP address and port are [::ffff:10.26.13.150]:11181 for **APPNAME=SAS Enterprise Guide**.

clangsmith:
It is worth noting that the IApplicationName interface was implemented (ApplicationName property on the ObjectFactory) in 9.3M2. A hotfix was also created for 9.2M3 to back port support to that version. Therefore, with these changes, it means a user will likely get an error if they attempt to connect to a server prior to 9.3 (or 9.2 without the hotfix). See https://support.sas.com/kb/40/596.html for details. (Internal DEFECTS references: S0856451, S0844154, and S0828940.)

We probably haven't tested the extension with anything prior to 9.4, but I suspect it would work fine (like EG and AMO do). If we want to support anything prior to 9.3, we'd have to skip this ApplicationName call. EG created a config property to skip the call. Again though, for now, I think it is fine to say the extension doesn't support SAS versions prior to 9.3. Just noting.

client/src/connection/itc/script.ts
```diff
@@ -12,6 +12,42 @@ export const scriptContent = `
 class SASRunner{
   [System.__ComObject] $objSAS
 
+  SASRunner([string]$interopPath) {
+    try {
+      $interopDir = $this.GetInteropDirectory($interopPath)
+      Add-Type -Path "$interopDir\\SASInterop.dll"
+      Add-Type -Path "$interopDir\\SASOManInterop.dll"
+    } catch {
+      Write-Error "${ERROR_START_TAG}Init error${ERROR_END_TAG}"
```
scnwwu, line 21: I see that other catch blocks print error details, do you want to do the same?
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/connection/itc/script.ts#L106
```suggestion
      Write-Error "${ERROR_START_TAG}Init error: $_${ERROR_END_TAG}"
```

client/src/connection/itc/index.ts
```diff
@@ -368,6 +369,10 @@ export class ITCSession extends Session {
       return l10n.t("This platform does not support this connection type.");
     }
 
+    if (/Init error/.test(msg)) {
```
scnwwu, line 372: nit: since it's specific to loading libraries, how about a more specific name? something like
```suggestion
    if (/LoadingInterop error/.test(msg)) {
```

scottdover: Addressed in https://github.com/sassoftware/vscode-sas-extension/pull/1409/commits/a2b4f0e8bd572d3067f6a9b2c1023d45aa88fde4

client/src/connection/itc/script.ts
scottdover: Addressed in https://github.com/sassoftware/vscode-sas-extension/pull/1409/commits/a2b4f0e8bd572d3067f6a9b2c1023d45aa88fde4

website/docs/Configurations/Profiles/additional.md
```diff
@@ -28,7 +28,8 @@ SAS system options can be set for each connection profile. Changes to the SAS sy
         "username": "username",
         "port": 8591,
         "sasOptions": ["NONEWS", "ECHOAUTO", "PAGESIZE=MAX"],
-        "ConnectionType": "iom"
+        "ConnectionType": "iom",
+        "interopLibraryFolderPath": "C:\\Program Files\\SASHome\\x86\\Integration Technologies"
```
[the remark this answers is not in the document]
scottdover: Ah, I see. I updated the docs in https://github.com/sassoftware/vscode-sas-extension/pull/1409/commits/efe42d25511596a0bb58cc6e0479268ffde6e5c7

client/src/connection/itc/index.ts
```diff
@@ -333,7 +334,11 @@ export class ITCSession extends Session {
     );
 
     // If we encountered an error in setup, we need to go through everything again
-    const fatalErrors = [/Setup error/, /powershell\.exe/];
+    const fatalErrors = [
+      /Setup error/,
```
smorrisj, line 338: Not related to this specific change. It looks like we're writing our error strings always as English. Does it make any sense to localize these strings like we do with others in the extension proper?

smorrisj:
Should the interop path be defined as an optional property for the profile types? E.g. on COMProfile.

scnwwu:
> Should the interop path be defined as an optional property for the profile types? E.g. on COMProfile. I suppose it might not make a ton of sense if we don't intend to prompt the user for input on profile creation. I noticed that we have declared other types however for fields that are also not prompted for. Thoughts?
> 
> ```ts
> export interface COMProfile extends BaseProfile {
>   connectionType: ConnectionType.COM;
>   host: string;
>   interopLibraryFolderPath?: string;
> }
> ```

I think it's good to NOT prompt user for this on profile creation, as it's only for special cases.

Yeah, I see some fields are declared in profile type but not be prompted or validated. I also see some fields in connection config type but not in profile type. I guess perhaps we can just keep one series of types instead of two (out of the scope of this PR).

scnwwu (approved):
Let me just give my vote. For Joe's last comment I'm good with either including it in the profile type or not.

ShermanWang0324:
@scottdover  I try to test this order
**1. The path specified in `interopLibraryFolderPath`. This should be an absolute path to the folder containing the files listed above.
2. If `interopLibraryFolderPath` is empty or invalid, the extension will attempt to resolve the path from Windows registry
3. If steps 1 and 2 fail, the extension will attempt to load libraries from the default Integration Technologies Client folder (`C:\Program Files\SASHome\x86\Integration Technologies`)**

As the screenshot, I made the "`interopLibraryFolderPath" as empty to failed the step#1
then change the registry to make the 'path' point to another location to failed the step#2
all DLL files still in default ".../Integration Technologies", then try to connect this profile, but still got error "Unable to load require libraries". -> Do this expected? 
image

smorrisj:
> > Should the interop path be defined as an optional property for the profile types? E.g. on COMProfile. I suppose it might not make a ton of sense if we don't intend to prompt the user for input on profile creation. I noticed that we have declared other types however for fields that are also not prompted for. Thoughts?
> > ```ts
> > export interface COMProfile extends BaseProfile {
> >   connectionType: ConnectionType.COM;
> >   host: string;
> >   interopLibraryFolderPath?: string;
> > }
> > ```
> 
> I think it's good to NOT prompt user for this on profile creation, as it's only for special cases.
> 
> Yeah, I see some fields are declared in profile type but not be prompted or validated. I also see some fields in connection config type but not in profile type. I guess perhaps we can just keep one series of types instead of two (out of the scope of this PR).

Seems reasonable. Should we make a separate issue to tackle this? I agree that it would be good to find a way to operate against one set of config types, whether it's in the connection provider layer or in profiles.

scottdover:
> @scottdover I try to test this order **1. The path specified in `interopLibraryFolderPath`. This should be an absolute path to the folder containing the files listed above. 2. If `interopLibraryFolderPath` is empty or invalid, the extension will attempt to resolve the path from Windows registry 3. If steps 1 and 2 fail, the extension will attempt to load libraries from the default Integration Technologies Client folder (`C:\Program Files\SASHome\x86\Integration Technologies`)**
> 
> As the screenshot, I made the "`interopLibraryFolderPath" as empty to failed the step#1 then change the registry to make the 'path' point to another location to failed the step#2 all DLL files still in default ".../Integration Technologies", then try to connect this profile, but still got error "Unable to load require libraries". -> Does this expected?
> 
> image
> 
> [updated0321]: @scottdover, my testing is on hold for this, please let me know your comments when you have time. Thanks:)

Hey @ShermanWang0324. This should now be resolved. My guess is that you changed the path, but changed it to a _valid_ folder. In that instance, I could see why there would be a failure. I've updated the code to account for that.

ShermanWang0324:
Test steps and results:
---------------------------------------------
1. Move all *Interop.dll to a new location
2. Connect with the profile with no `interopLibraryFolderPath`, report error "Unable to load required libraries." - correct
3. Connect with the profile with `interopLibraryFolderPath` point to new location, connect success. - Order#1 correct
4. Rename the registry 'path' to new location, then connet with the profile with `interopLibraryFolderPath` is empty, connect success. - Order#2 correct
5. Copy all *Interop.dll back to default location, then connect with the profile with `interopLibraryFolderPath` is empty, connect success. - Order#3 correct

ShermanWang0324 (approved):

PR #1417, opened 2025-03-06
feat: export sasnb to HTML

scnwwu:
**Summary**
Provide the ability to export SAS Notebook to HTML file.

_Notes_
- Extensions do not have access to the VS Code DOM, we're not able to get what rendered in VS Code. Therefore the exported HTML may not look same to what you see in VS Code. Same thing happens for Jupyter extension.
- The exported HTML will be in Light or Dark theme depending on your current VS Code theme kind.
- By default it doesn't include SAS log into the exported HTML file. To include log, check the `SAS.notebook.export.includeLog` setting.

**Testing**
Open a `.sasnb` file, click `...` from the top toolbar and select `Export`, specify an HTML file name.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -12,6 +12,8 @@
         "ag-grid-community": "^33.1.1",
         "ag-grid-react": "^33.1.1",
         "axios": "^1.8.2",
+        "highlight.js": "^11.11.1",
+        "marked": "^15.0.7",
         "media-typer": "^1.1.0",
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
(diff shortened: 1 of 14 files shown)


scnjdl (approved):

Zhirong2022:
The theme style of the output(SAS HTML render) would be mixed after specify a new ODS HTML5 results style. 
There are three code cells
1.Suppose the code cells 'ODS HTML5 results style' is 'HTMLBlue' initially
2.Change 'ODS THML5 results style' as 'HighContrast'
3.Run the second cell only
4.Export SASNotebook as .html file
themeMixed

scnwwu:
> The theme style of the output(SAS HTML render) would be mixed after specify a new ODS HTML5 results style. There are three code cells 1.Suppose the code cells 'ODS HTML5 results style' is 'HTMLBlue' initially 2.Change 'ODS THML5 results style' as 'HighContrast' 3.Run the second cell only 4.Export SASNotebook as .html file

I can't think of a use case why user want different themes for different cells in one report. Let's NOFIX for now. We can re-visit when user request.

Zhirong2022:
Export SASNotebook file to HTML.

Zhirong2022 (approved):

PR #1437, opened 2025-04-03
feat: provide API to get session id

scnwwu:
**Summary**
Provide extension level API `getSessionId` for other extensions to get Viya session id.

**Testing**
The added API can be called from another extension

### DIFF
--- a/client/src/node/extension.ts
+++ b/client/src/node/extension.ts
@@ -36,7 +36,7 @@ import { SASAuthProvider } from "../components/AuthProvider";
 import { installCAs } from "../components/CAHelper";
 import ContentNavigator from "../components/ContentNavigator";
 import { ContentSourceType } from "../components/ContentNavigator/types";
-import { setContext } from "../components/ExtensionContext";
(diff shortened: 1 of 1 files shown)


clangsmith:
I have some concerns with this request. We had similar requests and problems surfaced as a result in SAS Studio. Other components used by SAS Studio wanted Studio to provide the compute session ID, so they could then make any desired requests to the same compute session. The problem is that the compute session often services requests synchronously and having two separate/independent consumers using the same compute session opens the door for collisions/contention. To address this issue in SAS Studio, rather than providing the compute session ID (thus unfettered access to the same compute session), we opted to surface APIs such that the components could request specific information from the primary consumer/owner of the compute session (SAS Studio in this case), and the primary consumer/owner could still manage the requests and then return that information to the secondary consumers. This is the approach I think is probably best here as well...  the SAS Extension for VS Code is the primary owner/consumer of the Compute session, and will return specific information (via APIs) to any secondary consumers when requested, thus only one extension that directly interacts (thus manages contention) with the Compute session.

PR #1452, opened 2025-04-21
fix: reduce unnecessary rest calls

### LINKED ISSUE #1447: Every code run against Viya has a long connect step

With our Viya instance, in VS Code extension, _every time_ one needs to execute a bit of code [either select-a-block and click the running man, or each cell of a notebook], it takes a moment to do "Connecting to SAS session..."

Sometimes, that initial connection takes long enough to be a friction point. For example, when someone is going across a relatively slow VPN link, the initial connect is slow.

Is it possible to hold the HTTP session open after a command completes, and only reconnect if it drops for some reason [eg inactivity timeout]? That way it's immediately available to send the REST queries across it.

It's not obvious to me all the steps that are happening during that connect, so I have no suggestion on alternatives.

### Additional context

It's difficult to tell if the long login is truly just a slow network issue, or Viya taking a while to instantiate the necessities on each new HTTP connection, but either way it's frustrating.

### SAS version

Viya 4

scnwwu:
**Summary**
Fix #1447
The `job.getLogStream()` call is already polling job state and resolve on job complete. Remove unnecessary get state calls.

**Testing**
See execution is a little faster.

### DIFF
--- a/client/src/connection/rest/index.ts
+++ b/client/src/connection/rest/index.ts
@@ -142,21 +142,11 @@ class RestSession extends Session {
     if (!this._computeSession?.sessionId) {
       throw new Error();
     }
-
     //Get the job
     const job = await this._computeSession.execute({ code: [code] });
-    let state = await job.getState();
-
-    const retLog = this.printJobLog(job);
-
-    //Wait until the job is complete
-    do {
-      state = await job.getState({ onChange: true, wait: 2 });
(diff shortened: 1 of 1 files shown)


scottdover (approved):

tianliang657 (approved):
This issue is fixed in the playpen sas extension, please see the following screenshot for the details, no other issues were found, so approving.
image

Issue #1439, opened 2025-04-04
Table Viewer - Lazy Loading/Pagination


danila-grobov:
We work with moderately large datasets ~6m rows.

Currently I can't even view the dataset, it just shows up as empty.

I suspect that this is due to it loading the whole thing at once.

I would like to contribute, and try to solve this issue.

I like the implementation in Enterprise Guide so I will try to implement something similar, where you scroll and it loads as you go.

danila-grobov:
Forgot to mention, I'm talking about SAS 9.4 here. The issue is probably not very relevant for the modern SAS.

I took a closer look at the issue, and it's not related to pagination. The pagination (lazy loading) is already implemented, it's just very slow for datasets that have a lot of columns. It takes like 30 seconds to get 100 rows and then it's 30 seconds each time it loads a bit more.

I believe that this is due to the way data is being transported. The selected rows in the dataset are just printed out in the log and that is being interpreted on the client. And since pretty print is being used, that's a lot of rows.

If I turn off the pretty print it speeds up, it takes around 5 seconds, which is a lot better. However there is another problem that comes with that, now we are hitting the limit of line length in SAS logs. Since it's not pretty printed, everything is just dumped into a single line. This results in the data being cut off.

All of this seems like trying to solve the issue with the wrong tools. There is the SAS IOM data provider which is specifically made for this kind of data retrieval. There is even an adapter for local files, which would make it possible to read sas7bdat files locally. I believe it just comes bundled with SAS client tools? Not sure what's then policy with dependencies like that in this extension.

I will try to work on a POC and create a PR to migrate to that instead.

What do you guys think?

scnwwu:
Looks similar to #1233

I've tried the data provider approach but it crashes Powershell.
https://github.com/sassoftware/vscode-sas-extension/issues/1132#issuecomment-2406624304

danila-grobov:
Oh, cool, thank you for the context!

I will keep in mind the crashing issue. Haven't encountered it yet.

rwat1:
This might be the nail in the coffin if this isn't fixed. I will tell my wife to also learn R and quit SAS for good.

DmitryMK:
[...]

PR #1388, opened 2025-02-03
feat: implement sas file system for IOM/COM connections

scottdover:
**Summary**
This adds sas file system support for sas9 file system for IOM/COM connection types. Additionally, it includes a small refactor to the way interop libraries are loaded. Loading them at SASRunner construction occurs too late, and doesn't provide the type-checking necessary for file service.

**Testing**
 - [x] Test entering incorrect password **(note: an issue remains here that is present on main. You can't attempt to re-log in if you've entered a wrong password unless you reload window)**
 - [x] File/folder creation
    - [x] Create file/folder w/ context menu
    - [x] Create file/folder by upload
    - [x] Create file/folder by drag & drop (create multiple files)
 - [x] File/folder deletion
    - [x] Test file deletion with context menu
    - [x] Test multi-file deletion with context menu
 - [x] File/folder updates
    - [x] Test updating file/folder name
    - [x] Test updating file contents
    - [x] Test moving file/folder (multiple files/folders)
 - [x] Test downloading files/folders
 - [x] Make sure refresh works as expected
    - [x] Make sure connections are automatically refreshed after they become stale
 - [x] Make sure we're displaying all files/folders for sas server and that items are sorted by type (directories before files), then alphabetically
 - [x] Make sure we can collapse all folders
 - [x] Cases to test from previous rest implementation
    - [x] Create folder with special character, like `/`
    - [x] Create folder in same spot as folder with same name
    - [x] Create a .txt file named as `NewFile+!@$%^&*.txt` **(NOTE: File is not created in this instance. sas9's file service rejects this file name)**
    - [x] Check multi-selection behavior to make sure things work as expected
    - [x] Rename file (and folder) as same name as existing file
    - [x]  Create file in folder, keep file open, delete folder (make sure file goes away)
    - [x] Create a file named `!@$%^&*(.txt`, attempt to download it  **(NOTE: File is not created in this instance. sas9's file service rejects this file name)**
    - [x] Try dropping a file into a folder where there's already a file sharing the same name
    - [x] Try renaming an open file

Issue #1132, opened 2024-08-08
SAS.viewTable is not showing data with correct encoding


henrik-forsell:
**Describe the bug**
Create a data set with that contains national characters, e.g. ÅÄÖ. Open the data set and it will not show those chars correctly.

**Steps to reproduce**
*Create a data set with national chars;
data work.myencoding;
    check_encoding="aa å, ae ä, oe ö";
run;

*Print values to the log, it will show it correctly;
data _null_;
    set work.myencoding;
    put check_encoding;
run;

*Print the data set to output, it will show it correctly;
proc print data=work.myencoding;
run;  

Open the data set from the library and it will be showing the wrong characters.

**Expected behavior**
The view table should show the national characters correctly.

**Screenshots**
![image](https://github.com/user-attachments/assets/aa78dae0-6849-4db2-baa5-6ba5de8f10de)

**Environment (please complete the following information):**
Client OS: Windows 11
Extension version: v1.10.1
SAS version: 9.4

scnwwu:
Currently the extension is using PROC JSON to dump the table content to JSON to be easily consumed by JS. It appears that PROC JSON can't handle the encoding well. Run below code will produce wrong characters into log.
```sas
filename out temp;
proc json nokeys out=out pretty; export work.myencoding; run;
data _null_; infile out; input; put _infile_; run;
```
I think we should take the OLE DB Provider (ADO) approach like EG.

henrik-forsell:
As you guessed, there are odd chars in the log:
![image](https://github.com/user-attachments/assets/a155bc9d-fc15-45f8-a6ab-e6f464308de6)

scnwwu:
Writing down development details. I tried ADO approach with below code

```ps1
$objFactory = New-Object -ComObject SASObjectManager.ObjectFactoryMulti2
$objServerDef = New-Object -ComObject SASObjectManager.ServerDef
$objServerDef.MachineDNSName = "localhost"
$objServerDef.Protocol = 0

# Class Identifier for SAS Workspace
$objServerDef.ClassIdentifier = "440196d4-90f0-11d0-9f41-00a024bb830c"

# create and connect to the SAS session
$objSAS = $objFactory.CreateObjectByServer(
    "test",
    $true,
    $objServerDef, # built server definition
    "",
    ""
)

# Add the new workspace object to the ObjectKeeper
$objOK = New-Object -ComObject SASObjectManager.ObjectKeeper
[...]

PR #1466, opened 2025-05-03
Table Viewer optimization in ITC

danila-grobov:
**Summary**
Currently the table viewer is not usable for ITC connection, especially if you have a lot of columns, it takes like 20 seconds if not more to load. Related to the issue https://github.com/sassoftware/vscode-sas-extension/issues/1439

I've mostly fixed that by making use of ADO DB connector that SAS includes with interop library. I borrowed a bunch of setup for this from https://github.com/sassoftware/vscode-sas-extension/pull/1388,  so changes in CodeRunner for example are just a copy paste from that PR.

I was made aware of the crashing issue that others have faced with using this connector, however I didn't envounter it and I've tried oppening up a bunch of different tables.

I've also added a column type detection function, which I ripped from SAS Studio. So it interprets the formats in the same way and now we can see all of the beautiful icons.

image

Now it takes around 7 seconds including column loading, compared to 5 seconds that it takes in Enterprise Guide for the same table.

I did decide to stop at this point just to see what you guys think and release this into the wild. If this goes well, we can look into improving this further. There are at least three areas of pottential improvement:

1. Adding loading indicators, which I think would make it seem a bit more responsive.
2. Fetching metadata of the table and the data at the same time. Which I think would require a bunch of changes in ITC adapter architecture,  because currently it is setup to allow only a single thing happening at the same time.
3. Reducing the amount of  initial columns. This trick is used in all other SAS clients, so might as well.

**Testing**

I don't have access to locally running SAS, so all of this has been tested on a remote IOM connection to a server running on RHEL.

- [x] Opening a table with contents that are more than 100 row limit
- [x] Opening a table with contents that are less than 100 row limit
- [x] Opening an empty table
- [x] Downloading a table
- [x] Scrolling through a large table

### DIFF
--- a/client/src/connection/itc/script.ts
+++ b/client/src/connection/itc/script.ts
@@ -263,6 +263,12 @@ class SASRunner{
 
     $records = [List[List[object]]]::new()
     $fields = $objRecordSet.Fields.Count
+
+    if ($objRecordSet.EOF) {
+      Write-Host '{"rows": [], "count": 0}'
+      return
+    }
+
     $objRecordSet.AbsolutePosition = $start + 1
 
     for ($j = 0; $j -lt $limit -and $objRecordSet.EOF -eq $False; $j++) {
@@ -287,6 +293,39 @@ class SASRunner{
     $result | Add-Member -MemberType NoteProperty -Name "count" -Value $count
 
     Write-Host $($result | ConvertTo-Json -Depth 10)
+  }
+
+  [void]GetColumns([string]$libname, [string]$memname) {
+    $objRecordSet = New-Object -comobject ADODB.Recordset
+    $objRecordSet.ActiveConnection = $this.dataConnection
+    $query = @"
+      select name, type, format 
+      from sashelp.vcolumn 
+      where libname='$libname' and memname='$memname';
+"@
+    $objRecordSet.Open(
+      $query,
+      [System.Reflection.Missing]::Value, # Use the active connection
+      2, # adOpenDynamic
+      1, # adLockReadOnly
+      1  # adCmdTableDirect
+    )
+    
+    $rows = $objRecordSet.GetRows()
+
+    $objRecordSet.Close()
+
+    $parsedRows = @()
+    for ($i = 0; $i -lt $rows.GetLength(1); $i++) {
+      $parsedRow = [PSCustomObject]@{
+        index = $i + 1
+        name  = $rows[0, $i]
+        type  = $rows[1, $i]
+        format = $rows[2, $i]
+      }
+      $parsedRows += $parsedRow
+    }
+    Write-Host $($parsedRows | ConvertTo-Json -Depth 10)
   }
 }
 `;
--- a/client/src/connection/itc/script.ts
+++ b/client/src/connection/itc/script.ts
@@ -8,45 +8,49 @@ import {
 } from "./const";
 import { LineCodes } from "./types";
 
-export const scriptContent = `
+type ScriptProperties = {
+  interopLibraryFolderPath?: string;
+};
+
+export const getScript = ({
+  interopLibraryFolderPath = "",
+}: ScriptProperties) => `
+function GetInteropDirectory {
+  # try to load from user specified path first
+  if ("${interopLibraryFolderPath}") {
+    if (Test-Path -Path "${interopLibraryFolderPath}\\SASInterop.dll") {
+      return "${interopLibraryFolderPath}"
+    }
+  }
+
+  # try to load path from registry
+  try {
+    $pathFromRegistry = (Get-ItemProperty -ErrorAction Stop -Path "HKLM:\\SOFTWARE\\WOW6432Node\\SAS Institute Inc.\\Common Data\\Shared Files\\Integration Technologies").Path
+    if (Test-Path -Path "$pathFromRegistry\\SASInterop.dll") {
+      return $pathFromRegistry
+    }
+  } catch {
+  }
+
+  # try to load path from integration technologies
+  $itcPath = "C:\\Program Files\\SASHome\\x86\\Integration Technologies"
+  if (Test-Path -Path "$itcPath\\SASInterop.dll") {
+    return $itcPath
+  }
+
+  return ""
+}
+
+try {
+  $interopDir = GetInteropDirectory
+  Add-Type -Path "$interopDir\\SASInterop.dll"
+  Add-Type -Path "$interopDir\\SASOManInterop.dll"
+} catch {
+  Write-Error "${ERROR_START_TAG}LoadingInterop error: $_${ERROR_END_TAG}"
+}
+
 class SASRunner{
   [System.__ComObject] $objSAS
-
-  SASRunner([string]$interopPath) {
-    try {
-      $interopDir = $this.GetInteropDirectory($interopPath)
-      Add-Type -Path "$interopDir\\SASInterop.dll"
-      Add-Type -Path "$interopDir\\SASOManInterop.dll"
-    } catch {
-      Write-Error "${ERROR_START_TAG}LoadingInterop error: $_${ERROR_END_TAG}"
-    }
-  }
-
-  [string] GetInteropDirectory([string]$defaultInteropPath) {
-    # try to load from user specified path first
-    if ($defaultInteropPath) {
-      if (Test-Path -Path "$defaultInteropPath\\SASInterop.dll") {
-        return $defaultInteropPath
-      }
-    }
-
-    # try to load path from registry
-    try {
-      $pathFromRegistry = (Get-ItemProperty -ErrorAction Stop -Path "HKLM:\\SOFTWARE\\WOW6432Node\\SAS Institute Inc.\\Common Data\\Shared Files\\Integration Technologies").Path
-      if (Test-Path -Path "$pathFromRegistry\\SASInterop.dll") {
-        return $pathFromRegistry
-      }
-    } catch {
-    }
-
-    # try to load path from integration technologies
-    $itcPath = "C:\\Program Files\\SASHome\\x86\\Integration Technologies"
-    if (Test-Path -Path "$itcPath\\SASInterop.dll") {
-      return $itcPath
-    }
-
-    return ""
-  }
 
   [void]ResolveSystemVars(){
     try {
--- a/client/src/connection/itc/script.ts
+++ b/client/src/connection/itc/script.ts
@@ -15,6 +15,7 @@ type ScriptProperties = {
 export const getScript = ({
   interopLibraryFolderPath = "",
 }: ScriptProperties) => `
+using namespace System.Collections.Generic
 function GetInteropDirectory {
   # try to load from user specified path first
   if ("${interopLibraryFolderPath}") {
@@ -51,6 +52,8 @@ try {
 
 class SASRunner{
   [System.__ComObject] $objSAS
+  [System.__ComObject] $objKeeper
+  [System.__ComObject] $dataConnection
 
   [void]ResolveSystemVars(){
     try {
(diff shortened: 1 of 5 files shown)


scnwwu (commented):
Thank you @danila-grobov, overall looks great!

So it appears that `512 # adCmdTableDirect` doesn't crash. I used general command `select * from` in my previous trial https://github.com/sassoftware/vscode-sas-extension/issues/1132#issuecomment-2406624304 and it will crash.

client/src/connection/itc/script.ts
```diff
@@ -224,5 +247,86 @@ class SASRunner{
...
+      select name, type, format 
+      from sashelp.vcolumn 
+      where libname='$libname' and memname='$memname';
+"@
+    $objRecordSet.Open(
+      $query,
+      [System.Reflection.Missing]::Value, # Use the active connection
+      2, # adOpenDynamic
+      1, # adLockReadOnly
+      1  # adCmdTableDirect
```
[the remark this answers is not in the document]
danila-grobov: fixed in 2ab79996a930d512b7d72f74b79699e804b9fec3

client/src/connection/itc/script.ts
```diff
@@ -224,5 +247,86 @@ class SASRunner{
...
+      $records.Add($cell)
+      $objRecordSet.MoveNext()
+    }
+    $objRecordSet.Close()
+
+    $objRecordSet.Open(
+      "SELECT COUNT(1) FROM $tableName", 
+      $this.dataConnection, 3, 1, 1
+    ) # adOpenStatic, adLockReadOnly, adCmdText
+    $count = $objRecordSet.Fields.Item(0).Value
```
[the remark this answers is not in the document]
danila-grobov: I've tried this approach and it doesn't work for some reason, it just hangs forever doing nothing. It might be that my SAS deployment is not configured correctly, or just a weird ADODB implementation from SAS. Does it work for you?

client/src/connection/itc/script.ts
```diff
@@ -85,6 +92,20 @@ class SASRunner{
...
+        $this.dataConnection = New-Object -comobject ADODB.Connection
+        $this.dataConnection.Provider = "sas.IOMProvider"
+        $this.dataConnection.Properties("SAS Workspace ID") = $this.objSAS.UniqueIdentifier 
+        $this.dataConnection.Properties("Data Source") = "Data Source Name"
+        $this.dataConnection.Properties("SAS Port") = $port
+        $this.dataConnection.Properties("SAS Machine DNS Name") = $profileHost
+        $this.dataConnection.Properties("SAS Protocol") = $protocol
+        $this.dataConnection.Properties("User ID") = $username
+        $this.dataConnection.Properties("Password") = $password
+        $this.dataConnection.Open()
```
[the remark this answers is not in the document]
danila-grobov: Fixed in e78ff546b0fc4226222df2e31f39aacdfbb9d99a

client/src/connection/itc/script.ts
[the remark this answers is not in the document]
scnwwu: Yes it works for me.
Maybe you can open a PowerShell window and write those script to see if any errors or crashes.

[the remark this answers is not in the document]
danila-grobov: I think I figured it out. It works for me too, if I open a small dataset, 10 records.

But if I open something larger, a couple million records, then it hangs. I suspect it has to download the whole thing to get the "RecordCount".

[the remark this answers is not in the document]
scnwwu: OK. Makes sense.

danila-grobov:
@scnwwu Fixed the tests: cd0ef993c6586f657e833c5747fa27fb36b1351a

It got a bit messy with mocking of the uuid 😅

danila-grobov (commented):
@scnwwu Third time's the charm, I promise the tests will pass now.

client/src/connection/itc/script.ts
```diff
@@ -224,5 +243,86 @@ class SASRunner{
...
+      for ($i = 0; $i -lt $fields; $i++) {
+        $cell.Add($objRecordSet.Fields.Item($i).Value)
+      }
+      $records.Add($cell)
+      $objRecordSet.MoveNext()
+    }
+    $objRecordSet.Close()
+
+    $objRecordSet.Open(
+      "SELECT COUNT(1) FROM $tableName",
```
DmitryMK, line 281: Does it make sense to pass item name and library separately and request this information from sasHelp.vTables using `SELECT nobs FROM sasHelp.vTable WHERE libName = "$item.library" and memName = "$item.library"`
Although in both cases CPU time is extremely small, for 3 million records vTable approach is around 8 times faster and probably uses less resources.

DmitryMK: I realized it is better to keep the current approach. Later if we want to add filters in the data view, the current approach will be better suited.

client/src/connection/itc/script.ts
```diff
@@ -224,5 +243,86 @@ class SASRunner{
...
+
+    $records = [List[List[object]]]::new()
+    $fields = $objRecordSet.Fields.Count
+
+    if ($objRecordSet.EOF) {
+      Write-Host '{"rows": [], "count": 0}'
+      return
+    }
+
+    $objRecordSet.AbsolutePosition = $start + 1
```
DmitryMK, line 268: The [documentation](https://learn.microsoft.com/en-us/office/client-developer/access/desktop-database-reference/recordset-absoluteposition-property-dao) states:
> Setting the AbsolutePosition property to a value greater than zero on a newly opened but unpopulated Recordset object causes a trappable error. Populate the Recordset object first with the MoveLast method.

Should MoveLast method be applied first?

scnwwu:
Hi @danila-grobov, now that #1388 is merged. Please rebase your change with latest main branch. Thanks.
I'm not clear why you need to mock uuid while #1388 seems fine.

client/src/connection/itc/script.ts
danila-grobov: @DmitryMK 

Not sure what the documentation is talking about, maybe the behavior differs with sas data provider. It works with and without this statement. I can set the absolute position to whatever I want, as long as it is within bounds of the dataset. I did some testing on this.

I wouldn't like to add it, since it works fine now and going all the way to the end of the dataset without a need sounds wasteful.

danila-grobov (commented):
@scnwwu

The file browsing functionality is not covered by tests, but the table viewer is covered - that's why they didn't have to mock it, but I had to.

I've merged the changes in, and also tested the file browser on our environment works great!

You can go ahead and review again.

client/src/connection/itc/script.ts
DmitryMK: I though $objRecordset.MoveFirst() should be used. At the same time I read the note again and I realized it is not applicable in this case. It talks about cases when Recordset is not populated, which is not the case here. I guess the problem can happen when $start + 1 exceeds the number of records, logically it should not happen, as we start reading with $start = 0.

Still one more question, why do you use $start + 1 and not $start. From the description it looks like AbsolutePosition  is zero-based, so the first record is 0. Does SAS use the first record for something else?

danila-grobov: No, it throws an error if you try to access a record at position 0. And I've checked, no records are missing if we start reading from 1. I'm guessing it's just the SAS data provider implementation quirk, it doesn't completely match what's written in adodb documentation.

scnwwu:
Thank you. Overall good to me. I'd like @scottdover who originally wrote these to take a look and approve.

client/src/connection/util.ts
```diff
@@ -24,3 +24,74 @@ export const extractTextBetweenTags = (
...
+    ["DATEAMPM", "DATETIME", "NLDATM", "NLDATMAP", "NLDATMW"].some((f) =>
+      format.includes(f),
+    );
+
+  const isCurrencyFormat = () =>
+    ["NLMNI", "NLMNL", "NLMNY", "YEN", "DOLLAR", "EURO"].some((f) =>
+      format.includes(f),
+    );
+
+  if (type === "num") {
```
scottdover, line 81: nitpick. Can we do 
```
if (type !== "num") {
  return type;
}
```
here and un-indent the rest

scottdover (commented):
This looks great! The performance is so much better! Thank you, @danila-grobov. 

Did we test this with a COM connection? I saw things in the script that _looked_ IOM specific, but maybe that isn't an issue?

client/src/connection/util.ts
danila-grobov: Fixed ce94b23a051614c345f4e2ed449f4cdae80b1d60

scnwwu (approved):

scnwwu:
> @scnwwu @scottdover I have rewritten getLibraries and getTables using the same approach @danila-grobov used and it changed loading time of libraries from seconds to milliseconds in my case (which in my opinion makes user experience much better). I think it will be nice to be released together with this change. Should I submit a separate PR or add a commit in this one?

Thank you. That would be a good improvement. This PR is from @danila-grobov's forked repo I'm not sure if you're able to add a commit to the branch. I guess you can create a PR to @danila-grobov's branch and if @danila-grobov approved and merged it, it will be in this PR. Or you can create a separate PR after this one merged.

DmitryMK:
> > @scnwwu @scottdover I have rewritten getLibraries and getTables using the same approach @danila-grobov used and it changed loading time of libraries from seconds to milliseconds in my case (which in my opinion makes user experience much better). I think it will be nice to be released together with this change. Should I submit a separate PR or add a commit in this one?
> 
> Thank you. That would be a good improvement. This PR is from @danila-grobov's forked repo I'm not sure if you're able to add a commit to the branch. I guess you can create a PR to @danila-grobov's branch and if @danila-grobov approved and merged it, it will be in this PR. Or you can create a separate PR after this one merged.

In this case I think it will be better if I create a separate PR, I'll try to do it tomorrow.

client/src/connection/util.ts
```diff
@@ -24,3 +24,76 @@ export const extractTextBetweenTags = (
...
+      "DD",
+      "YY",
+      "EURDF",
+      "JUL",
+      "YEAR",
+      "DAY",
+      "MONTH",
+      "MON",
+      "DOWNAME",
+    ].some((f) => format.includes(f)) &&
```
DmitryMK, line 52: Can we add [8601](https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.5/leforinforref/p1a0qt18rxydrkn1b0rtdfh2t8zs.htm) formats? They are common in clinical programming.



Date: E8601DA, E8601DN, D8601DA, D8601DN: ([B|E]8601D[N|A])
Datetime: ([B|E]8601(D[T|Z|X]|LX))
Time: ([B|E]8601(T[M|Z|X]|LZ))

snlwih: @scnwwu: Perhaps enhancing SAS format support to include the lateset available SAS-provided formats should be moved into a dedicated pull request?

@DmitryMK: Also wanted to create awarenes for the fact the the doc link mentioned in the previous comment for National Language Support formats is really old. Here are doc links that will always point to the latest production: 
 * [doc link for Dictionary of Formats](https://go.documentation.sas.com/doc/en/pgmsascdc/default/leforinforref/p0z62k899n6a7wn1r5in6q5253v1.htm)
 * [doc link for the Dictionary of Formats for NLS](https://go.documentation.sas.com/doc/en/pgmsascdc/default/nlsref/p0qenieob5jr4mn1dcthvv8822vo.htm).

Tip: Use .../**default**/... instead of .../**v_xxx**/... in SAS Viya 4 doc links, so they resolve to the latest documentation.

Zhirong2022 (approved):

danila-grobov:
@scnwwu will do later today

danila-grobov:
@scnwwu Done, you can go ahead and merge.

scnwwu:
Thank you again. Great contribution!

Issue #1272, opened 2024-11-11
Failed to open a file having character '#' in its name, The editor could not be opened due to an unexpected error: Unable to read file ...


Zhirong2022:
**Describe the bug**
Suppose there is a valid Viya connection. Access the SASStudio, create a file with name test#.sas and save it on SAS Content->My Folder. Open VSC SAS Extension and go to SAS Content Explorer, try to open the file. It failed to open the file in VSC. 

**Steps to reproduce**
1. Sign in with a valid Viya connection
2. Go to SASStudio and create a file with name test#.sas
3. Save the file under SAS Content->My Folder
4. Back to VSC and go to SAS Content, open the file test#.sas under My Folder

**Expected behavior**
It can open the file in VSC successfully

**Screenshots**
![image](https://github.com/user-attachments/assets/d4c82193-6c8e-4f3e-871b-9ddc20376311)

**Environment (please complete the following information):**
Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.11.0]

PR #1530, opened 2025-06-18
fix:deal with the hash sign

scnnaw:
**Summary**
Fix https://github.com/sassoftware/vscode-sas-extension/issues/1272

**Testing**
the test#.sas can be open now;
the icon in the tree view is a correct sas icon;
it can create the xxx#.sas file from extension.


scnwwu (approved):

Zhirong2022:
Character "#" is handled properly if trying to create or open a file with a name that contains "#".

Zhirong2022 (approved):

PR #1534, opened 2025-06-19
feat: be able to ignore region from format

### LINKED ISSUE #1197: [Code formatting] Support option to exclude sections of code when using Format Document
I'd like it is not to be formated in the PROC FORMAT or PROC SQL, or other special situations. For example, if I format the code below
```SAS
proc format;
  value range
    40000-high='High'
    26000-< 40000='Medium'
    other ='Low';

  invalue evaluation
    'O'=4
    'S'=3
    'E'=2
    'C'=1
    'N'=0;
run;
```
It becomes
```SAS
proc format;
  value range 40000-high='High' 26000-< 40000='Medium' other='Low';

  invalue evaluation 'O'=4 'S'=3 'E'=2 'C'=1 'N'=0;
run;
```

I'd like to keep the format of the first one. 

For PROC SQL, it is similar. Usually I'll make a separate sql script file to format code and then paste it back, however, sas extension's format will make it `flat` again. 

I'd like to suggest add some thing like REGION. For example,
```SAS
proc format;
  /* sas-format-disable */
  value range
    40000-high='High'
    26000-< 40000='Medium'
    other ='Low';

  invalue evaluation
    'O'=4
    'S'=3
    'E'=2
    'C'=1
    'N'=0;
  /* sas-format-enable */
run;

proc sql;
  * sas-format-disable;
  select
    momage,
    (weight * 0.01) as newweight
  from
    outdata
  where
    calculated newweight > 5;
  * sas-format-enable ;
run;
```

**Environment**
SAS version 9.4

scnwwu:
**Summary**
Resolves #1197

User can now add a custom region as below, which will be ignored by SAS code formatter.
```sas
proc format library=library;
/* region format-ignore */
  invalue evaluation 'O'=4
                     'S'=3
                     'E'=2
                     'C'=1
                     'N'=0;
/* endregion */
run;
```

Note that if your custom region intersect with other regions, the result may be unexpected.

**Testing**
See code within the custom region remain unchanged after performing "Format Document" command. Other code get formatted as expected.

### DIFF
--- a/client/testFixture/formatter/expected.sas
+++ b/client/testFixture/formatter/expected.sas
@@ -172,3 +172,13 @@ print('hello')
 endinteractive;
   /* comment */
 run;
+
+proc format library=library;
+/* region format-ignore */
+  invalue evaluation 'O'=4
+                     'S'=3
+                     'E'=2
+                     'C'=1
+                     'N'=0;
+/* endregion */
+run;
--- a/client/testFixture/formatter/unformatted.sas
+++ b/client/testFixture/formatter/unformatted.sas
@@ -166,3 +166,12 @@ print('hello')
 endinteractive;
 /* comment */
 run;
+proc format library=library;
+/* region format-ignore */
+  invalue evaluation 'O'=4
+                     'S'=3
+                     'E'=2
+                     'C'=1
+                     'N'=0;
+/* endregion */
+run;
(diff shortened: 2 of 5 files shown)


ShermanWang0324:
Tested the code within /* region format-ignore */ will not be formatted.
fixed

PR #1557, opened 2025-07-11
feat(wip): add support for user-defined file navigation root

scottdover:
**Summary**

**Testing**
How did you test this change?


scottdover:
Hey @clangsmith / @smorrisj . Do either of you know how sas server via viya should be have with `fileNavigationRoot` set to `ROOT`? `FileSystemApi` didn't seem to provide any guidance for how to proceed with that. Right now, it's basically undefined behavior.

scottdover:
@skyeaussie This introduces a new "Copy Path" context menu item to copy the path of a file or folder. Here is how it is positioned in the list of items:

with a file - 
Screenshot 2025-07-11 at 1 26 23 PM

with a folder - Screenshot 2025-07-11 at 1 26 29 PM

Let me know if it should be ordered differently in the menu.

Also,  this is the error message for when a user specifies an invalid file nav root: "Unable to load SAS Server files with provided fileNavigationRoot/fileNavigationCustomRootPath settings." Let me know if this language could use some tweaking.

client/src/connection/rest/index.ts
```diff
@@ -39,6 +39,36 @@ class RestSession extends Session {
     this._config = value;
   }
 
+  public async contextAttributes(): Promise<SessionContextAttributes> {
+    const context = await this.getContext();
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    return context.attributes as SessionContextAttributes;
```
scnwwu, line 46: nit: shall we cache the attributes when create session so that we don't need to send getContext HTTP request twice?

package.nls.json
```diff
@@ -44,6 +45,8 @@
   "configuration.SAS.connectionProfiles.profiles.connectionType": "SAS Profile Connection Type",
   "configuration.SAS.connectionProfiles.profiles.context": "SAS Viya Context",
   "configuration.SAS.connectionProfiles.profiles.endpoint": "SAS Viya Connection Profile Endpoint",
+  "configuration.SAS.connectionProfiles.profiles.fileNavigationCustomRootPath": "File navigation custom root path",
```
scnwwu: When doing translation, I realized that it seems we never say "File navigation" in UI. It's a general phrase and may not look obivous when translated. Shall we say something matching the UI (e.g. SAS Server)?

package.json
```diff
@@ -583,6 +628,11 @@
         "title": "%commands.SAS.authorize%",
         "category": "SAS"
       },
+      {
```
scnwwu, line 631: I believe it's similar to other context menu commands that is not suitable for Command Palette. Would you add in below "commandPalette" entry to have a `"when": false` entry?

scnwwu (commented):
Do you want to add some doc to
https://sassoftware.github.io/vscode-sas-extension/Configurations/Profiles/additional

package.nls.json
scottdover: Replaced `File navigation` with `SAS Server` in https://github.com/sassoftware/vscode-sas-extension/pull/1557/commits/aef333173872a977a4f148c82e88abf42903d2f2

package.json
scottdover: Done in https://github.com/sassoftware/vscode-sas-extension/pull/1557/commits/149b547f1a96e2954ba0773579d4810785521a6e

client/src/connection/rest/index.ts
scottdover: Done in https://github.com/sassoftware/vscode-sas-extension/pull/1557/commits/300f518dd253adfe729278ff9cfdcfb0f7a12e96

scnwwu:
@scottdover the description of this PR mentions `fileNavigationRoot === 'ROOT'`, I guess it means `"SYSTEM"`? Would you kindly update it to avoid confusing? Thanks.

client/src/connection/rest/RestServerAdapter.ts
```diff
@@ -75,6 +98,7 @@ class RestSASServerAdapter implements ContentAdapter {
     const reconnect = async () => {
       return await this.establishConnection();
     };
+    // TODO (feat/root-nav-path) Make sure we're overriding config options with compute context data
```
scnwwu: Is this still a TODO?

scottdover: Oops. Good catch. Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/1557/commits/50f47d6d85a66b0f048448760854123890f20501

Zhirong2022:
Suppose the connection type is Local94(COM), and specify the value of fileNavigationCustomRootPath as blank. 
image

Zhirong2022:
fileNavigationCustomRootPath is configured as the same value, it has different behavior between SAS Studio and SAS Extension.

In SAS Studio, it will show the specified folder name instead of '**Home**'

image

scottdover:
> 1. The icon for the specified folder has different look.

This one should be resolved by https://github.com/sassoftware/vscode-sas-extension/pull/1557/commits/0d2f45d7b9189ea8446610df37b503edfc9a5869 . Now, if the directory says "Home", we'll still show the "Home" icon. Otherwise, we'll show a regular folder icon.

scottdover:
> 2. It can create a new folder under the specified rootPath in SAS Studio, and some other actions on ContextMenu. On the other hand, it has only 'Copy path' in VSC extension.

This _should_ be resolved now

Zhirong2022:
All issues are fixed properly.

Issue #1136, opened 2024-08-12
No warning message if trying to delete a folder which contains unsaved files 


Zhirong2022:
**Is your feature request related to a problem? Please describe.**
Open VSC SAS Extension base on Main branch. Create a new folder with some unsaved files, delete the created folder, it has no warning message in such kind of situation. However, it has warning for the user if do the same on VSC Explorer.

Steps to reproduce
1.Sign in with Viya connection
2.Go to SAS Content Explorter 
3.Create a new folder name 'Test' under My Folder
4.Create 2 new unsaved files 'test01.sas', 'test02.sas' under folder 'Test'
5.Delete folder 'Test'

**Describe the solution you'd like**
It would be better to show warning message if trying to delete a folder which contains unsaved files.

**Additional context**
![image](https://github.com/user-attachments/assets/7f1a17cc-5fa3-4076-a246-b775c8a8cb03)

PR #1559, opened 2025-07-16
Added feature to check for unsaved files in a folder upon deletion. 

kpatl1:
**Summary**
I have added a function that checks if there are unsaved files within a folder. If so, it will open a modal asking if the user wants to actually delete the folder. On deletion it will then move to recycle bin. 
**Testing**
How did you test this change?
I followed the steps to recreate the feature request by joining a viya server, making a new folder with two unsaved files, and then try to delete the folder. After deleting it, I got the message if I truly did want to and then I had the option to continue or cancel. On continuation, the folder then moved to recycle bin. 

From:
https://github.com/sassoftware/vscode-sas-extension/issues/1136

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -56,7 +56,26 @@ import {
   getEditorTabsForItem,
   getFileStatement,
   isContainer as getIsContainer,
-} from "./utils";
+} from "./utils"; // already imported
+
+async function getDirtyFilesInFolder(item: ContentItem, provider: ContentDataProvider): Promise<TextDocument[]> {
+  const dirtyDocs: TextDocument[] = [];
+  async function checkChildren(folder: ContentItem) {
+    const children = await provider.getChildren(folder);
+    for (const child of children) {
+      if (getIsContainer(child)) {
+        await checkChildren(child);
+      } else {
+        const doc = workspace.textDocuments.find(d => d.uri.toString() === child.vscUri.toString());
+        if (doc && doc.isDirty) {
+          dirtyDocs.push(doc);
+        }
+      }
+    }
+  }
+  await checkChildren(item);
+  return dirtyDocs;
+}
 
 class ContentDataProvider
   implements
@@ -353,6 +372,21 @@ class ContentDataProvider
   }
 
   public async deleteResource(item: ContentItem): Promise<boolean> {
+    if (getIsContainer(item)) {
+      const dirtyDocs = await getDirtyFilesInFolder(item, this);
+      if (dirtyDocs.length > 0) {
+        const confirm = await window.showWarningMessage(
+          l10n.t("There are unsaved files in this folder. Deleting it will discard unsaved changes. Do you want to continue?"),
+          { modal: true },
+          l10n.t("Delete Anyway"),
+          l10n.t("Cancel")
+        );
+        if (confirm !== l10n.t("Delete Anyway")) {
+          return false;
+        }
+      }
+    }
+
     if (!(await closeFileIfOpen(item))) {
       return false;
     }
--- a/client/src/connection/rest/SASContentAdapter.ts
+++ b/client/src/connection/rest/SASContentAdapter.ts
@@ -1,6 +1,6 @@
 // Copyright © 2024, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { FileType, Uri, authentication } from "vscode";
+import { FileType, Uri, authentication, window, workspace } from "vscode";
 
 import axios, {
   AxiosError,
@@ -560,6 +560,10 @@ class SASContentAdapter implements ContentAdapter {
   public async recycleItem(
     item: ContentItem,
   ): Promise<{ newUri?: Uri; oldUri?: Uri }> {
+    if (!(await this.checkForUnsavedFilesInFolder(item, "recycle"))) {
+      return {};
+    }
+
     const recycleBin = this.getRootFolder("@myRecycleBin");
     if (!recycleBin) {
       // fallback to delete
@@ -737,6 +741,37 @@ class SASContentAdapter implements ContentAdapter {
     }
 
     return `${getResourceIdFromItem(myFavoritesFolder)}/members/${favoriteId}`;
+  }
+
+  private async checkForUnsavedFilesInFolder(
+    item: ContentItem,
+    action: "delete" | "recycle",
+  ): Promise<boolean> {
+    if (!isContainer(item)) {
+      return true;
+    }
+    const children = await this.getChildItems(item);
+    const unsavedFiles = children.filter((child) => {
+      if (!child.vscUri) {
+        return false;
+      }
+      const doc = workspace.textDocuments.find(
+        (d) => d.uri.toString() === child.vscUri.toString(),
+      );
+      return doc?.isDirty;
+    });
+
+    if (unsavedFiles.length > 0) {
+      const actionText = action === "delete" ? "Delete Anyway" : "Move Anyway";
+      const message = `There are unsaved files in this folder. Do you want to continue?`;
+      const result = await window.showWarningMessage(
+        message,
+        { modal: true },
+        actionText,
+      );
+      return result === actionText;
+    }
+    return true;
   }
 }
 


client/src/connection/rest/SASContentAdapter.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright © 2024, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { FileType, Uri, authentication } from "vscode";
```
scnwwu: This file is an adapter to implement on REST service. This feature looks not necessarily specific to REST.
UI feature should be in UI component
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/ContentNavigator/index.ts#L105
You can see warning message there. Please follow the pattern to make UI strings translatable.

client/src/connection/rest/SASContentAdapter.ts
```diff
@@ -738,6 +742,36 @@ class SASContentAdapter implements ContentAdapter {
...
-}
 
+  private async checkForUnsavedFilesInFolder(
+    item: ContentItem,
+    action: "delete" | "recycle",
+  ): Promise<boolean> {
+    if (!isContainer(item)) {
+      return true;
+    }
+    const children = await this.getChildItems(item);
```
scnwwu: Looks like this can only get one level of children. VS Code native file explorer can track across multiple levels.
I'm concerned that it will take too long to traverse all children.

kpatl1:
Added fixes. Put the logic into the UI and Service layer. Children are found recursively for the folder and added to a set. Open files are then referenced into the set to see if they are within it. If the file is dirty and in the children set, a popup will appear. Overall, time complexity boils down to O(N + M) I believe where N is the number of children and M being the number of files being open in the window.

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -505,6 +505,46 @@ class ContentDataProvider
...
   }
 
+  public async checkFolderDirty(resource: ContentItem): Promise<boolean> {
+    if (!resource.vscUri) {
+      return false;
+    }
+
+    const descendants: ContentItem[] = [];
+    async function recurse(item: ContentItem) {
+      const kids = await this.model.getChildren(item);
```
scnwwu: Each getChildren call is an HTTP request. There can be many nested folders in real world so there can be many HTTP calls. Due to network latency, it can take several seconds to respond to user, which is not acceptable.

kpatl1: Would it be faster to do a BFS approach with a promise.all() so the HTTP requests are made concurrently?


```suggestion
public async checkFolderDirty(resource: ContentItem): Promise<boolean> {
  if (!resource.vscUri) {
    return false;
  }

  const descendants: ContentItem[] = [];
  const queue: ContentItem[] = [resource];


  while (queue.length > 0) {
    const currentBatch = queue.splice(0); 
    const childrenArrays = await Promise.all(
      currentBatch.map(item => this.model.getChildren(item))
    );

    for (const children of childrenArrays) {
      for (const child of children) {
        descendants.push(child);
        if (getIsContainer(child)) {
          queue.push(child);
        }
      }
    }
  }

```

scnwwu: The number of concurrent requests a server can handle are limited. In addition, a user may have many levels of nested folders. It's still risky when a user clicked "Delete" and no response in a few seconds. A user may click again and result in confusing or error status.
This feature request looks trivial. The dirty file won't lost. The user can continue edit and save it to other place. If there's not a neat way to maintain good performance and user experience, I'd vote to not support this.

Some ideas if you want to continue think about it.
Perhaps we should start with traversing dirty files rather than traversing child folders. If there's no dirty files or no dirty files that belongs to our content schema we don't need to do anything. Guessing in most cases the dirty files may not be too much.
And for each dirty file belongs to our content, check its ancesters to see if any of them is the uri to be deleted. This way sounds more efficient than traversing child folders.
For each `ContentItem`, we can directly get its [parentFolderUri](https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/ContentNavigator/types.ts#L14), no HTTP call needed.
Now the only problem is to get the `ContentItem` based on an uri. There's a [getItemOfUri](https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/ContentNavigator/types.ts#L84) method for this. But it involves HTTP call and I think we should avoid it.
In most cases, to open a file to edit, user click the tree item and the item was already loaded by [getTreeItem](https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/ContentNavigator/ContentDataProvider.ts#L213). So in `getTreeItem` we can cache the uri and parentFolderUri map. Thus no extra HTTP call needed.
There might be some cases where `getTreeItem` didn't get called for a dirty file. For example if you reload the VS Code window the dirty file preserves but the tree might not be loaded. But as I said I guess it's acceptable to not show this dialog in some cases if it's not easy.

kpatl1: I agree. I am open to implementing this through the caching of the getTreeItem. However,  in the case that getTreeItem is not called, then the feature is inherently deployed with a "bug" built-in. But, as you said it is okay in some cases to not display the message we can implement this, and then continue to see if there is a way to improve it.

kpatl1: I have gone in and changed the logic to reduce the HTTP requests -- only using it as a fallback if the file is not already cached. Also added a depth check limit of 10 to prevent overflow. However, I am having trouble with it working with a file that is in a subfolder of a folder that is not saved when trying to delete the main folder. 

Ex. 
Test/
├── 1.sas
└── Test2/
    └── 2.sas

When trying to delete 2.sas, it does not show a warning.

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -214,6 +216,11 @@ class ContentDataProvider
     const isContainer = getIsContainer(item);
     const uri = await this.model.getUri(item, false);
 
+    // Cache the URI to parent mapping
+    if (item.parentFolderUri) {
+      this.uriToParentMap.set(uri.toString(), item.parentFolderUri);
```
scnwwu:
```suggestion
      this.uriToParentMap.set(item.uri, item.parentFolderUri);
```

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -505,6 +512,96 @@ class ContentDataProvider
...
+
+        const scheme = doc.uri.scheme;
+        return (
+          scheme === "sasContent" ||
+          scheme === "sasServer" ||
+          scheme === "sasContentReadOnly" ||
+          scheme === "sasServerReadOnly"
+        );
+      })
+      .map((doc) => doc.uri.toString());
```
scnwwu:
```suggestion
      .map((doc) => doc.uri);
```

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -505,6 +512,96 @@ class ContentDataProvider
...
+
+    if (dirtyFiles.length === 0) {
+      return false;
+    }
+
+    for (const dirtyFileUri of dirtyFiles) {
+      if (await this.isDescendantOf(dirtyFileUri, targetFolderUri)) {
```
scnwwu: `import { getResourceId } from "../../connection/rest/util";`
```suggestion
      if (await this.isDescendantOf(getResourceId(dirtyFileUri), targetFolderUri)) {
```

package-lock.json
scnwwu, line 1: please do NOT change this file when no dependencies changed

client/src/components/ContentNavigator/index.ts
```diff
@@ -523,5 +540,4 @@ class ContentNavigator implements SubscriptionProvider {
     }
   }
 }
-
```
scnwwu: unnecessary change

client/src/components/ContentNavigator/index.ts
```diff
@@ -107,30 +107,47 @@ class ContentNavigator implements SubscriptionProvider {
...
-              const deleteResult = moveToRecycleBin
-                ? await this.contentDataProvider.recycleResource(resource)
-                : await this.contentDataProvider.deleteResource(resource);
-              if (!deleteResult) {
-                window.showErrorMessage(
-                  isContainer
-                    ? Messages.FolderDeletionError
-                    : Messages.FileDeletionError,
-                );
+              if (resource.contextValue.includes("delete")) {
```
scnwwu: looks like a bad merge. This line was deleted by
https://github.com/sassoftware/vscode-sas-extension/pull/1557/files#diff-67d022ea7fd74e7337764a8ef0bd8aaf9c4aada502744fdd38facc2f4f3e9372L109

client/src/components/ContentNavigator/index.ts
scnwwu: this is unresolved. it's still showing an unnecessary removed line

client/src/components/ContentNavigator/index.ts
scnwwu: this is not resolved. it's not necessary to add this line in this change. this check was moved to line 106 above by a previous PR

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -214,6 +217,11 @@ class ContentDataProvider
     const isContainer = getIsContainer(item);
     const uri = await this.model.getUri(item, false);
 
+    // Cache the URI to parent mapping
+    if (item.parentFolderUri) {
+      this.uriToParentMap.set(item.uri, item.parentFolderUri);
+    }
```
scnwwu: Top level folder doesn't have a parentFolder.
Currently in `isDescendantOf`, as long as there's not a parentFolder, it will call `this.model.getResourceByUri`. And again the parentFolder is still empty. They're unnecessary network calls recursively until hit the 10 times limit. They should be avoided.
The fallback call is only used in case an item didn't go through this line. But if there's not a parentFolderUri here, it means this item really doesn't have a parent. Calling `this.model.getResourceByUri` is unnecessary and will result the same.
Perhaps something like

```suggestion
   const STOP_SIGN = '__STOP_SIGN__';
   this.uriToParentMap.set(item.uri, item.parentFolderUri ?? STOP_SIGN);
```

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -505,6 +513,94 @@ class ContentDataProvider
...
+        return false;
+      }
+    }
+
+    let depth = 0;
+    while (currentParentUri || depth >= 10) {
+      if (currentParentUri === ancestorFolderUri) {
+        return true;
+      }
+
```
scnwwu, line 595:
```suggestion
  if (currentParentUri === STOP_SIGN) {
    return false;
  }
```

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -79,6 +80,8 @@ class ContentDataProvider
   public dropMimeTypes: string[];
   public dragMimeTypes: string[];
 
+  private uriToParentMap = new Map<string, string>();
```
scnwwu, line 84: I guess the cache should be cleared when [refresh](https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/ContentNavigator/ContentDataProvider.ts#L445)

kpatl1 (commented):
resolved comments, and tested functionality again.

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -214,6 +218,9 @@ class ContentDataProvider
     const isContainer = getIsContainer(item);
     const uri = await this.model.getUri(item, false);
 
+    // Cache the URI to parent mapping
+    this.uriToParentMap.set(item.uri, item.parentFolderUri ?? STOP_SIGN);
```
scnwwu: sorry when I try it more, it looks like it's possible that `item.parentFolderUri` is `''`.
```suggestion
    this.uriToParentMap.set(item.uri, item.parentFolderUri ? item.parentFolderUri : STOP_SIGN);
```

scnwwu (approved):

Zhirong2022:
It has no further response if trying to delete the folder which contains unsaved file after switch to a new Viya profile.
Steps to reproduce:
Suppose having two valid Viya connection
1.Sign in with one of the Viya server
2.Go to SAS Server, create a new file under a specified personal file folder and kept it unsaved
3.Switch to another Viya server
4.Delete the specified file folder which contains the unsaved file

scnwwu:
> > It has no further response if trying to delete the folder which contains unsaved file after switch to a new Viya profile. Steps to reproduce: Suppose having two valid Viya connection 1.Sign in with one of the Viya server 2.Go to SAS Server, create a new file under a specified personal file folder and kept it unsaved 3.Switch to another Viya server 4.Delete the specified file folder which contains the unsaved file
> 
> I believe that is because VSCode will no longer mark the file as dirty since you are reloading the connection. I am unsure is there is a way to work around this as it is a base issue coming from VSCode no?

Hi @kpatl1, @Zhirong2022 and I looked at it. I think it's not by VS Code. The dirty file is still dirty file. It's the underlying SAS server changed. It doesn't make sense to look up the parent of a file from a previous server on the new server. I would suggest to remove the fallback logic to fetch parent from server. If the cache doesn't contain the uri, it's acceptable to not pop up.

Zhirong2022:
The raised issue has been fixed properly.

Zhirong2022 (approved):

Issue #1125, opened 2024-08-01
Download option for SAS output (HTML, ODS, LOG etc)


ATFischer:
**Is your feature request related to a problem? Please describe.**
Please let users download SAS output in SAS Notebook cells.

**Describe the solution you'd like**
Right mouse button > download.

PR #1576, opened 2025-07-29
added ability to save cell output

kpatl1:
**Summary**
Added a button to call the command to save a cells output from a SAS Notebook. It will get the cell at the selected index and use that to save the output after a user chooses between ODS HTML or LOG output. 
**Testing**
Created a SAS Notebook with multiple cells and tried to save them, there are checks to make sure that the cell has output as well. 

https://github.com/sassoftware/vscode-sas-extension/issues/1125

### DIFF
--- a/client/src/components/notebook/exporters/index.ts
+++ b/client/src/components/notebook/exporters/index.ts
@@ -139,5 +139,5 @@ export const saveOutput = async () => {
 
   await workspace.fs.writeFile(uri, Buffer.from(content));
 
-  window.showInformationMessage(l10n.t("Saved downloaded to {0}", uri.fsPath));
+  window.showInformationMessage(l10n.t("Saved to {0}", uri.fsPath));
 };
--- a/client/src/components/notebook/exporters/index.ts
+++ b/client/src/components/notebook/exporters/index.ts
@@ -30,3 +30,34 @@ export const exportNotebook = async (client: LanguageClient) => {
 
   workspace.fs.writeFile(uri, Buffer.from(content));
 };
+
+export const exportNotebookCell = async (client: LanguageClient) => {
+  const notebook = window.activeNotebookEditor?.notebook;
+  const activeCell = window.activeNotebookEditor?.selection?.start;
+
+  if (!notebook || activeCell === undefined) {
+    return;
+  }
+
+  const cell = notebook.cellAt(activeCell);
+  if (!cell) {
+    return;
+  }
+
+  const uri = await window.showSaveDialog({
+    filters: { SAS: ["sas"], HTML: ["html"] },
+    defaultUri: Uri.parse(
+      `${path.basename(notebook.uri.path, ".sasnb")}_cell_${activeCell + 1}`,
+    ),
+  });
+
+  if (!uri) {
+    return;
+  }
+
+  const content = uri.path.endsWith(".html")
+    ? await exportToHTML(notebook, client, activeCell)
+    : exportToSAS(notebook, activeCell);
+
+  workspace.fs.writeFile(uri, Buffer.from(content));
+};
--- a/client/src/components/notebook/exporters/index.ts
+++ b/client/src/components/notebook/exporters/index.ts
@@ -52,6 +52,31 @@ export const exportNotebookCell = async (client: LanguageClient) => {
     return;
   }
 
+  const exportChoice = await window.showQuickPick(
+    [
+      {
+        label: "HTML Output Only",
+        description: "Export only tables, charts, and visualizations",
+        detail: "Clean output without log messages",
+        includeLog: false,
+      },
+      {
+        label: "HTML Output + Log",
+        description: "Export output with execution log",
+        detail: "Includes notes, warnings, and execution details",
+        includeLog: true,
+      },
+    ],
+    {
+      placeHolder: "What would you like to include in the export?",
+      ignoreFocusOut: true,
+    },
+  );
+
+  if (!exportChoice) {
+    return;
+  }
+
   const uri = await window.showSaveDialog({
     filters: { HTML: ["html"] },
     defaultUri: Uri.parse(
@@ -63,7 +88,13 @@ export const exportNotebookCell = async (client: LanguageClient) => {
     return;
   }
 
-  const content = await exportToHTML(notebook, client, activeCell, true);
+  const content = await exportToHTML(
+    notebook,
+    client,
+    activeCell,
+    true,
+    exportChoice.includeLog,
+  );
 
   workspace.fs.writeFile(uri, Buffer.from(content));
 };
--- a/client/src/components/notebook/exporters/index.ts
+++ b/client/src/components/notebook/exporters/index.ts
@@ -66,25 +66,19 @@ export const saveOutput = async () => {
 
   const choices: Array<{
     label: string;
-    description: string;
-    detail: string;
     outputType: "html" | "log";
   }> = [];
 
   if (odsItem) {
     choices.push({
-      label: l10n.t("Download as ODS HTML"),
-      description: "",
-      detail: "",
+      label: l10n.t("Save as ODS HTML"),
       outputType: "html",
     });
   }
 
   if (logItem) {
     choices.push({
-      label: l10n.t("Download as Log"),
-      description: "",
-      detail: "",
+      label: l10n.t("Save as Log"),
       outputType: "log",
     });
   }
--- a/client/src/components/notebook/exporters/index.ts
+++ b/client/src/components/notebook/exporters/index.ts
@@ -1,7 +1,6 @@
 // Copyright © 2025, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { Uri, window, workspace } from "vscode";
-import * as vscode from "vscode";
+import { Uri, l10n, window, workspace } from "vscode";
 import type { LanguageClient } from "vscode-languageclient/node";
 
 import path from "path";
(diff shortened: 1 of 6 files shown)


kpatl1:
Fixed the feature to meet specs as given on teams.

package.json
```diff
@@ -1097,6 +1103,27 @@
...
+        }
+      ],
+      "notebook/cell/title": [
+        {
+          "command": "SAS.notebook.exportCell",
+          "when": "notebookType == 'sas-notebook'",
+          "group": "inline"
+        }
+      ],
+      "notebook/cell/output": [
```
scnwwu: looks like there's no such entry defined by VS Code. Does it take any effect? If no, please remove it.

package.json
```diff
@@ -1097,6 +1103,27 @@
           "command": "SAS.notebook.export"
         }
       ],
+      "notebook/cell/execute": [
```
scnwwu: unnecessary to inject to execute, it's not an execute command

package.json
```diff
@@ -1097,6 +1103,27 @@
...
+        {
+          "command": "SAS.notebook.exportCell",
+          "when": "notebookType == 'sas-notebook'",
+          "group": "inline"
+        }
+      ],
+      "notebook/cell/title": [
```
scnwwu, line 1105: It looks like unfortunately VS Code doesn't provide a dedicated menu register entry for output only. The title entry displays on the code cell with a lot of menu items. It looks not ideal since it's not exporting the whole cell. It may be hard to find if the user just want to save the output.
If we can display a context menu only on the output, it looks more intuitive.
We're rendering the output by our own [renderer](https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/notebook/renderers/HTMLRenderer.ts#L23). Perhaps we can render a context menu by ourselves. And [communicate](https://code.visualstudio.com/api/extension-guides/notebook#interactive-notebooks-communicating-with-the-controller) with the controller to call a command.
However it looks too involved. What do you think? Perhaps we can live with the title entry for now.

package.json
```diff
@@ -825,6 +825,12 @@
         "title": "%commands.SAS.notebook.export%",
         "category": "SAS Notebook"
       },
+      {
+        "command": "SAS.notebook.exportCell",
```
scnwwu: Now that it's not exporting the whole cell. Please rename it something like saveOutput

client/src/components/notebook/exporters/toSAS.ts
```diff
@@ -2,11 +2,19 @@
 // SPDX-License-Identifier: Apache-2.0
 import { NotebookCell, NotebookDocument } from "vscode";
 
-export const exportToSAS = (notebook: NotebookDocument) =>
-  notebook
+export const exportToSAS = (notebook: NotebookDocument, cellIndex?: number) => {
```
scnwwu: Please remove unused changes. Thanks

client/src/components/notebook/exporters/toHTML.ts
```diff
@@ -32,8 +32,14 @@ hljs.registerLanguage("sql", sql);
 export const exportToHTML = async (
   notebook: NotebookDocument,
   client: LanguageClient,
+  cellIndex?: number,
```
scnwwu: Please remove unused changes. Thanks.

package.json
kpatl1: That sounds like a good ux improvement. I believe we can go with this version for now, and continue to iterate to improve the feature to perhaps implement in this way.

package.json
```diff
@@ -1097,6 +1103,13 @@
           "command": "SAS.notebook.export"
         }
       ],
+      "notebook/cell/title": [
+        {
+          "command": "SAS.notebook.saveOutput",
+          "when": "notebookType == 'sas-notebook'",
+          "group": "inline"
```
scnwwu: I would suggest to remove this line. It's not a primary command to be the first one on the hover cell toolbar.

client/src/node/extension.ts
```diff
@@ -202,6 +202,7 @@ export function activate(context: ExtensionContext) {
     commands.registerCommand("SAS.notebook.export", () =>
       exportNotebook(client),
     ),
+    commands.registerCommand("SAS.notebook.saveOutput", () => saveOutput()),
```
scnwwu: nit: unnecessary to wrap a function
```suggestion
    commands.registerCommand("SAS.notebook.saveOutput", saveOutput),
```

client/src/components/notebook/exporters/toSAS.ts
```diff
@@ -2,11 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 import { NotebookCell, NotebookDocument } from "vscode";
 
-export const exportToSAS = (notebook: NotebookDocument) =>
-  notebook
+export const exportToSAS = (notebook: NotebookDocument) => {
```
scnwwu: nit: unnecessary to change this file

client/src/components/notebook/exporters/index.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright © 2025, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import { Uri, window, workspace } from "vscode";
+import * as vscode from "vscode";
```
scnwwu: there's already a `vscode` import above. looks like you just need `l10n`, add `l10n` above seems simpler

package.json
```diff
@@ -1097,6 +1103,13 @@
           "command": "SAS.notebook.export"
         }
       ],
+      "notebook/cell/title": [
+        {
+          "command": "SAS.notebook.saveOutput",
+          "when": "notebookType == 'sas-notebook'",
```
scnwwu:
```suggestion
          "when": "notebookType == 'sas-notebook' && notebookCellHasOutputs"
```

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +31,134 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+    return;
+  }
+
+  const cell = notebook.cellAt(activeCell);
+  if (!cell) {
+    return;
+  }
+
+  if (cell.outputs.length === 0) {
+    window.showWarningMessage(
```
scnwwu: We should NOT display the menu entry when there's no output. I've added a suggestion in `package.json`
Then I think we don't need this message

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +31,134 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+    output.items.some((item) => item.mime === "application/vnd.sas.ods.html5"),
+  );
+  const hasLogOutput = cell.outputs.some((output) =>
+    output.items.some(
+      (item) => item.mime === "application/vnd.sas.compute.log.lines",
+    ),
+  );
+
+  if (!hasOdsOutput && !hasLogOutput) {
+    window.showWarningMessage(
```
scnwwu: Given this is a SAS Notebook and has output, I think it's unnecessary to have this message.

package.json
```diff
@@ -825,6 +825,12 @@
         "title": "%commands.SAS.notebook.export%",
         "category": "SAS Notebook"
       },
+      {
+        "command": "SAS.notebook.saveOutput",
+        "title": "%commands.SAS.notebook.saveOutput%",
+        "category": "SAS Notebook",
+        "icon": "$(export)"
```
scnwwu: nit: After remove the `"group": "inline"`, the icon seems unnecessary

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +31,134 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+  }
+
+  if (cell.outputs.length === 0) {
+    window.showWarningMessage(
+      vscode.l10n.t("Selected cell has no output to download."),
+    );
+    return;
+  }
+
+  const hasOdsOutput = cell.outputs.some((output) =>
```
scnwwu: unnecessary to do `some` separately. Simply do `find` here and you don't need to `find` later

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +31,134 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+  const choices: Array<{
+    label: string;
+    description: string;
+    detail: string;
+    outputType: "html" | "log";
+  }> = [];
+
+  if (hasOdsOutput) {
+    choices.push({
+      label: "Download as HTML",
```
scnwwu: all user-facing strings need `l10n`
currently the UI looks a little crowded to me. The label and description just says the same thing with different words. We should try best to reduce the user's cognitive load. SAS users understand what's ODS. Technically it may not really contains charts or visualizations.
Since the quick pick placeholder already asks "Choose the output type to save", perhaps simply a label like "ODS HTML" and "SAS Log" would be enough.

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +31,134 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+    choices.push({
+      label: "Download as Log",
+      description: "Save execution log",
+      detail: "Text file with SAS log messages",
+      outputType: "log",
+    });
+  }
+
+  const exportChoice = await window.showQuickPick(choices, {
+    placeHolder: "Choose output type to download",
```
scnwwu: If we say "Save Output" for this feature, let's always say "save" so it's clear it's refering to exact the same thing.
```suggestion
    placeHolder: "Choose the output type to save",
```

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +31,134 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+      }
+    }
+    fileExtension = "log";
+    fileName = `${path.basename(notebook.uri.path, ".sasnb")}_cell_${
+      activeCell + 1
+    }_output.log`;
+  }
+
+  if (!content) {
+    window.showErrorMessage(vscode.l10n.t("Failed to extract output content."));
```
scnwwu: nit: looks like too many checks. there's no try catch to the "extract". Looks like the only case entering here is when no output which checked before

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +31,134 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+  });
+
+  if (!uri) {
+    return;
+  }
+
+  await workspace.fs.writeFile(uri, Buffer.from(content));
+
+  window.showInformationMessage(
+    vscode.l10n.t("Output downloaded to {0}", uri.fsPath),
```
scnwwu: If we say "Save Output" for this feature, let's always say "save"

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +30,114 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+
+    if (odsItem && logItem) {
+      break;
+    }
+  }
+
+  const choices: Array<{
+    label: string;
+    description: string;
+    detail: string;
```
scnwwu: nit: unnecessary to declare these 2 fields

client/src/components/notebook/exporters/index.ts
```diff
@@ -30,3 +30,114 @@ export const exportNotebook = async (client: LanguageClient) => {
...
+    outputType: "html" | "log";
+  }> = [];
+
+  if (odsItem) {
+    choices.push({
+      label: l10n.t("Download as ODS HTML"),
```
scnwwu: As my previous comment I suggest to not say Download, for consistency. I think simply "ODS HTML" would be fine here as it's under the "Choose output type to save" question. If you prefer a verb, I would suggest say Save.

scnwwu (approved):

PR #1582, opened 2025-07-31
Magic Command Support

kpatl1:
**Summary**
Added a MagicCommandProcessor and LanguageSwitcher. This is used in the controller file to allow for magic command support within a notebook. A user can type "%sas, %py, %md" and automatically have the cell change to that type. There are also other aliases that allow for the same changes.
**Testing**
Created a notebook and tested changing the first line of cells back and forth from different types and with different text.


scnwwu:
> I understand. Is there a version of this feature that we should try and implement, or leave as is?

adding @snlwih

PR #1597, opened 2025-08-14
feat(ContentNavigator): added more detailed path information

kpatl1:
**Summary**

This change improves the user experience in the SAS Content Navigator by showing full hierarchical paths when hovering over files and folders, instead of just displaying the filename.

### Key Changes:
1. **Modified `ContentDataProvider.getTreeItem()`** to fetch and display full path information in tooltips
2. **Added fallback behavior** - if full path retrieval fails, tooltips gracefully degrade to showing just the item name

### Before:
- Hovering over a file showed: `/test1.sas`
### After:
- Hovering over a file shows: `/My Folder/Subfolder/test1.sas` 

---

## **Testing**
-Verified tooltips show full paths for files in nested folder structures

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -213,6 +213,15 @@ class ContentDataProvider
   public async getTreeItem(item: ContentItem): Promise<TreeItem> {
     const isContainer = getIsContainer(item);
     const uri = await this.model.getUri(item, false);
+    let tooltip = item.name; // fallback to item name
+    try {
+      const fullPath = await this.model.getPathOfItem(item);
+      if (fullPath) {
+        tooltip = fullPath;
+      }
+    } catch {
+      // If getPathOfItem fails, tooltip will remain as item.name
+    }
 
     return {
       collapsibleState: isContainer
@@ -230,6 +239,7 @@ class ContentDataProvider
       id: item.uid,
       label: item.name,
       resourceUri: uri,
+      tooltip: tooltip,
     };
   }
 


scottdover:
Thinking about it further, you can just check if the path can be copied _before_ you show the tooltip (i.e. checking `item.contextValue.includes('copyPath')`)

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -735,6 +752,30 @@ class ContentDataProvider
         }
       : undefined;
   }
+
+  private async getCachedParentPath(item: ContentItem): Promise<string> {
```
scottdover: Since we're really using this to get tooltip information, it might be worth updating this to be `getItemTooltip` and calling `tooltip: this.getItemTooltip(item)` up above

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -735,6 +752,30 @@ class ContentDataProvider
...
+    const parentUri = item.parentFolderUri;
+    if (!parentUri) {
+      return "";
+    }
+
+    if (!this.parentPathCache.has(parentUri)) {
+      try {
+        const fullPath = await this.model.getPathOfItem(item);
```
scottdover: It may also be worth updating `RestContentAdapter.getPathOfItem` to handle its own caching since it's really the only `ContentAdapter` that requires it. And doing that w/n RestContentAdapter may make things easier w/r/t Wei's comment about using `uri`

kpatl1:
This makes sense. I have changed it so the Uri now includes a full path and it can be used in all elements of VSC. Tested it with same named files and it resolved well. Also, moved caching to the RestContentAdapter.

client/src/connection/rest/util.ts
```diff
@@ -38,12 +38,18 @@ export const getResourceIdFromItem = (item: ContentItem): string | null => {
   return getLink(item.links, "GET", "self")?.uri || null;
 };
 
-export const getSasContentUri = (item: ContentItem, readOnly?: boolean): Uri =>
+export const getSasContentUri = (
```
scottdover, line 57: Instead of having a nested ternary here, could we just _always_ pass in `fullPath`. So the function would become something like
```
getSasContentUri(item: ContentItem, path: string, readOnly?: boolean)
```
and we'd just _always_ use path

client/src/connection/rest/RestContentAdapter.ts
```diff
@@ -142,19 +150,27 @@ class RestContentAdapter implements ContentAdapter {
...
+            ...this.enrichWithDataProviderProperties(childItem, {
+              isInRecycleBin,
+              isInMyFavorites: parentIdIsFavoritesFolder || !!favoriteUri,
+              favoriteUri,
+            }),
+          };
+
+          if (enrichedItem.contextValue?.includes("copyPath")) {
+            await this.updateUriWithFullPath(enrichedItem);
+          }
```
scottdover: It looks like the only time we _don't_ call `updateUriWithFullPath` is when...
 - getting root items
 - moving items

I'd advocate for moving this call inside `enrichWithDataProviderProperties` (I know that make things a _tiny_ bit more complicated since this function doesn't return a promise)

client/src/connection/rest/RestContentAdapter.ts
```diff
@@ -186,21 +202,59 @@ class RestContentAdapter implements ContentAdapter {
     if (!item) {
       return "";
     }
+    const baseKey = item.uri || item.id || item.name;
```
scottdover: This seems like a lot. I'd probably update things such that we're setting the values here, and updating in calling functions where we know the full path might change (rename, move, delete, restore). So the general outline might be something like...
```
// In a function that changes the value
this.pathCache.set(item.id, await this.calculatePathOfItem(item));
```
then the implementation would use the id for the cache key since that's mostly unchanging...
```
if (this.pathCache.has(item.id)) {
      return this.pathCache.get(item.id)!;
    }

    try {
      const path = await this.calculatePathOfItem(item, folderPathOnly);
      this.pathCache.set(item.id, path);
      return path;
    } catch {
      return "";
    }
```

Would something like this work?

client/src/connection/rest/util.ts
kpatl1: Would this not cause some of the times that this method is called to change into an async function because we would have to get the path of the item? I am just unsure why we cannot keep it as an optional parameter?

client/src/connection/rest/RestContentAdapter.ts
kpatl1: I agree. I will go ahead and refactor to allow for this.

kpatl1 (commented):
> > Made a new commit by changing the uri to contain the full path. I have also moved the caching into the RestContentAdapter. I cannot seem to get the content in SAS Server to show the full path, however, only when I view items in My Folder. Do you guys have any advice on how I could approach this?
> 
> Fwiw, sas server uris have a different structure and call `getSasServerUri`. But maybe I'm misunderstanding what you're saying 

I understand. I think my confusion came from poor wording. When you hover over a file under the SAS Content folder, it does not resolve a path only the item.name. However, for files in My Folder it works, so would I need to utilize getSasServerUri for those files?

scottdover:
Looked over this a bit more. Here's what I'm noticing:
 - The tooltips start out with `/Users/...`. That seems like a step in the right direction
 - We're still setting `tooltip` in `ContentDataProvider`. Looking at @scnwwu 's comments up above, I don't think we want to do that. Instead, the uri should contain the full path so that the tooltip is automatically populated correctly. At the moment, hovering over the tab on an open file doesn't show the full path
 - Also, the filename statement (where you drag a file from sas content into a sas file while holding shift) generation isn't working as expected. The folder path now contains the file, like...
 ```sas
filename fileref filesrvc folderpath='/Users/scdove/My Folder/test.sas' filename='test.sas';
```

I think the best path forward is to get rid of `tooltip` in `ContentDataProvider`, and make sure the uris for both sas content (RestContentAdapter) and sas server (RestServerAdapter, ItcServerAdapter) contain the full path (_or_ just the name in instances where the path isn't relevant/copyable/etc)

File: client/src/connection/rest/util.ts (hunk 3 above)
Comment (scottdover) [reply client/src/connection/rest/util.ts:57]: I feel like an earlier version of your changes was using a nested ternary, which seemed unnecessary and made the code a bit harder to follow. I like how you cleaned this up.

File: client/src/connection/rest/util.ts [hunk 6]
```diff
@@ -38,23 +38,48 @@ export const getResourceIdFromItem = (item: ContentItem): string | null => {
...
+  // Decode SAS server path encoding (~fs~ represents /)
+  pathToUse = pathToUse.replace(/~fs~/g, "/");
+
+  // Remove leading slash if present since we'll add one in the URI construction
+  if (pathToUse.startsWith("/")) {
+    pathToUse = pathToUse.substring(1);
+  }
+
+  // URL encode special characters
+  pathToUse = pathToUse.replace(/#/g, "%23").replace(/\?/g, "%3F");
```
scottdover: Could you move this into a shared function and call it from `getSasContentUri` and `getSasServerUri`. You could pass in a "replacements" regex to handle the difference in paths
```
pathToUse = replacement ? pathToUse.replace(replacement, "/") : pathToUse;
```

client/src/connection/rest/RestContentAdapter.ts
```diff
@@ -187,30 +193,96 @@ class RestContentAdapter implements ContentAdapter {
...
+        return parentPath;
+      } catch {
+        return "";
+      }
+    }
+    const filePathParts = [];
+    let currentContentItem: Pick<
+      ContentItem,
+      "parentFolderUri" | "name" | "id"
+    > = item;
```
scottdover: Could this just be...
```suggestion
    let currentContentItem: ContentItem = item;
```

scottdover (commented):
Only a couple minor things and I think this one is good to go. I'm not 100% on fully understanding the path cache code, but it seems to work alright

PR #1606, opened 2025-08-19
feat: add test coverage

scottdover:
**Summary**
This revives @smorrisj's previous work around using c8 for test coverage. This is still a wip

**Testing**
- [x] Made sure tests passed w/o failure
- [x] Made sure what was represented in coverage html is what is expected

### DIFF
--- a/.c8rc
+++ b/.c8rc
@@ -3,16 +3,18 @@
   "src": ["client/src", "server/src"],
   "exclude": [
     "**/node_modules/**",
-    "*.test.*",
+    "**/*.test.*",
     "client/dist",
+    "client/src/connection/rest/api",
     "server/dist",
-    "client/src/connection/rest/api"
+    "tools",
+    "server/src/python"
   ],
   "reporter": ["html", "text"],
   "excludeAfterRemap": true,
   "checkCoverage": true,
-  "lines": 57,
-  "statements": 57,
-  "branches": 75,
-  "functions": 44
+  "lines": 52,
+  "statements": 52,
+  "branches": 71,
+  "functions": 43
 }
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -20,6 +20,5 @@ ContributorAgreement.txt
 eslint.config.mjs
 pull_request_template.md
 l10n/bundle.l10n.json
-.eslintrc.js
 .c8rc
 coverage/


.c8rc
```diff
@@ -0,0 +1,20 @@
...
+  "all": true,
+  "src": ["client/src", "server/src"],
+  "exclude": [
+    "**/node_modules/**",
+    "**/*.test.*",
+    "client/dist",
+    "client/src/connection/rest/api",
+    "server/dist",
+    "tools",
+    "server/src/python"
```
scottdover, line 11: `server/src/python` is copied from node_modules as part of the build process

.c8rc
```diff
@@ -0,0 +1,20 @@
+{
+  "all": true,
+  "src": ["client/src", "server/src"],
+  "exclude": [
+    "**/node_modules/**",
+    "**/*.test.*",
+    "client/dist",
+    "client/src/connection/rest/api",
```
scottdover, line 8: These are files auto-generated by OpenAPI

.c8rc
```diff
@@ -0,0 +1,20 @@
...
+    "tools",
+    "server/src/python"
+  ],
+  "reporter": ["html", "text"],
+  "excludeAfterRemap": true,
+  "checkCoverage": true,
+  "lines": 52,
+  "statements": 52,
+  "branches": 71,
+  "functions": 43
```
scottdover, line 19: What do you think about adding minimum levels now _just_ to make sure things don't get worse (the numbers above are right below the _current_ values), with the eventual goal being having code coverage at 80%. 

And, do we care about these individual numbers, or do we really only care about lines (I'm inclined to believe this should probably _only_ be "lines": 80)? 

Last, is 80% code coverage reasonable in an open source repository, or would that be asking too much from contributors?

.c8rc
scnwwu: sorry this directory looks not copied from node_modules to me. where did you see the copy?

.c8rc
```diff
@@ -0,0 +1,20 @@
+{
+  "all": true,
+  "src": ["client/src", "server/src"],
+  "exclude": [
+    "**/node_modules/**",
+    "**/*.test.*",
```
scnwwu, line 6: I see `client/test` in the coverage report. I guess we should exclude it.

.c8rc
scnwwu: I would vote to not fail the build until we're fully ready. I guess we're not setting up testing infrastructure to the webviews. So if lines added in webviews and make the numbers worse it's not straightforward to fix.

Issue #1566, opened 2025-07-21
Allow users to define custom mapping when "(auto)" is specified for SAS.results.html.style


smorrisj:
### Is your feature request related to a problem? Please describe.

Currently, a user can specify "(auto)" value for `SAS.results.html.style` and the extension will attempt to [map](https://github.com/sassoftware/vscode-sas-extension/blob/e242d1f4c5200c3fc36c05026688cfca20e0226c/client/src/components/utils/SASCodeDocumentHelper.ts#L108) the existing vscode theme to an ODS style that resembles the VSCode theme for consistency. It would be a nice feature to also let users override this mapping and define their own custom mapping for when "(auto)" gets used. This would be useful in scenarios where users might want to extend existing ODS themes, for example overriding `HtmlEncore` to have larger font sizes for tables, where this extended theme could be matched with a given VSCode color theme to extend functionality of the feature.

### Describe the solution you'd like

As a user, I'd like the ability to tailor how the "(auto)" feature maps a VSCode color theme over to a SAS ODS theme so that custom themes could be defined in the mapping scheme. I'd like to define this mapping as a JSON object in the user settings JSON file. If the object key is present, and "(auto)" is specified for `SAS.results.html.style`, then I would like for the specified mapping to be applied when generating the ODS output statement in the preamble. If the object key is NOT present and "(auto)" is specified for `SAS.results.html.style`, then I'd like for the extension to infer the mapping based on the currently defined mapping scheme in the extension code.

PR #1618, opened 2025-08-26
feat: allow users to define custom mapping

tarikutk:
**Summary**
Allows users to use customized style when SAS.results.html.style is auto. for example to use customized ignite, user defines SAS.results.html.styleMapping in settings as an object like this:

"SAS.results.html.styleMapping": {
    "ignite": "customizedIgnite"
  }

customizedIgnite style should be defined before it is added to the settings. for example:

proc template;
  define style styles.customizedIgnite;
    parent = styles.ignite;
    style body /
      backgroundcolor = red;
  end;
run;

in this case whenever 'SAS dark' (ignite) is selected as a theme, the customizedIgnite will be applied in each ODS html display.

Fix: #1566

### DIFF
--- a/client/src/components/utils/SASCodeDocumentHelper.ts
+++ b/client/src/components/utils/SASCodeDocumentHelper.ts
@@ -8,6 +8,7 @@ import {
   TextDocument,
   commands,
   window,
+  workspace,
 } from "vscode";
 
 import { v4 } from "uuid";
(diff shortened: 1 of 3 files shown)


tarikutk:
@smorrisj , could you check if this is the right feature you asked please ? Thanks

PR #1619, opened 2025-08-26
feat: allow users to define custom mapping

tarikutk:
**Summary**
Allows users to use customized style when SAS.results.html.style is auto. for example to use customized ignite, user defines SAS.results.html.styleMapping in settings as an object like this:

"SAS.results.html.styleMapping": {
"ignite": "customizedIgnite"
}

customizedIgnite style should be defined before it is added to the settings. for example:

proc template;
define style styles.customizedIgnite;
parent = styles.ignite;
style body /
backgroundcolor = red;
end;
run;

in this case whenever 'SAS dark' (ignite) is selected as a theme, the customizedIgnite will be applied in each ODS html display.

Fix https://github.com/sassoftware/vscode-sas-extension/issues/1566

### DIFF
--- a/client/src/components/utils/SASCodeDocumentHelper.ts
+++ b/client/src/components/utils/SASCodeDocumentHelper.ts
@@ -8,6 +8,7 @@ import {
   TextDocument,
   commands,
   window,
+  workspace,
 } from "vscode";
 
 import { v4 } from "uuid";
@@ -103,26 +104,55 @@ function selectionsAreNotEmpty(
 
 function getHtmlStyleValue(): string {
   const htmlStyleSetting = getHtmlStyle();
+  const isValidStyle = (style: string) => /^[a-zA-Z0-9_]+$/.test(style);
 
-  switch (htmlStyleSetting) {
-    case "(auto)":
-      switch (window.activeColorTheme.kind) {
-        case ColorThemeKind.Light:
-          return "Illuminate";
-        case ColorThemeKind.Dark:
-          return "Ignite";
-        case ColorThemeKind.HighContrast:
-          return "HighContrast";
-        case ColorThemeKind.HighContrastLight:
-          return "Illuminate";
-        default:
-          return "";
-      }
-    case "(server default)":
-      return "";
-    default:
-      return htmlStyleSetting;
+  let styleValue = "";
+  if (htmlStyleSetting === "(auto)") {
+    // get the styleMapping object from user settings
+    const styleMapping: {
+      illuminate?: string;
+      ignite?: string;
+      highContrast?: string;
+    } =
+      window.activeColorTheme &&
+      typeof window.activeColorTheme.kind === "number"
+        ? workspace.getConfiguration("SAS").get<{
+            illuminate?: string;
+            ignite?: string;
+            highContrast?: string;
+          }>("results.html.styleMapping") || {}
+        : {};
+
+    switch (window.activeColorTheme.kind) {
+      case ColorThemeKind.Light:
+        styleValue = styleMapping.illuminate || "Illuminate";
+        break;
+      case ColorThemeKind.Dark:
+        styleValue = styleMapping.ignite || "Ignite";
+        break;
+      case ColorThemeKind.HighContrast:
+        styleValue = styleMapping.highContrast || "HighContrast";
+        break;
+      case ColorThemeKind.HighContrastLight:
+        styleValue = styleMapping.illuminate || "Illuminate";
+        break;
+      default:
+        styleValue = "";
+    }
+  } else if (htmlStyleSetting === "(server default)") {
+    styleValue = "";
+  } else {
+    styleValue = htmlStyleSetting;
   }
+
+  // Validate style value
+  if (styleValue && !isValidStyle(styleValue)) {
+    console.warn(
+      `Invalid ODS style name: ${styleValue}. Only a-z, A-Z, 0-9, and _ are allowed.`,
+    );
+    return "";
+  }
+  return styleValue;
 }
 
 // if no valid selection, return whole text as only selection
--- a/package.json
+++ b/package.json
@@ -510,6 +510,25 @@
             ],
             "description": "%configuration.SAS.results.html.style%"
           },
+          "SAS.results.html.styleMapping": {
+            "type": "object",
+            "default": {},
+            "description": "%configuration.SAS.results.html.styleMapping%",
+            "properties": {
+              "illuminate": {
+                "type": "string",
+                "description": "Style to use for illuminate theme"
+              },
+              "ignite": {
+                "type": "string",
+                "description": "Style to use for ignite theme"
+              },
+              "highContrast": {
+                "type": "string",
+                "description": "Style to use for high contrast theme"
+              }
+            }
+          },
           "SAS.problems.log.enabled": {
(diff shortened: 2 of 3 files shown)


client/src/components/utils/SASCodeDocumentHelper.ts
```diff
@@ -103,26 +104,55 @@ function selectionsAreNotEmpty(
 
 function getHtmlStyleValue(): string {
   const htmlStyleSetting = getHtmlStyle();
+  const isValidStyle = (style: string) => /^[a-zA-Z0-9_]+$/.test(style);
```
scnwwu: I tried define a style with a namelit and it appears to work.
```
proc template;
define style styles.'中文'n;
parent = styles.ignite;
style body /
backgroundcolor = red;
end;
run;
```
Is it necessary to check it? If a invalid style, user can see errors from SAS log.
Current implementation hide the error and fallback to server default. If user doesn't see console they won't know there's an error and may think this feature doesn't work.

client/src/components/utils/SASCodeDocumentHelper.ts
```diff
@@ -103,26 +104,55 @@ function selectionsAreNotEmpty(
...
+  let styleValue = "";
+  if (htmlStyleSetting === "(auto)") {
+    // get the styleMapping object from user settings
+    const styleMapping: {
+      illuminate?: string;
+      ignite?: string;
+      highContrast?: string;
+    } =
+      window.activeColorTheme &&
+      typeof window.activeColorTheme.kind === "number"
```
scnwwu: any reason to check this?

package.json
```diff
@@ -510,6 +510,25 @@
             ],
             "description": "%configuration.SAS.results.html.style%"
           },
+          "SAS.results.html.styleMapping": {
+            "type": "object",
+            "default": {},
+            "description": "%configuration.SAS.results.html.styleMapping%",
+            "properties": {
+              "illuminate": {
```
scnwwu: There's no this word "Illuminate" in VS Code UI. We just by default map Light theme kind to "Illuminate".
So I guess the key should be "light".

client/src/components/utils/SASCodeDocumentHelper.ts
tarikutk: no. it is not necessary. I will remove it.

client/src/components/utils/SASCodeDocumentHelper.ts
tarikutk: it was generated by copilot. I will remove that.

package.json
tarikutk: oh. okay. it is because I used them directly. illuminate, ignite and hightContrast. I think I can update them to light, dark, and highContrast.

tarikutk: those may also give users a hint which style to extend to when they create the custom style.

client/test/components/util/SASCodeDocumentHelper.test.ts
```diff
@@ -0,0 +1,118 @@
+import * as vscode from "vscode";
+
+import { assert } from "chai";
+import sinon from "sinon";
+
+import { profileConfig } from "../../../src/commands/profile";
+import { ConnectionType } from "../../../src/components/profile";
+import { getCodeDocumentConstructionParameters } from "../../../src/components/utils/SASCodeDocumentHelper";
+
+describe("getCodeDocumentConstructionParameters", () => {
```
smorrisj: Looks really good. Thanks for adding this. One small nit: Can we have a top level describe with a friendly name? Maybe something like "sas code document helper"? Then we could put this describe inside of that one. It will make the test reports read a bit nicer. So something like:

```typescript
describe("sas code document helper", () => {
    describe("getCodeDocumentConstructionParameters", () => {
        // ..test code...
    });
});
```

tarikutk: I updated the description. Thanks

package.json
```diff
@@ -510,6 +510,29 @@
...
             "description": "%configuration.SAS.results.html.style%"
           },
+          "SAS.results.html.custom.style": {
+            "type": "object",
+            "default": {},
+            "description": "%configuration.SAS.results.html.custom.style%",
+            "properties": {
+              "light": {
+                "type": "string",
+                "description": "Style to use for light theme"
```
scnwwu: every user-facing strings need to be externalized. Please refer to other descriptions

client/src/components/utils/SASCodeDocumentHelper.ts
```diff
@@ -103,26 +104,44 @@ function selectionsAreNotEmpty(
 
 function getHtmlStyleValue(): string {
   const htmlStyleSetting = getHtmlStyle();
+  let styleValue = "";
+  if (htmlStyleSetting === "(auto)") {
+    // get the results.html.custom.style object from user settings
+    const customStyle: {
```
scnwwu: nit: unnecessary to declare the type twice. The type can be infered from the `get` method.

client/src/components/utils/SASCodeDocumentHelper.ts
```diff
@@ -103,26 +104,44 @@ function selectionsAreNotEmpty(
 
 function getHtmlStyleValue(): string {
   const htmlStyleSetting = getHtmlStyle();
+  let styleValue = "";
```
scnwwu: nit: It looks like there's no operation on the `styleValue` variable. I would prefer the previous directly-`return` style.

package.nls.json
```diff
@@ -71,6 +71,7 @@
   "configuration.SAS.results.html.style": "Specifies the style for ODS HTML5 results.",
   "configuration.SAS.results.html.style.(auto)": "Let the extension pick a style that most closely matches the color theme.",
   "configuration.SAS.results.html.style.(server default)": "Default to the style configured on the SAS server.",
+  "configuration.SAS.results.html.custom.style": "Map editor color themes to ODS HTML5 styles. Specify an object with properties like 'light', 'dark', 'highContrast', and 'highContrastLight' to control which style is used for each theme.",
```
scnwwu: I'm just thinking shall we mention something like "Takes effect when `SAS.results.html.style` is `(auto)`" here.
If the sentence becomes longer, perhaps the mention of property names can be removed. The schema in the `package.json` file will produce autocomplete when user editing the setting.

package.nls.json
```diff
@@ -71,6 +71,11 @@
   "configuration.SAS.results.html.style": "Specifies the style for ODS HTML5 results.",
   "configuration.SAS.results.html.style.(auto)": "Let the extension pick a style that most closely matches the color theme.",
   "configuration.SAS.results.html.style.(server default)": "Default to the style configured on the SAS server.",
+  "configuration.SAS.results.html.custom.style": "Specify which SAS style to use for the light, dark, highContrast, or highContrastLight themes. This option takes effect when SAS.results.html.style is (auto).",
```
scnwwu: nit: Please refer to 
https://code.visualstudio.com/api/references/contribution-points#contributes.configuration
"Linking to settings" section.
It will make it look better

smorrisj (approved):

ShermanWang0324:
Tested results: success to apply custom style to change the color and font.
Screenshot 2025-09-08 152238

PR #1622, opened 2025-08-29
feat: implement column sorting

scottdover:
**Summary:**
This updates the data viewer code to support sorting columns. 

On the backend, this is supported by passing a new `sortModel` to our `getRows` function. ITC
connections support this by creating an order by statement, but REST connections are a bit more
complicated. With a non-empty sort model, the general process for REST is to:
 - Create a sorted view
 - Query that view and get results
 - Delete the view

This is consistent with how studio sorts data.

On the front-end, a `ColumnMenu` has been introduced to facilitate adding and removing sorting
criteria. 

Last, these changes required exposing l10n messages to the `ColumnMenu` component. To do this, a
few changes were made to `WebView`:
 - `WebView` is now responsible for rendering it's content
 - `WebView.render` makes use of some new class methods to generate content:
    - `body`: returns the html that should show up in the `<body>` section of a webview
    - `scripts` (optional): An array of relative script sources to load for the webview
    - `styles` (optional): An array of relative style sources to load for the webview
    - `l10nMessages` (optional): A map of l10n translation messages to expose to React
      (NOTE: These messages can be read using the new `localize` helper)

The changes to `WebView` should make future webview panels work in a more consistent way.

**Testing:**
- [x] Test sorting a single column
- [x] Test sorting multiple columns (test sorting 3 or more columns and deleting a "middle" sort
  to make sure sort indexing is maintained)
- [x] Make sure sorting columns pops the grid back to the first row
- [x] Test sorting for REST and ITC-based connections
- [x] Make sure the column menu displays as expected
  - [x] If the parent menu and child menu fit left to right, display them as such
  - [x] If the parent menu fits in the view, but the child menu _doesn't_, the child menu
    should be displayed to the left
  - [x] If neither the parent menu _not_ the child menu fit, the parent menu should be adjusted
    to be near the right edge of the screen, and the child menu should be displayed to the left
- [x] Make sure column menu can be dismissed by clicking outside of the menu, or by pressing `esc`
- [x] Test opening properties view for a table, and opening properties view for a column within a table

### DIFF
--- a/client/src/webview/DataViewer.css
+++ b/client/src/webview/DataViewer.css
@@ -102,14 +102,9 @@ button,
   height: 16px;
 }
 
-.dropdown button span {
+.dropdown button {
   background: url(../../../icons/light/more.svg);
   background-size: 16px 16px;
-}
-
-.dropdown button:hover,
-.dropdown.active button {
-  background: var(--vscode-toolbar-hoverBackground);
 }
 
 .vscode-dark .dropdown button {


scnwwu:
Workbench goes through REST but no studio

scnwwu:
Looks like `getRowSet` supports a `sortBy`, will it work?
https://developer.sas.com/rest-apis/compute/getRowSet

client/src/panels/WebviewManager.ts
```diff
@@ -25,11 +25,71 @@ export abstract class WebView {
...
+          }
+        </body>
+      </html>`;
+
+    return this;
+  }
+
   abstract processMessage(event: Event): void;
 
   public withPanel(webviewPanel: WebviewPanel): WebView {
```
kpatl1, line 95: This overall looks good for the webview and dataviewer, I think definitely makes the code cleaner. One thing is maybe just make sure the panel/webview and webviewUri are always set before render() is called (call render only after withPanel()).

client/src/connection/itc/script.ts
```diff
@@ -244,17 +244,21 @@ class SASRunner{
...
+  [void]GetDatasetRecords([string]$tableName, [int]$start = 0, [int]$limit = 100, [string]$sortCriteria = "") {
     $objRecordSet = New-Object -comobject ADODB.Recordset
     $objRecordSet.ActiveConnection = $this.dataConnection # This is needed to set the properties for sas formats.
     $objRecordSet.Properties.Item("SAS Formats").Value = "_ALL_"
 
+    $query = "SELECT * FROM $tableName"
+    if ($sortCriteria -ne "") {
+      $query = $query + " ORDER BY " + $sortCriteria
+    }
     $objRecordSet.Open(
```
scnwwu, line 261: It looks breaking the SAS Formats.
If you open the `sashelp.air` table you'll see that the DATE column is displaying wrong.

client/src/webview/useDataViewer.ts
scnwwu, line 1: sorry what's the reason to rename the extension to `.tsx`?
it looks a little uncommon for a hook name (useXXX) with a tsx extension which means rendering markups.

client/src/webview/ColumnHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <span className="sort-icon-wrapper">
+          {!!column.sort && (
+            <>
+              <span className={`icon ${column.sort}`}></span>
+              {!!columnNumber && <span className="number">{columnNumber}</span>}
+            </>
+          )}
+        </span>
+        <div className={dropdownClassname}>
+          <button ref={ref} type="button" onClick={displayColumnMenu}>
```
scnwwu: is there a keyboard-only way to operate the sorting?
When I try keyboard focus on a column header and press Enter, it seems sorting but no icon displayed.

client/src/webview/localize.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright © 2025, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// `localize` provides functionality to use localized strings in react webviews
+// This expects for translations to be defined in a script block with an id of
+// `l10-messages`.
+let localizedTerms = null;
+const localize = (term: string) => {
```
scnwwu: what about parameter substitution

client/src/connection/rest/RestLibraryAdapter.ts
scnwwu, line 76: When I tried it, there seems a problem. When get from a view, the `count` is `undefined`, thus the scrollbar doesn't show enough space.

client/src/connection/itc/script.ts
```diff
@@ -244,17 +244,21 @@ class SASRunner{
     Write-Host "${LineCodes.ResultsFetchedCode}"
   }
 
-  [void]GetDatasetRecords([string]$tableName, [int]$start = 0, [int]$limit = 100) {
+  [void]GetDatasetRecords([string]$tableName, [int]$start = 0, [int]$limit = 100, [string]$sortCriteria = "") {
     $objRecordSet = New-Object -comobject ADODB.Recordset
     $objRecordSet.ActiveConnection = $this.dataConnection # This is needed to set the properties for sas formats.
     $objRecordSet.Properties.Item("SAS Formats").Value = "_ALL_"
 
+    $query = "SELECT * FROM $tableName"
```
smorrisj: For the script changes, would love if we could put some quoting/escaping around $tableName and $sortCriteria?

client/src/webview/useDataViewer.ts
scottdover: Ah, I was inlining some react code for column headers previously. I'll move it back.

client/src/connection/itc/script.ts
smorrisj: Perhaps also validate the table and sorting criteria are valid inputs?

client/src/connection/rest/RestLibraryAdapter.ts
scottdover: Ohh, good catch. I noticed there was some weirdness around loading larger tables while sorting, but thought it was mainly due to the fact that sorting took marginally longer than just displaying the table. Will take a look.

client/src/webview/ColumnHeader.tsx
scottdover: Ah, I did include the ability to escape with keyboard, but not the ability to open the context menu/etc. Will investigate that one further.

client/src/webview/localize.ts
scottdover: I thought about including it, but my code didn't require it, so I opted to keep it simple. But since you asked, I'll include it :)

client/src/panels/WebviewManager.ts
scottdover: I think we're safe here, since we always call `withPanel` _before_ calling `render` in code (and none of the other methods in `WebView` have an expectation that `panel` can be undefined)

client/src/webview/useDataViewer.ts
scottdover: Resolved in https://github.com/sassoftware/vscode-sas-extension/pull/1622/commits/8ed259f8e6bcdd1304d86cd465b81a63d7d15428

client/src/webview/ColumnHeader.tsx
scottdover: This one is partially resolved by https://github.com/sassoftware/vscode-sas-extension/pull/1622/commits/8ed259f8e6bcdd1304d86cd465b81a63d7d15428 .

Now, enter and shift + f10 display the context menu _instead_ of sorting. In a future PR, I'll make improvements for keyboard accessibility overall, as this only _really_ helps us display the context menu, but doesn't focus context menu items/etc

client/src/webview/localize.ts
scottdover: Included in https://github.com/sassoftware/vscode-sas-extension/pull/1622/commits/8ed259f8e6bcdd1304d86cd465b81a63d7d15428

client/src/connection/itc/script.ts
scottdover: Yeah, I'll pick up this part of it on Friday. As Wei mentioned up above, there are some formatting issues with the way we're getting records in ITC-based connections. One method that I'm working with is grabbing format information from the columns. So, I should have them available, and should be able to validate against them 🤞

scottdover:
> It looks like the "SAS Formats" only works well with `adCmdTableDirect` (what main branch is doing). I guess an easy way is to do similar like Viya.
> 
> ```powershell
> $this.dataConnection.Execute("create view work.test as select * from sashelp.air order by air asc")
> $objRecordSet = New-Object -comobject ADODB.Recordset
> $objRecordSet.ActiveConnection = $this.dataConnection # This is needed to set the properties for sas formats.
> $objRecordSet.Properties.Item("SAS Formats").Value = "_ALL_"
> $objRecordSet.Open(
>       "work.test",
>       [System.Reflection.Missing]::Value, # Use the active connection
>       2,  # adOpenDynamic
>       1,  # adLockReadOnly
>       512 # adCmdTableDirect
> )
> $this.dataConnection.Execute("drop view work.test")
> ```

This seems v obvious now. Thanks for the second set of eyes

client/src/connection/rest/RestLibraryAdapter.ts
scottdover: This was resolved in https://github.com/sassoftware/vscode-sas-extension/pull/1622/commits/2a167892ee6c39c2672b39cef577060d146dd5bf. This doesn't _quite_ fix the issue with count being undefined...there wasn't a straightforward way to find that as it isn't returned by a view or by the createView statement. 

However, this _does_ resolve the issue enough (imo) as it stops us from indefinitely trying to load more data by inferring the total count.

client/src/connection/itc/script.ts
scottdover: I addressed this in https://github.com/sassoftware/vscode-sas-extension/pull/1622/commits/c9b4fd96178f507f001c7cc76c00c14c0f1ead9f

client/src/connection/itc/script.ts
scottdover: I looked a bit into this, and opted to not include changes here. At the moment, the user cannot set `$tableName` or `$sortCriteria`. Those are both set w/ some user interaction w/ the grid and not manually edited by the user, so I think we're pretty safe.

client/src/connection/itc/ItcLibraryAdapter.ts
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import { l10n } from "vscode";
 
+import { SortModelItem } from "ag-grid-community";
```
scnwwu: nit: similar to https://github.com/sassoftware/vscode-sas-extension/pull/1631#discussion_r2453699520
I would suggest `import type`. It would look weird if a connection implementation really depends on a UI component (ag-grid).

client/src/connection/itc/script.ts
```diff
@@ -288,10 +296,14 @@ class SASRunner{
...
 
+    if ($sortCriteria -ne "") {
+      $this.dataConnection.Execute("DROP VIEW $tableName")
+    }
+
     Write-Host $(ConvertTo-Json -Depth 10 -InputObject $result -Compress)
   }
 
-  [void]GetColumns([string]$libname, [string]$memname) {
+  [object[]] GetColumnData([string]$libname, [string]$memname) {
```
scnwwu: nit: just curious, what's the reason to split this function?

client/src/connection/rest/RestLibraryAdapter.ts
scnwwu: So the user experience will be inconsistent. When I open a table (no sorting initially), I can see the scrollbar as if all rows are there. I can drag the scrollbar to a position and see empty cells and I know the content is being loaded. And later the content filled in.
Now when I add a sort, the scrollbar will change to like there're only 100 rows. When I scroll to the bottom, it's not obvious if it's really at end or loading.
It would be great if we can provide the count to ag grid. When the table opened initially, no sorting applied, we should already get the count. Can we cache the count?

client/src/panels/WebviewManager.ts
```diff
@@ -25,11 +30,71 @@ export abstract class WebView {
...
+          )}" />
+          ${styles}
+          <title>${this.title}</title>
+        </head>
+        <body>
+          ${this.body()}
+          ${scripts}
+          ${
+            this?.l10nMessages
+              ? `<script type="application/json" id="l10n-messages">${JSON.stringify(this?.l10nMessages())}</script>`
```
scnwwu: nit: it looks like only used as a string. How about make it in a data-* attribute to be clear it's not something executable?

client/src/connection/itc/ItcLibraryAdapter.ts
scottdover: Thanks for clarifying :). Updated in https://github.com/sassoftware/vscode-sas-extension/pull/1622/commits/23610b4348345041994650265921245d4072b9a7

client/src/connection/itc/script.ts
scottdover: It was something I was working towards in a temporary commit and bailed on. I've since reverted this (in https://github.com/sassoftware/vscode-sas-extension/pull/1622/commits/23610b4348345041994650265921245d4072b9a7)

client/src/panels/WebviewManager.ts
scottdover: In https://github.com/sassoftware/vscode-sas-extension/pull/1622/commits/718b3f077e065e71797e49338f3fd9080a0eb31a, I moved this to a data attribute on `body` and got rid of this extra script tag

client/src/connection/rest/RestLibraryAdapter.ts
scottdover: Yeah, I'd rather it not be inconsistent, either. I did consider caching things. That would fix the issue temporarily. However, once we incorporate filtering, I think we're going to run into the same issue and will have to manage this cache differently. Fwiw, it looks like scrollbars also get a little inconsistent in studio when we're applying both a sort and filter.

I'm inclined to not address this one until a user complains, unless @snlwih thinks it is worth exploring further before releasing?

scnwwu (approved):

Zhirong2022:
Drag and drop a column to a new position, it will restore to the initial place when trying to perform sort on a column.
Steps to reproduce:
1.Sign in with a valid Viya connection
2.Open SASHELP.CLASS on libraries pane
3.Drag and drop column Age to the second column
4.Try to perform sort on column Name
columnRestore

Zhirong2022:
UI diff on column name
1.There is no tooltip for type icon, truncated column name, sorting indicator and menu icon
2.Type icon truncation
image

Zhirong2022:
Sorting sequence number changed unexpectedly
Steps to reproduce:
1.Sign in with a valid Viya connection
2.Open SASHELP.AIRLINE on libraries pane
3.Sorting all three columns 'DATE', 'AIR' and 'REGION'
4.Remove sorting on the first column 'DATE'
image

scnwwu:
> Sorting sequence number changed unexpectedly Steps to reproduce: 1.Sign in with a valid Viya connection 2.Open SASHELP.AIRLINE on libraries pane 3.Sorting all three columns 'DATE', 'AIR' and 'REGION' 4.Remove sorting on the first column 'DATE' image

The behavior in SAS Extension looks correct.

scottdover:
NOTE: @dahils found an issue where the menu trigger (...) isn't working in the light theme. Just capturing that here where I don't forget it

dahils (approved):

PR #1624, opened 2025-09-05
fix: fix multiple establish connection calls

scottdover:
**Summary**
This updates our base session class to _wait_ to call `establishConnection` more than once, instead returning the already created promise if we're still waiting on a call to complete.

**Testing**
 - [x] Made sure there was only one call to `contextApi.createSession`

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -9,8 +9,8 @@
       "version": "0.0.1",
       "license": "Apache-2.0",
       "dependencies": {
-        "ag-grid-community": "^34.1.1",
-        "ag-grid-react": "^34.1.1",
+        "ag-grid-community": "^34.1.2",
+        "ag-grid-react": "^34.1.2",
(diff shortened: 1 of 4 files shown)


scottdover:
Hey @clangsmith @Zhirong2022 . I'm looking into this a bit today and again on Friday, but I'm wondering if this should hold up this PR since it seems like a less common use case that isn't _really_ related. This issue tackles a race condition where multiple calls were being made when they shouldn't have been. The bug you're describing has more to do with re-using a compute session (as opposed to creating a new one).

If it's okay with y'all, and there are no further bugs, I suggest merging this and working on the other part of this bug in a separate PR

scottdover:
Hey @Zhirong2022 / @clangsmith . I created https://github.com/sassoftware/vscode-sas-extension/issues/1641 to track the second part of this issue

PR #1631, opened 2025-09-12
feat(LibraryNavigator): added ability to view dataset properties

kpatl1:
**Summary**
This PR implements dataset properties functionality for the VSCode extension, bringing feature parity with SasStudio online version. Users can now right-click on datasets in the Library Navigator and view comprehensive metadata including creation dates, modification dates, dataset labels, compression information, and detailed column information.

**Testing**
Went into SASStudio and viewed properties for identical datasets and made sure that the information matched. 

Fixes https://github.com/sassoftware/vscode-sas-extension/issues/1537

### DIFF
--- a/client/src/connection/itc/ItcLibraryAdapter.ts
+++ b/client/src/connection/itc/ItcLibraryAdapter.ts
@@ -11,7 +11,7 @@ import {
   TableData,
   TableRow,
 } from "../../components/LibraryNavigator/types";
-import type { ColumnCollection, TableInfo } from "../rest/api/compute";
+import { ColumnCollection, TableInfo } from "../rest/api/compute";
 import { getColumnIconType } from "../util";
 import { executeRawCode, runCode } from "./CodeRunner";
 import { Config } from "./types";


client/src/connection/itc/ItcLibraryAdapter.ts
```diff
@@ -11,7 +11,7 @@ import {
   TableData,
   TableRow,
 } from "../../components/LibraryNavigator/types";
-import { ColumnCollection } from "../rest/api/compute";
+import { ColumnCollection, TableInfo } from "../rest/api/compute";
```
scnwwu: nit: to make me feel safer 😄 ITC won't really rely on Rest, but just types
```suggestion
import type { ColumnCollection, TableInfo } from "../rest/api/compute";
```

client/src/panels/TablePropertiesViewer.ts
```diff
@@ -0,0 +1,310 @@
...
+    // No messages to process for this static viewer
+  }
+
+  public getContent(): string {
+    return `
+      <!DOCTYPE html>
+      <html lang="en">
+      <head>
+          <meta charset="UTF-8">
+          <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
scnwwu, line 42: Would you follow the CSP setup like others?
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/panels/DataViewer.ts#L43

client/src/panels/TablePropertiesViewer.ts
```diff
@@ -0,0 +1,310 @@
...
+      }
+      try {
+        return new Date(String(value)).toLocaleString();
+      } catch {
+        return String(value);
+      }
+    };
+
+    return `
+      <div class="section-title">General Information</div>
```
scnwwu: there're many hard coded strings in this file. Any chance to get them translatable?

scnwwu:
Overall looks good to me. Will defer to @scottdover to approve. Thanks.

client/src/components/LibraryNavigator/LibraryModel.ts
```diff
@@ -126,6 +126,14 @@ class LibraryModel {
     }
   }
 
+  public async getTableInfo(item: LibraryItem) {
+    await this.libraryAdapter.setup();
```
scottdover: Might be better to nest this inside the `if` below

client/src/connection/itc/ItcLibraryAdapter.ts
```diff
@@ -196,6 +196,57 @@ class ItcLibraryAdapter implements LibraryAdapter {
...
+        rowCount: tableInfo.rowCount || 0,
+        columnCount: tableInfo.columnCount || 0,
+        logicalRecordCount: tableInfo.rowCount || 0,
+        physicalRecordCount: tableInfo.rowCount || 0,
+        recordLength: 0, // Not available in vtable
+        bookmarkLength: 0, // Not available in vtable
+        compressionRoutine: tableInfo.compressionRoutine || "",
+        encoding: "", // Not available in vtable
+        creationTimeStamp: tableInfo.creationTimeStamp || "",
+        modifiedTimeStamp: tableInfo.modifiedTimeStamp || "",
```
scottdover: nitpick. Can we alphabetize these?

client/src/connection/itc/ItcLibraryAdapter.ts
```diff
@@ -196,6 +196,57 @@ class ItcLibraryAdapter implements LibraryAdapter {
...
+        columnCount: 0,
+        logicalRecordCount: 0,
+        physicalRecordCount: 0,
+        recordLength: 0,
+        bookmarkLength: 0,
+        compressionRoutine: "",
+        encoding: "",
+        creationTimeStamp: "",
+        modifiedTimeStamp: "",
+      };
```
scottdover: Can we store this as `const basicInfo` at the top of this function, and extend it with our data up above (i.e. `...basicInfo`). Might also be worth moving the relevant comments (i.e. `// Not available in vtable`) in line w/ the basic info

client/src/webview/TablePropertiesViewer.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (selectedContent) {
+    selectedContent.classList.add("active");
+  }
+
+  if (clickedTab) {
+    clickedTab.classList.add("active");
+  }
+}
+
+document.addEventListener("DOMContentLoaded", () => {
```
scottdover: If we add event listeners, it's preferable we have code to remove those event listeners. This may not be completely necessary since it's all isolated to a "static" tab, but if it isn't too hard a change, it might be worth doing

client/src/connection/itc/ItcLibraryAdapter.ts
```diff
@@ -11,7 +11,7 @@ import {
   TableData,
   TableRow,
 } from "../../components/LibraryNavigator/types";
-import { ColumnCollection } from "../rest/api/compute";
+import type { ColumnCollection, TableInfo } from "../rest/api/compute";
```
scottdover: Interesting. I've never seen `import type` before

kpatl1: Honestly, I have not either. I think it may have been put in there from CoPilot when I was experiencing syntax issues. I can go ahead and revert it to a normal import because it seems to be working fine now.

enzhihpp:
> Hey @enzhihpp . These changes were accidentally merged prior to testing. Can we get this tested and perhaps create additional bugs for things that are oustanding?

Yes, we will do that.

client/src/connection/itc/ItcLibraryAdapter.ts
scnwwu: This is actually suggested by me at https://github.com/sassoftware/vscode-sas-extension/pull/1631#discussion_r2354624233
`import type` is desired as ITC should NOT really import REST implementation but just referencing some types since the types are identical.
`import type` will be removed after compiling to JS thus is clear to see that there's no real dependency.

Issue #1579, opened 2025-07-30
Collapsing Comment Blocks


chunky:
### Is your feature request related to a problem? Please describe.

Using the mouseover arrow in the margin just to the left of code, I'm able to collapse procs, data step, etc.

Currently that arrow doesn't appear to multiline comments. Please can that collapsing be added for multiline comments?

### Describe the solution you'd like

Arrow to the left of the code block, that matches the one in procs/data steps.

### Describe alternatives you've considered

It's possible to do something like a null data step wrapping around the comment, then collapsing that block, but ... that feels like a random contortion that shouldn't be necessary, and makes the code itself much messier.

### SAS version

SAS 9.4 on Linux with SSH connection type

PR #1638, opened 2025-09-15
feat(LanguageServiceProvider): added code comment collapsing

kpatl1:
**Summary**
Added multiline comment folding support for SAS files. Users can now collapse/expand multiline comment blocks (/* ... */) using folding arrows in the editor margin, matching the existing folding behavior for data steps and procs.

**Testing**
- Created test file with various multiline comment scenarios (standalone comments, comments within code blocks)
- Verified folding arrows appear for multiline comments spanning multiple lines
- Confirmed single-line comments do not show folding arrows (expected behavior)
- Tested in Extension Development Host to ensure functionality works correctly
- Validated that existing code block folding (data steps, procs) remains unaffected

Fixes https://github.com/sassoftware/vscode-sas-extension/issues/1579

### DIFF
--- a/server/src/sas/SyntaxProvider.ts
+++ b/server/src/sas/SyntaxProvider.ts
@@ -437,4 +437,7 @@ export class SyntaxProvider {
     }
     return block.name;
   }
+  getMultilineComments(): Array<{ startLine: number; endLine: number }> {
+    return this._multilineComments;
+  }
 }


server/src/sas/LanguageServiceProvider.ts
```diff
@@ -210,9 +211,66 @@ export class LanguageServiceProvider {
...
     }
+
     return result;
   }
 
+  private addCommentFolding(result: FoldingRange[]) {
+    const lineCount = this.model.getLineCount();
+
+    // Method 1: Try token blocks (works for comments inside code blocks)
+    if (this.syntaxProvider.lexer.tknBlks) {
```
scnwwu: This looks like an internal implementation. It would be better for the SyntaxProvider to provide a public API rather than directly access through several levels.

server/src/sas/LanguageServiceProvider.ts
```diff
@@ -210,9 +211,66 @@ export class LanguageServiceProvider {
...
+    }
+
+    // Method 2: Scan for multiline comments using syntax tokens (for standalone comments)
+    let inBlockComment = false;
+    let commentStartLine = -1;
+
+    for (let lineNum = 0; lineNum < lineCount; lineNum++) {
+      const line = this.model.getLine(lineNum);
+
+      if (!inBlockComment && line.includes("/*")) {
```
scnwwu: Simple string match is not reliable. For example there can be a string containing `/*` like `a="a/*a"`. There're various cases. We should always rely on Lexer.

server/src/sas/SyntaxProvider.ts
```diff
@@ -423,4 +437,10 @@ export class SyntaxProvider {
     }
     return block.name;
   }
+  getTokenBlocks(): any[] {
```
scnwwu: nit: this is not needed now

server/src/sas/LanguageServiceProvider.ts
```diff
@@ -210,10 +211,22 @@ export class LanguageServiceProvider {
...
+    const commentRanges = this.syntaxProvider.getMultilineComments();
+
+    for (const range of commentRanges) {
+      result.push({
+        startLine: range.startLine,
+        endLine: range.endLine,
+        kind: "comment",
+      });
+    }
+  } // DFS
```
scnwwu: nit: the DFS comment is for the next function, it would look better to be in a separate line

scnwwu (commented):
Overall good to me. Thanks!

scnwwu (approved):

Issue #1648, opened 2025-09-28
Take Save Output action against the SASNotbook code cell, focus does not change to the specified cell with right mouse clicking


Zhirong2022:
### Describe the bug

Open VSC SAS Extension base on Main branch. New a .sasnb file, create two notebook cells. Put some code in each cell and run, focus on the first cell and move mouse to the second one, take 'Save Output' action from the ContextMenu, it will save the output of the first cell unexpectedly.

### Steps to reproduce

1.Open VSC
2.New a .sasnb file and create 2 code cells
3.Run code 
4.Click the first code cell 
5.Move mouse to the second one and open ContextMenu
6.Select 'Save Output' 
7.Select ODS HTML
8.Open the saved file

### SAS code (if applicable)

```SAS
code in cell 1:
proc print data=sashelp.cars(obs=1);
run;

code in cell 2:
title 'The SAS System';
proc sgplot data=sashelp.class; histogram age; run;
```

### Expected behavior

Focus changes to the specified code cell with right mouse click, otherwise, it will not save the specified cell contents.

### Screenshots

<img width="1568" height="598" alt="Image" src="https://github.com/user-attachments/assets/c64ead2e-ddb5-4124-9213-c3fd5ef6f072" />

### Extension version

1.17.0

scnwwu:
> Hi, [@scnwwu](https://github.com/scnwwu). I wanted to get your input on this because I have been running into blockers when trying to address this bug. Basically, when a command is registered in notebook/cell/title , VS Code does not pass any information about which cell the menu was opened on. Instead, it only passes a generic context object: {from: 'cellContainer'}. I attempted to extract cell from command arguments but I could only see VS Code pass {from: 'cellContainer'} with no cell information Therefore, I believe the bug cannot be fixed due to VS Code API limitations.
> 
> As of now, I have added a confirmation dialog that shows which cell's output will be saved and instructs users to click on the desired cell first before using "Save Output"

Hi @kpatl1, the confirmation dialog looks not easy to use. There're no cell numbers displayed on the UI. It's not easy for end user to tell the cell number when there're many cells. Users just have to get used to making sure they click on the cell first, then the confirmation dialog always adds an extra step.
I guess it's worth to try my suggestion 
https://github.com/sassoftware/vscode-sas-extension/pull/1576#discussion_r2250422688
If it works it can resolve this bug and provides a better experience.

PR #1664, opened 2025-10-14
fix(exporters): added save output button on rendered output

kpatl1:
**Summary:**
added a confirmation dialog that shows which cell's output will be saved and instructs users to click on the desired cell first before using "Save Output"

**Testing:**
Created new Sasnb file and tried to save output from two different cells. 
**TODOs:**

Fixes https://github.com/sassoftware/vscode-sas-extension/issues/1648

### DIFF
--- a/client/src/components/notebook/renderers/LogRenderer.ts
+++ b/client/src/components/notebook/renderers/LogRenderer.ts
@@ -24,7 +24,7 @@ export const activate: ActivationFunction = (context) => ({
       const toolbar = document.createElement("div");
       toolbar.style.cssText = `
         position: absolute;
-        top: -22px;
+        top: -10px;
         right: 8px;
         display: flex;
         gap: 4px;


scnwwu:
I tried it and it still covers the content.
image
I know it only appears on hover but some users may feel annoying. The code cell toolbar flows beyond the top border of the cell which will not cover any content. Are you able to make it to the similar position? (I guess a negative `top` style).

And, where is the icon from? To me, it doesn't look like the save/export or related icon we are familiar with.

kpatl1:
> I tried it and it still covers the content. image I know it only appears on hover but some users may feel annoying. The code cell toolbar flows beyond the top border of the cell which will not cover any content. Are you able to make it to the similar position? (I guess a negative `top` style).
> 
> And, where is the icon from? To me, it doesn't look like the save/export or related icon we are familiar with.

I added in a negative top, let me know what you think. I was also thinking of maybe adding it to the code cell toolbar instead what are your thoughts on that? I have also adjusted the icon to get one from the VSCode library.

kpatl1:
Gotcha, I have gone and fixed the log output negative top as well.

client/src/components/notebook/exporters/index.ts
```diff
@@ -31,86 +31,32 @@ export const exportNotebook = async (client: LanguageClient) => {
...
-
-  if (!notebook || activeCell === undefined) {
-    return;
-  }
-
-  const cell = notebook.cellAt(activeCell);
-  if (!cell) {
-    return;
-  }
+let timesOutputSaved = 0;
```
scnwwu, line 34: it looks like this one is global. So if I just save another notebook, the notebook name is different but the count simply added which may look weird.
In my opinion it's not necessary to manage a count. If a name duplicated, system will prompt the user to change a name, right?

client/src/components/notebook/renderers/HTMLRenderer.ts
```diff
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 import type { ActivationFunction } from "vscode-notebook-renderer";
 
+let outputIndex = 0;
```
scnwwu, line 5: looks like this one is global. So if I switch between multiple notebooks the output index will be wrong?

client/src/components/notebook/renderers/HTMLRenderer.ts
```diff
@@ -19,18 +21,94 @@ function replaceLast(
...
+        transition: opacity 0.1s ease;
+        background: var(--vscode-editor-background);
+        border: 1px solid var(--vscode-widget-border);
+        border-radius: 4px;
+        padding: 2px;
+        z-index: 1000;
+      `;
+
+      const saveButton = document.createElement("button");
+      saveButton.title = "Save Output";
```
scnwwu, line 55: i18n?

client/src/components/notebook/renderers/LogRenderer.ts
```diff
@@ -4,19 +4,93 @@ import type { ActivationFunction } from "vscode-notebook-renderer";
...
+export const activate: ActivationFunction = (context) => ({
   renderOutputItem(data, element) {
+    const logs: LogLine[] = data.json();
+    const currentIndex = outputIndex++;
+
+    const container = document.createElement("div");
+    container.style.position = "relative";
+
+    if (context.postMessage) {
+      const toolbar = document.createElement("div");
```
scnwwu, line 24: are you able to re-use some code with HTMLRenderer?

Issue #1578, opened 2025-07-30
There is no error validation if trying to drag and drop a file to a new position which contains a file with the same name on SAS Server


Zhirong2022:
### Describe the bug

Open VSC SAS Extension base on Main branch. Drag and drop a file to a new position which contains a file with the same name, it has no error validation.

### Steps to reproduce

1.Open VSC
2.Sign in with a valid Viya connection
3.Drag and drop a file 

### SAS code (if applicable)

```SAS

```

### Expected behavior

It has error validation if drag and drop a file to a new position which contains a file with the same name on SAS Server

### Screenshots

<img width="459" height="198" alt="Image" src="https://github.com/user-attachments/assets/d54d7dbf-4e2c-4ccf-81fa-ab9e0fe94e1a" />

### Extension version

1.15.0

PR #1665, opened 2025-10-16
fix(ContentNavigator): added name checking for duplicates on drag and drop

kpatl1:
**Summary:**

I implemented error handling for drag-and-drop file operations to the SAS Server by adding a try-catch block to RestServerAdapter.moveItem(). The Viya server's updateFileOrDirectoryOnSystem API already validates for duplicate file names by default and returns an error if a file with the same name exists at the destination. The try-catch properly catches these server errors and returns undefined, allowing the UI to display an error message to the user. This fix makes RestServerAdapter consistent with the error handling patterns in RestContentAdapter and ItcServerAdapter, and relies on server-side validation as the single source of truth rather than unreliable client-side checks.

**Testing:**
Made a new file with the same name as one within a folder on a SAS Server. Then, I dragged the root file into the folder and it prevented me from moving it. 

Fixes https://github.com/sassoftware/vscode-sas-extension/issues/1578

**TODOs:**

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -681,21 +681,6 @@ class ContentDataProvider
     return !!newUri;
   }
 
-  private async checkIfItemExistsInTarget(
-    itemName: string,
-    target: ContentItem,
-  ): Promise<boolean> {
-    try {
-      const children = await this.getChildren(target);
-      return children.some((child) => child.name === itemName);
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    } catch (error) {
-      // If we can't retrieve children, assume the item doesn't exist
-      // to avoid blocking the move operation unnecessarily
-      return false;
-    }
-  }
-
   private async handleContentItemDrop(
     target: ContentItem,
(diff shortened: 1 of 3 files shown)


scnwwu:
I think the check should NOT be done by client. Client side check is not reliable as it's async. The file system on server may changed after your getChildren call. The server should already report error for this situation. The client just needs to display the error to user. It's not limited to name collision but whatever errors failed the operation.

As #1578 mentioned, the SAS Content pane already report error.
The UI is shared:
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/components/ContentNavigator/ContentDataProvider.ts#L707

Both `RestContentAdapter` and `ItcServerAdapter` return falsy on error.
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/connection/rest/RestContentAdapter.ts#L225
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/connection/itc/ItcServerAdapter.ts#L300

So I guess `RestServerAdapter` should do similar.
https://github.com/sassoftware/vscode-sas-extension/blob/main/client/src/connection/rest/RestServerAdapter.ts#L450

PR #1703, opened 2025-11-18
chore: add vsix build script

scottdover:
**Summary:**
This copies part of the package & publish action to a new action just for building vsix files for testers


scottdover:
We _could_ tweak the `Package & Publish VSIX` script to conditionally publish, but I was a little reluctant to mess with that too much

PR #1704, opened 2025-11-19
fix:  reopens files that were opened when dragging and dropping files or folders

### LINKED ISSUE #1478: File folder will be opened as a file when drag/drop it to a new position on SAS SERVER
### Describe the bug

Open VSC SAS Extension base on Main branch. Create a 3 level file folder tree, create a file under the third level file folder. Move the third level tree node under the root node. The file folder will be opened as a file unexpectedly.

### Steps to reproduce

1.Open VSC
2.Log in with a valid Viya connection
3.Go to SAS SERVER 
4.Create a three level file folder tree (NewFolder->NewFolder1->NewFolder11)under a directory 
5.Create a new .sas file under 'NewFolder11'
6.Move 'NewFolder11' under 'NewFolder'

### SAS code (if applicable)

```SAS

```

### Expected behavior

The file folder will not be opened as a file 

### Screenshots

Image

### Client OS

_No response_

### Extension version

1.14.0

### SAS version

_No response_

### Additional context

_No response_

tarikutk:
**Summary:**

1. it opens previously opened files when dragging and dropping a file or a folder.
2. it prevents VSCode from opening a folder as a file when moving a folder  from one folder to another folder. it  instead opens the file that was opened in that folder

**Testing:**
 verified it opens previously opened files instead of opening a folder as a file.
 
 Fixes https://github.com/sassoftware/vscode-sas-extension/issues/1478

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -36,6 +36,7 @@ import { basename, join } from "path";
 import { promisify } from "util";
 
 import { profileConfig } from "../../commands/profile";
+import { SAS_FILE_SEPARATOR } from "../../connection/rest/RestServerAdapter";
 import { getResourceId } from "../../connection/rest/util";
 import { SubscriptionProvider } from "../SubscriptionProvider";
 import { ViyaProfile } from "../profile";
@@ -59,8 +60,6 @@ import {
   getFileStatement,
   isContainer as getIsContainer,
 } from "./utils";
-
-const SAS_FILE_SEPARATOR = "~fs~";
 
 class ContentDataProvider
   implements
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -60,6 +60,8 @@ import {
   isContainer as getIsContainer,
 } from "./utils";
 
+const SAS_FILE_SEPARATOR = "~fs~";
+
 class ContentDataProvider
   implements
     TreeDataProvider<ContentItem>,
@@ -684,7 +686,7 @@ class ContentDataProvider
           newUri,
         );
         if (newFileUri) {
-          commands.executeCommand("vscode.open", newFileUri);
+          await commands.executeCommand("vscode.open", newFileUri);
         }
       }
     }
@@ -733,10 +735,12 @@ class ContentDataProvider
           );
           try {
             return decodeURIComponent(uriWithoutPrefix);
-          } catch {
+          } catch (error) {
+            console.error("Failed to decode URI component:", error);
             return uriWithoutPrefix;
           }
-        } catch {
+        } catch (error) {
+          console.error("Failed to extract path from URI:", error);
           return "";
         }
       };
@@ -748,7 +752,7 @@ class ContentDataProvider
       if (
         oldBasePath &&
         closedFilePath &&
-        closedFilePath.startsWith(oldBasePath + "~fs~")
+        closedFilePath.startsWith(oldBasePath + SAS_FILE_SEPARATOR)
       ) {
         try {
           const relativePath = closedFilePath.substring(oldBasePath.length);
@@ -767,9 +771,10 @@ class ContentDataProvider
           );
 
           return Uri.parse(
-            `sasServer:/${filename}?${encodeURIComponent(newQuery)}`,
+            `${newItemUri.scheme}:/${filename}?${encodeURIComponent(newQuery)}`,
           );
-        } catch {
+        } catch (error) {
+          console.error("Failed to construct new file URI:", error);
           return null;
         }
       }
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -667,18 +667,115 @@ class ContentDataProvider
     if (!targetUri) {
       return false;
     }
-
     const closing = closeFileIfOpen(item);
-    if (!(await closing)) {
+    const closedFiles = await closing;
+    if (!closedFiles) {
       return false;
     }
 
     const newUri = await this.model.moveTo(item, targetUri);
-    if (closing !== true) {
-      commands.executeCommand("vscode.open", newUri);
+    if (Array.isArray(closedFiles) && closedFiles.length > 0) {
+      // Reopen only the files that were closed
+      for (const closedFileUri of closedFiles) {
+        // Calculate the new URI for each closed file
+        const newFileUri = this.calculateNewFileUri(
+          closedFileUri,
+          item,
+          newUri,
+        );
+        if (newFileUri) {
+          commands.executeCommand("vscode.open", newFileUri);
+        }
+      }
     }
 
     return !!newUri;
+  }
+
+  private calculateNewFileUri(
+    closedFileUri: Uri,
+    movedItem: ContentItem,
+    newItemUri: boolean | Uri,
+  ): Uri | null {
+    if (typeof newItemUri === "boolean" || !newItemUri) {
+      return null;
+    }
+
+    const isFolder = getIsContainer(movedItem);
+
+    // If the moved item is a file and matches the closed file, return the new URI
+    if (
+      !isFolder &&
+      closedFileUri.toString() === movedItem.vscUri?.toString()
+    ) {
+      return newItemUri;
+    }
+
+    // If the moved item is a folder, calculate the new path for files within it
+    if (isFolder && movedItem.vscUri) {
+      const extractPathFromUri = (uri: string): string => {
+        try {
+          const queryStart = uri.indexOf("?");
+          if (queryStart === -1) {
+            return "";
+          }
+
+          const queryString = uri.substring(queryStart + 1);
+
+          const decodedQuery = decodeURIComponent(queryString);
+          const idMatch = decodedQuery.match(/id=(.+)/);
+          if (!idMatch || !idMatch[1]) {
+            return "";
+          }
+          const uriWithoutPrefix = idMatch[1].replace(
+            /\/compute\/sessions\/[a-zA-Z0-9-]*\/files\//,
+            "",
+          );
+          try {
+            return decodeURIComponent(uriWithoutPrefix);
+          } catch {
+            return uriWithoutPrefix;
+          }
+        } catch {
+          return "";
+        }
+      };
+
+      const oldBasePath = extractPathFromUri(movedItem.vscUri.toString());
+      const closedFilePath = extractPathFromUri(closedFileUri.toString());
+
+      // Check if the closed file was inside the moved folder
+      if (
+        oldBasePath &&
+        closedFilePath &&
+        closedFilePath.startsWith(oldBasePath + "~fs~")
+      ) {
+        try {
+          const relativePath = closedFilePath.substring(oldBasePath.length);
+          const filename = relativePath.replace(/^~fs~/, "");
+          const newUriStr = newItemUri.toString();
+          // Extract and modify the query to append the filename path
+          const queryMatch = newUriStr.match(/\?(.+)$/);
+          if (!queryMatch) {
+            return null;
+          }
+
+          const decodedQuery = decodeURIComponent(queryMatch[1]);
+          const newQuery = decodedQuery.replace(
+            /(\/files\/[^&]*)/,
+            `$1~fs~${filename}`,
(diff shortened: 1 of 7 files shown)


client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -667,20 +667,117 @@ class ContentDataProvider
...
+          const queryMatch = newUriStr.match(/\?(.+)$/);
+          if (!queryMatch) {
+            return null;
+          }
+
+          const decodedQuery = decodeURIComponent(queryMatch[1]);
+          const newQuery = decodedQuery.replace(
+            /(\/files\/[^&]*)/,
+            `$1~fs~${filename}`,
+          );
```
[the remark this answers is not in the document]
tarikutk: did not work.

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -667,20 +667,117 @@ class ContentDataProvider
     if (!targetUri) {
       return false;
     }
-
     const closing = closeFileIfOpen(item);
-    if (!(await closing)) {
+    const closedFiles = await closing;
+    if (!closedFiles) {
```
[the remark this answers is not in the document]
tarikutk: closedFiles will not be false

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -667,20 +667,117 @@ class ContentDataProvider
...
 
     return !!newUri;
   }
 
+  private calculateNewFileUri(
+    closedFileUri: Uri,
+    movedItem: ContentItem,
+    newItemUri: boolean | Uri,
+  ): Uri | null {
+    if (typeof newItemUri === "boolean" || !newItemUri) {
```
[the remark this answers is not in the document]
tarikutk: no. newItemUri will be boolean or Uri. that is what this.model.moveTo returns

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -667,20 +667,117 @@ class ContentDataProvider
...
+
+      // Check if the closed file was inside the moved folder
+      if (
+        oldBasePath &&
+        closedFilePath &&
+        closedFilePath.startsWith(oldBasePath + "~fs~")
+      ) {
+        try {
+          const relativePath = closedFilePath.substring(oldBasePath.length);
+          const filename = relativePath.replace(/^~fs~/, "");
```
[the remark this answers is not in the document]
tarikutk: it did not work.

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -667,20 +669,120 @@ class ContentDataProvider
...
-    if (!(await closing)) {
+    const closedFiles = await closing;
+    if (!closedFiles) {
       return false;
     }
 
     const newUri = await this.model.moveTo(item, targetUri);
-    if (closing !== true) {
-      commands.executeCommand("vscode.open", newUri);
+    if (Array.isArray(closedFiles) && closedFiles.length > 0) {
```
scottdover: With drag and drop, at the moment, we call `moveItem` for each item dragged, and we call `closeFileIfOpen` for _one_ item. And, `newUri` _should_ be returned after the item is moved.

Is there an issue where the `newUri` is wrong? If _that's_ the case, this should be happening at the model level

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -60,6 +60,8 @@ import {
   isContainer as getIsContainer,
 } from "./utils";
 
+const SAS_FILE_SEPARATOR = "~fs~";
```
scottdover: Fwiw, this is sas server specific. Anything that is specific to connection type should happen in the adapter if possible.

tarikutk: okay. I updated that. I am importing from the adapter instead.

client/src/components/ContentNavigator/ContentDataProvider.ts
tarikutk: no. I did not see an issue there. I was just adding additional checking in case it returns empty array at some point.

scottdover: This code could be simplified to...
```
const newUri = await this.model.moveTo(item, targetUri);
if (!newUri || !Array.isArray(closedFiles)) {
  return !!newUri;
}

// Reopen only the files that were closed
for (const closedFileUri of closedFiles) {
  // Calculate the new URI for each closed file using the adapter
  const newFileUri = this.model
    .getAdapter()
    .calculateNewFileUri?.(closedFileUri, item, newUri);
  if (newFileUri) {
    await commands.executeCommand("vscode.open", newFileUri);
  }
}
```
With the above, we're only checking Array.isArray since `closeFiles` can also be a boolean. We don't need to check if there's actually anything in the array (i.e. `closedFiles.length`) as that check is unnecessary.

Additionally, since `newUri` is unchanging, I moved it outside of the for loop. 

Finally, this change did require updating the type here: https://github.com/sassoftware/vscode-sas-extension/blob/734694e505269b718aeca11ac6a7afecf8ef942c/client/src/components/ContentNavigator/ContentModel.ts#L172 ...this should return `Promise<Uri | undefined>`

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -787,10 +799,11 @@ class ContentDataProvider
           return;
         }
 
+        const fileBuffer = await promisify(readFile)(itemUri.fsPath);
         const fileCreated = await this.createFile(
           target,
           name,
-          await promisify(readFile)(itemUri.fsPath),
+          fileBuffer.buffer,
```
scottdover: I'm unsure about all the file -> file.buffer changes. If we can convert them back, and github actions don't complain, I'd probably say leave them for now. Otherwise, just let me know if they _have_ to be changed in this PR and we'll make sure we've some use cases for testers to check.

client/src/components/ContentNavigator/types.ts
```diff
@@ -94,6 +94,11 @@ export interface ContentAdapter {
     item: ContentItem,
     targetParentFolderUri: string,
   ) => Promise<Uri | undefined>;
+  calculateNewFileUri?: (
```
scottdover: Should this be optional? It _should_ if this isn't an issue that affects SAS content. If it _does_ impact sas content, I would expect to see changes in `RestContentAdapter` as well

Also, super nitpicky, but these functions for this type are all alphabetized. Will you alphabetize this new function?

client/src/connection/itc/ItcServerAdapter.ts
```diff
@@ -306,6 +306,79 @@ class ItcServerAdapter implements ContentAdapter {
     }
   }
 
+  public calculateNewFileUri(
```
scottdover, line 313: It feels like there's a lot of duplication between this and the rest server adapter implementation. From what I could tell, the three pieces that are unique for this function and the other are:
 - How directory separators are determined
 - how to extractPathFromUri
 - how to generateNewFilePath (basically, lines 352-371 here, 482-505 in the other file)

With that in mind,  I'm wondering if this method can look more like this...
```
const extractPathFromUri = (uri: string) => {
      /* process things */
      return '';
    };
    const getDirectorySeparator = (basePath: string) => {
      /*
      use getDirectorySeparator in ItcServerAdapter, or return
      the separator for rest
      */
    };
    const generateNewFilePath = (
      oldBasePath: string,
      closedFilePath: string,
      newItemUri: Uri
    ) => {
      /* Do specific calculation for ITC or REST */
    };

    return calculateNewFileUri(
      extractPathFromUri,
      getDirectorySeparator,
      generateNewFilePath,
    );
```

With the above, `extractPathFromUri`, `getDirectorySeparator`, and `generateNewFilePath` are unique for `ItcServerAdapter` _or_ `RestServerAdapter` are unique, while the code moved to a shared `calculateNewFileUri` function would contain the duplicated bits (NOTE: This is an oversimplification to some extent and there may need to be more parameters passed into the shared function)

client/src/components/ContentNavigator/ContentDataProvider.ts
tarikutk: I converted them back.

client/src/components/ContentNavigator/types.ts
tarikutk: yes. I think it should be optional. restContentAdapter is not calling it at all because closeFileIfOpen is returning true rather than array of closed files for restContentAdapter.

client/src/connection/itc/ItcServerAdapter.ts
tarikutk: I broke down the functions to utils instead of passing those functions as a prop. passing 6 parameters and passing 3 of them back to those functions was not looking good. I added them in utils of a specific adapter instead.

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -667,15 +667,25 @@ class ContentDataProvider
...
-      commands.executeCommand("vscode.open", newUri);
+    if (!newUri || !Array.isArray(closedFiles)) {
+      return !!newUri;
+    }
+    // Reopen only the files that were closed
+    for (const closedFileUri of closedFiles) {
+      // Calculate the new URI for each closed file using the adapter
+      const newFileUri = this.model
+        .getAdapter()
+        .calculateNewFileUri?.(closedFileUri, item, newUri);
```
scottdover, line 685: Calls to adapter methods should be contained within a `ContentModel` (i.e. this should be `this.model.calculateNewFileUri`). It doesn't look like we expose adapter in any other content data provider methods.

client/src/connection/itc/util.ts
```diff
@@ -18,5 +19,51 @@ export const decodeEntities = (msg: string): string => {
 export const escapePowershellString = (unescapedString: string): string =>
   unescapedString.replace(/(`|"|'|\$|\(|\)|%|{|}|\[|\])/g, "`$1");
 
+export const extractPathFromUri = (uri: string): string => {
+  try {
+    const queryStart = uri.indexOf("?");
```
scottdover, line 24: nitpick. We're doing this `indexOf` query param check here and below. Creating a small shared function to handle this functionality would prevent us from repeating ourselves.

client/src/connection/itc/util.ts
```diff
@@ -18,5 +19,51 @@ export const decodeEntities = (msg: string): string => {
...
+};
+
 export const getDirectorySeparator = (path: string): string =>
   path.lastIndexOf("/") !== -1 ? "/" : "\\";
+
+export const generateNewFilePath = (
+  oldBasePath: string,
+  closedFilePath: string,
+  newItemUri: Uri,
+  dirSeparator: string,
```
scottdover, line 42: There's no need to pass this in as a parameter. The function is available to call _just_ above this one.

client/src/connection/rest/util.ts
```diff
@@ -89,3 +89,66 @@ export const getItemContentType = (item: ContentItem): string | undefined => {
...
 export const getResourceId = (uri: Uri): string => uri.query.substring(3); // ?id=...
+
+export const extractPathFromUri = (uri: string): string => {
+  try {
+    const queryStart = uri.indexOf("?");
+    if (queryStart === -1) {
+      return "";
+    }
+
+    const queryString = uri.substring(queryStart + 1);
```
scottdover, line 100: Seeing this behavior once more...if this behavior is shared across multiple connection types, a suitable place to define a helper function may be in `client/src/connection/util`

client/src/connection/rest/util.ts
```diff
@@ -89,3 +89,66 @@ export const getItemContentType = (item: ContentItem): string | undefined => {
...
+    // Extract and modify the query to append the filename path
+    const queryMatch = newUriStr.match(/\?(.+)$/);
+    if (!queryMatch) {
+      return null;
+    }
+
+    const decodedQuery = decodeURIComponent(queryMatch[1]);
+    const newQuery = decodedQuery.replace(
+      /(\/files\/[^&]*)/,
+      `$1~fs~${filename}`,
```
scottdover, line 144: Would it be possible to use `SAS_FILE_SEPARATOR` here and above (you'll likely need to change that to `new RegExp(...)`) instead of hard coding this?

scottdover (commented):
This is looking a lot better...just a few minor things.

Issue #1216, opened 2024-09-20
Delete multiple Tables from a library in Libraries pane


SebastianReimann:
**Describe the bug**
In the libraries pane all current sas libraries including their tables are displayes. If one wants to delete more than one table from the library, one can select multiple tables by pressing STRG while clicking. All selected tables are highlighted in light blue color. On right clicking one of the selected tables the highlighting changes to a light gray and the right-clicked table get sourrounded by a blue sqare. In the context menue one can select to delete the table. Only the blue surrounded table gets deleted, not all previously selected tables as exprected.

**Steps to reproduce**

1. Copy multiple tables to work library (proc copy in=sashelp out=work; select cars class; run)
2. Go to "LIBRARIES" pane and select the work library
3. Hold STRG and left-click both just copied tables
4. Righ-click the just selected tables and select Delete from the context menue
5. Only one table is deleted instead of all selected tables

**Expected behavior**
All selected tables are deleted at the same time. Ideally there would be a confirmation question about deleting multipe tables

**Screenshots**
![image](https://github.com/user-attachments/assets/177488ed-56fd-41e1-ad23-5933c63ee697)

**Environment (please complete the following information):**
Client OS:  Windows 11
Extension version: 1.10.2
SAS version: Viya 2024.03

**Additional context**

PR #1705, opened 2025-11-20
fix: delete multiple tables from lib pane

IreneS1:
**Summary:**
Enhanced the delete functionality to support multiple table selection with a confirmation dialog that lists all selected tables. Once the user confirms, the tables are deleted. Single table deletion proceeds without a confirmation dialog.

**Testing:**
1) deleteTables - deletes multiple tables successfully
-Tests successful deletion of 2 tables
-Verifies API is called correct number of times

2) deleteTables - reports failures when some tables fail to delete
-Tests partial failure scenario (3 tables: succeed, fail, succeed)
-Verifies error message contains only failed table names
-Confirms all deletion attempts were made

Fixes: [#1216](https://github.com/sassoftware/vscode-sas-extension/issues/1216)

### DIFF
--- a/client/src/components/LibraryNavigator/const.ts
+++ b/client/src/components/LibraryNavigator/const.ts
@@ -5,7 +5,7 @@ import { l10n } from "vscode";
 export const Messages = {
   TableDeletionError: l10n.t("Unable to delete table {tableName}."),
   TablesDeletionWarning: l10n.t(
-    "Are you sure you want to delete {count} table(s): {tableNames}?",
+    "Are you sure you want to delete the selected tables?",
   ),
   ViewTableCommandTitle: l10n.t("View SAS Table"),
 };
--- a/client/src/components/LibraryNavigator/index.ts
+++ b/client/src/components/LibraryNavigator/index.ts
@@ -21,6 +21,7 @@ import DataViewer from "../../panels/DataViewer";
 import TablePropertiesViewer from "../../panels/TablePropertiesViewer";
 import { WebViewManager } from "../../panels/WebviewManager";
 import { SubscriptionProvider } from "../SubscriptionProvider";
+import { treeViewSelections } from "../utils/treeViewSelections";
 import LibraryAdapterFactory from "./LibraryAdapterFactory";
 import LibraryDataProvider from "./LibraryDataProvider";
 import LibraryModel from "./LibraryModel";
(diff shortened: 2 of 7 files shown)


client/src/components/LibraryNavigator/index.ts
```diff
@@ -128,6 +157,15 @@ class LibraryNavigator implements SubscriptionProvider {
     this.libraryDataProvider.useAdapter(this.libraryAdapterForConnectionType());
   }
 
+  private treeViewSelections(item: LibraryItem): LibraryItem[] {
+    const items =
+      this.libraryDataProvider.treeView.selection.length > 1 || !item
+        ? this.libraryDataProvider.treeView.selection
+        : [item];
+
+    return items.filter(Boolean);
```
scottdover: Is there ever a case where its would be `falsy` here? Do we _need_ this `.filter`?

client/src/components/LibraryNavigator/index.ts
```diff
@@ -128,6 +157,15 @@ class LibraryNavigator implements SubscriptionProvider {
     this.libraryDataProvider.useAdapter(this.libraryAdapterForConnectionType());
   }
 
+  private treeViewSelections(item: LibraryItem): LibraryItem[] {
+    const items =
```
scottdover: I see this was copy/pasted from `ContentNavigator`. Could we make this a shared function that takes a treeview, perhaps in `client/src/components/utils`? 

Right now, I think there's a bug with this (here and in content navigator). Putting this functionality in one place will make it easier to resolve. 

In case you're curious, or want to tackle this in your PR (definitely out of scope), here's the bug:
 - Create some deletable tables. I ended up running...
```
data work.class;
    set sashelp.class;
run;
data work.classfit;
    set sashelp.classfit;
run;
data work.air;
    set sashelp.air;
run;
data work.aarfm;
    set sashelp.aarfm;
run;
```
 - Select `class` and `classfit`
 - With those two selected, right click on `aarfm` and press delete

In this scenario, I would expect `aarfm` to be the item up for deletion. However, what you get instead is a prompt to delete class and classfit.

Like I said, this also happens with sas content and server files, so feel free to not resolve it in this PR and I can file a bug.

client/src/components/LibraryNavigator/index.ts
```diff
@@ -66,8 +68,35 @@ class LibraryNavigator implements SubscriptionProvider {
...
       commands.registerCommand("SAS.deleteTable", async (item: LibraryItem) => {
+        const selectedItems = this.treeViewSelections(item);
+
+        if (selectedItems.length === 0) {
+          return;
+        }
+
         try {
-          await this.libraryDataProvider.deleteTable(item);
+          if (selectedItems.length === 1) {
```
scottdover: Instead of having two different paths here, how about we:
 - Get rid of `LibraryDataProvider.deleteTable` (you'll likely need to make a small update to tests)
 - Always call `LibraryDataProvider.deleteTables`
 - Always show a prompt regardless if you're deleting one or more tables

client/src/components/LibraryNavigator/const.ts
```diff
@@ -4,6 +4,9 @@ import { l10n } from "vscode";
 
 export const Messages = {
   TableDeletionError: l10n.t("Unable to delete table {tableName}."),
+  TablesDeletionWarning: l10n.t(
+    "Are you sure you want to delete {count} table(s): {tableNames}?",
```
scottdover: It might be worth asking @skyeaussie what message to use here. We may not need the count or extra parens around the s. Maybe something like...
```
Are you sure you want to delete the following tables: {tableNames}?
```
It looks like studio uses the following two messages: 
```
Are you sure you want to delete the selected tables?
Are you sure you want to delete the table "{tableName}"
```

client/test/components/LibraryNavigator/LibraryDataProvider.test.ts
```diff
@@ -270,4 +270,87 @@ describe("LibraryDataProvider", async function () {
 
     deleteTableStub.restore();
   });
+
+  it("deleteTables - deletes multiple tables successfully", async () => {
```
scottdover, line 278: 🙌

client/src/components/LibraryNavigator/const.ts
IreneS1: I asked @skyeaussie about her thoughts and she liked your suggestions. Thanks!

client/src/components/LibraryNavigator/index.ts
IreneS1: You're right, removed.

client/src/components/LibraryNavigator/index.ts
IreneS1: I moved the main portion of treeViewSelections to the utils. If you don't mind, can you create a bug for that - you can assign it to me. Thanks!

scottdover: Thanks for moving that. Here's the bug - https://github.com/sassoftware/vscode-sas-extension/issues/1744

Let me know if there's anything I need to clarify with that one :)

PR #1734, opened 2025-12-18
fix: correctly render katex symbols on html export

### LINKED ISSUE #1735: KaTaX symbols do not render correctly when exporting to HTML from a SAS Notebook file
### Describe the bug

KaTaX symbols do not render correctly when exporting to HTML from a SAS Notebook file

### Steps to reproduce

1. Create a SAS notebook containing KaTaX symbols in one or more markdown cells. Observe that they render correctly when viewing the rendered markdown.
2. Click Export.
3. Select HTML. Click OK.
4. Open the generated html file and observe that KaTaX syntax is rendered as raw text.

[test-math-export-BEFORE.html](https://github.com/user-attachments/files/24240465/test-math-export-BEFORE.html)

### SAS code (if applicable)

```SAS

```

### Expected behavior

A SAS Notebook containing cells with KaTaX syntax should display the rendered result when exported to HTML the same as it does when viewing the rendered markdown cell in the notebook.

### Screenshots

_No response_

### Client OS

_No response_

### Extension version

latest

### SAS version

_No response_

### Additional context

_No response_

smorrisj:
**Summary:**
Fixes #1735 
KaTeX symbols in sas notebooks currently do not render when exported to HTML. This PR does several things:

1. Add a quick pick menu when clicking on "Export" from the notebook to present export types (html, sas). This allows for a consistent user experience across different environments that the extension runs in (Desktop / Viya Workbench / etc). Current behavior is relying on the hosts native file picker to select the correct export type, which has usability issues in hosted environments.

2. Adds marked-katex-extension to make marked aware of katex syntax. This will ensure that marked interprets the katex symbols correctly and generates the correct html for them.

3. Updates build to pull in minified katex css and fonts. This is needed so that the generated html from step 2 has the correct styling.

**Testing:**
I created a SAS notebook file that reproduces the export bugs, and then re-exported after the above changes were implemented. Tested html export of the notebook to ensure that KaTeX symbols are rendered, both with and without SAS code that produces ODS output. For internal contributors, I can provide the notebook upon request (github does not seem to allow upload of .sasnb files on PRs 😞 )


client/src/components/notebook/exporters/toHTML.ts
```diff
@@ -73,8 +87,17 @@ const exportCells = async (cells: NotebookCell[], client: LanguageClient) => {
 };
 
 const markdownToHTML = (doc: TextDocument) => {
+  let text = doc.getText();
+
+  // Normalize display math blocks: ensure $$ delimiters are on their own lines with blank lines around them
+  text = text.replace(/\$\$([\s\S]*?)\$\$/g, (_match, content) => {
```
scottdover: nitpick. It could be helpful to provide an example input/output for this one

tools/build.mjs
```diff
@@ -80,6 +80,49 @@ const copyFiles = () => {
   foldersToCopy.forEach((item) =>
     fs.cpSync(item.src, item.dest, { recursive: true }),
   );
+
+  // Process KaTeX CSS - embed fonts as base64 data URIs for self-contained HTML
+  let katexCss = fs.readFileSync(
```
scottdover, line 62: This feels a little misplaced in `copyFiles`. I suppose it is copying/processing/etc, but it almost feels like this should be in its own function

client/src/components/notebook/exporters/toHTML.ts
smorrisj: Fixed in [b8e0978](https://github.com/sassoftware/vscode-sas-extension/pull/1734/commits/b8e0978c28ae58e0e473ecb17207bae6ca36329d)

tools/build.mjs
smorrisj: Fixed in [b8e0978](https://github.com/sassoftware/vscode-sas-extension/pull/1734/commits/b8e0978c28ae58e0e473ecb17207bae6ca36329d)

Issue #1681, opened 2025-10-29
[Table Viewer] Missing numeric value not shown as decimal point - Using Viya connection


Zhirong2022:
### Describe the bug

Open VSC SAS Extension base on Main branch. Open a table which contains missing numeric values, they are not shown as a single decimal point.

### Steps to reproduce

1.Sign in with a valid Viya connection
2.Open SASHELP.GAS on libraries pane
3.Scroll to row109 or row118

### SAS code (if applicable)

```SAS

```

### Expected behavior

Missing numeric value is shown as a single decimal point with Viya connection

### Screenshots

<img width="991" height="177" alt="Image" src="https://github.com/user-attachments/assets/24923790-9e82-41c0-95ee-0638a4f5a56e" />

### Extension version

v1.17.0

PR #1755, opened 2026-01-14
show missing value as '.' instead of null

danichiavegatto:
**Summary:**
When making the request for rows, we weren't requesting missing values to be formatted, so instead they were being returned as (null). Adding in the right request parameter returns the expected missing '.' value instead.

**Testing:**
- [x] Made sure empty values displayed in the table viewer as '.' instead of empty values


danichiavegatto:
this fixes #1681

PR #1757, opened 2026-01-15
feat: add filter for data viewer

scottdover:
**Summary:**
This change adds a filter bar to the dataviewer and makes sure the data remains filtered
when switching between tabs.

**Testing:**
 - [x] Tested interacting with the filter bar using keyboard only
 - [x] Tested filtering results w/ sorted columns
 - [x] Tested Rest (viya) and ITC (IOM) connections
 - [x] Made sure filters persisted when switching between multiple data viewer tabs


smorrisj:
LGTM. Scott and I chatted online for a bit. Recommend creating a separate issue to improve display of error messages in the component level. Not an issue specific to this PR, but would make for a better user experience to show better error messages coming from left hand side components.

PR #1800, opened 2026-02-20
chore: update webviews to retain context when hidden

scottdover:
**Summary:**
This changes how `DataViewer` maintains context. In our first iteration,
`retainContextWhenHidden: true` was avoided because the
[documentation](https://code.visualstudio.com/api/extension-guides/webview)
said:

> keep in mind that this has high memory overhead and should only be used when
> other persistence techniques will not work

With that in mind, the implementation relied on re-fetching data each time a
view was displayed. As `DataViewer` grows increasingly complex, the amount of
state required to persist/fetch data will be increase. 

Thus, this PR changes things such that we retain context when switching between
tabs. The downside is that there is a higher memory footprint (it was hard to
find data on this but a stack overflow post suggested an empty web view
using `retainContextWhenHidden: true` consumes 50mb of memory).
However, there are numerous upsides:

 - This solves https://github.com/sassoftware/vscode-sas-extension/issues/1791
 - Grid state no longer has to be persisted/fetched when a tab is displayed
 - The excess xhr requests that were made to re-fetch data are no longer needed
 - From a developer perspective, adding new features to the dataviewer will
   require less time worrying about data persistence/retrieval for things that
   should be in-memory

**Testing:**
- [x] Open a table, apply filters & sorts, change scroll position, etc
- [x] Tab away and back to table, and make sure state hasn't changed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add the ability to pin columns ([#1781](https://github.com/sassoftware/vscode-sas-extension/pull/1781))
 - Allow running R code files in SAS Content ([#1845](https://github.com/sassoftware/vscode-sas-extension/pull/1845))
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


smorrisj (approved):

allenfrazelle-vsce:
Fix looks good!

PR #1820, opened 2026-03-06
chore: create table copy/paste functionality

### LINKED ISSUE #1408: Being Able to Copy Data from Data View

Sometimes I want to select data out of a table to use in a query but `Ctrl+C` doesn't work even after highlighting the value. It is frustrating to not be able to select data out of tables to pull over to code for use. A lot of times the reason for copying data out of a table is for troubleshooting purposes and is used in throw away code.

I would like to be able to copy data out of a table based on what I've selected. If I select multiple columns then have it be tab delimited and if multiple lines, just put the data on multiple lines. I think both offering `Ctrl+C` functionality as well as a context menu with "Copy" would suffice.

Writing the data I need down elsewhere so I can then copy it and use it in my code.

### Additional context

I tried searching to see if this issue was already opened, but came up empty handed.

### SAS version

All

scottdover:
**Summary:**
This introduces copy/paste for tables. A user is able to make a selection by clicking and dragging up or down, but not left to right. Additionally, there's no keyboard accessibility included as this time.

Scrolling left to right has been left out of this because of how column pinning works. If a user pins columns to the left, and drags to select, do we include those in the selection? Currently, we can't get them to scroll of page (otherwise, they aren't pinned). Additionally, if we have columns that are pinned to the right, there's not a good path forward of how those should be included in the selection. One thing worth exploring is unpinning columns _just_ as we start to drag, and re-pinning after. One thing to look out for is if this causes a layout shift.

Keyboard accessibility is in the works. You can see some code in place for it. However, I've been unable to get it working without error. It would be my suggestion to get this in, and follow-up with a separate PR for including keyboard accessibility.

**Testing:**
 - [x] Make sure dragging selection works on table data, but doesn't negatively
   impact dragging columns around
 - [x] Make sure table selection includes data underneath selection rectangle,
   along with headers for that data
 - [x] Make sure pinned columns could be selected and that the resulting csv
   data is organized in order of how it's displayed in the table


scottdover:
I've reached out to @danichiavegatto about wrapping this one up. There's some commented code around keyboard accessibility that'll need to be ironed out prior to merge, but the bulk of the functionality should be there.

noahhowell117:
Planning on looking at this after Sept. release. Will follow https://github.com/sassoftware/vscode-sas-extension/issues/1930 to experiment with new localization process for the following release.

PR #1845, opened 2026-04-02
feat: updating package.json to also support R and SQL as editorLangIds

Wizzzzzzard:
### Summary
After support was added to allow running Python code files in SAS Content (https://github.com/sassoftware/vscode-sas-extension/pull/1815) this PR enables equivalent support to allow R code files in SAS Content to also be run now that Proc R is available.

#### Automated Tests
Run the extension's unit tests using npm run test to make sure we didn't break basic startup or command execution.

#### Manual Verification
I compiled the UI and tested opening a `.r` file from `sasContent` and verified that the context values effectively enable the play button. Then ran it and verified that code was wrapped using `proc r` in the logs.


Wizzzzzzard:
Just a heads up for any testers, the R files might not run on a default Viya deployment as Proc R is disabled by default (similar to Proc Python). You can enable it on a deployment using the [docs here](https://go.documentation.sas.com/doc/en/pgmsascdc/v_073/proc/p0cj61ds1h448wn12j9oplxonhhs.htm) or can use my Viya deployment that currently has it enabled- https://d115930.contour.nasrk8s-m1.plt-asr-k8s.hpos2.rnd.sas.com

anitap-11628:
Testing updates:
  Using the Proc R-enabled feature VIYA server deployment shared by Elijah Reid, I tested the setup by creating and opening a '**.r'** file from SAS Content, and verified that the context values correctly enable the Play (Run) button. 

1. Verified during execution, that code has wrapped using proc r in the logs. 
2. Validated for SQL and Python file contexts that the Play button is visible and enabled. On clicking the Play button, execution of the respective code begins, and the logs confirm that the code is wrapped using proc sql and proc python, respectively.
3. Validated by creating and opening files with unsupported extensions such as ‘.req’ or ‘.cc’ etc. Even when the file context matches the R language, the Play button is not visible.
4. Validated in SAS notebook, all supported languages like (SAS, python, R ) are working as expected.

One observation during testing, will confirm from @Wizzzzzzard .
Confirmed the observation: the presence of the play button depends solely on the file type extension, not on the file’s actual content. If there is a mismatch between the file content and its extension, the error will be handled at runtime by the backend.

scottdover (approved):

Issue #1290, opened 2024-11-25
Multi-deletion enhancement, it does not need to ask the confirmation for files or folders one by one


Zhirong2022:
**Describe the bug**
Open VSC SAS Extension base on Main branch. Sign in with a valid Viya connection, open SAS server, multi-select files or folders and delete them. It will ask the confirmation one by one and it can be enhanced to be a single message.

**Steps to reproduce**
1. Sign in with a valid Viya connection
2. Open SAS Server
3. Multi-select files or folders
4. Perform delete through ContextMenu

**Expected behavior**
It can be enhanced to be a single confirmation towards the multi-deletion

**Environment (please complete the following information):**
Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.12.0]

mnca16:
Hey @scnwwu  and @Zhirong2022 - I’d like to work on this ticket, but it looks like I’m unable to assign it to myself. Would you mind assigning it to me? Thank you! :)

scottdover:
Assigning this one to @Wizzzzzzard as there has already been some progress made :)

PR #1846, opened 2026-04-02
feat: improve multiple deletion process

Wizzzzzzard:
**Summary:**
Implements a single confirmation dialog for multi-file deletion instead of prompting for each file individually. When users select multiple files or folders and choose to delete them via the context menu, they now see one confirmation dialog showing the total count of items to be deleted, rather than being prompted separately for each item.

- Added new localized messages (`DeleteMultipleWarningMessage`, `RecycleMultipleWarningMessage`, `RecycleMultipleDirtyWarning`) in `const.ts` to support batch deletion confirmations
- Refactored the `deleteResource` command in `index.ts` to collect all selected items and show a single confirmation based on the total count

Addresses https://github.com/sassoftware/vscode-sas-extension/issues/1290

**Testing:**
How did you test this change?

1. Sign in with a valid Viya connection
2. Open SAS Server or SAS Content in the explorer
3. Select multiple files and/or folders (use Cmd/Ctrl+Click or Shift+Click)
4. Right-click and select "Delete" from the context menu
5. Verify that only ONE confirmation dialog appears showing the count of items (e.g., "Are you sure you want to permanently delete 5 items?" or "Are you sure you want to move 3 items to the Recycle Bin?")
6. Test with a single file selection to ensure the original behavior is preserved (shows file name in confirmation)
7. Test with folders containing unsaved files to verify the appropriate warning message appears

**TODOs:**

- [x] Add any supporting documentation and (optionally) update [CHANGELOG.md](CHANGELOG.md)


allenfrazelle-vsce:
Looks good, thanks Elijah!

client/src/components/ContentNavigator/ContentDataProvider.ts
```diff
@@ -847,23 +851,32 @@ class ContentDataProvider
...
-                  tab.input.uri,
-              )
-              .filter((exists) => exists),
-          ),
-        )
+        .then((results) => {
+          // Check if all tabs were successfully closed
+          const allClosed = results.every((result) => result === true);
+          if (!allClosed) {
+            // User canceled the save dialog
```
scottdover: This comment is slightly confusing

scottdover (commented):
Hey @Wizzzzzzard. I haven't looked through all of this yet, but here are my general thoughts:
 - It was decided a while ago that we don't need a confirmation when moving files to a recycle bin. My assumption is that that still holds. With that in mind, I think we can remove the recycling bin notifications
 - For deletion, since the number of deletions is somewhat unbounded, I'd suggest keeping it simple and having two messages: One for single deletion, and one for multiple deletion. For multiple deletion, I'd stick to a message like "Are you sure you want to permanently delete {count} items?" without relying on spelling out every file. That'll hopefully cut down on some of the logic here
 - Last, for the dirty files issue, I'm wondering if a single message like this works for all cases: "The selected folders contain unsaved files, are you sure you want to delete?"

Wizzzzzzard:
@scottdover I've actioned your feedback by removing the recycling notifications and simplifying the delete logic down, if you would be ok to re-review when you get the chance? :)

client/src/components/ContentNavigator/index.ts
```diff
@@ -102,54 +102,114 @@ class ContentNavigator implements SubscriptionProvider {
...
+
+          let confirmed = false;
+          if (deletableItems.length === 1) {
+            const resource = deletableItems[0];
+            const isContainer = getIsContainer(resource);
+            const canRecycle =
+              this.contentDataProvider.canRecycleResource(resource);
+            const itemHasUnsavedFiles = isContainer
+              ? await this.contentDataProvider.checkFolderDirty(resource)
+              : false;
```
scottdover, line 150: How is this different than `hasUnsavedFiles`? Presumably, if we have a single deletable item, then it should've have already triggered this same code up on line 133-137. I think we can probably clean this up.

client/src/components/ContentNavigator/index.ts
```diff
@@ -102,54 +102,114 @@ class ContentNavigator implements SubscriptionProvider {
...
+            } else if (itemHasUnsavedFiles) {
+              confirmed = !!(await window.showWarningMessage(
+                l10n.t(Messages.DirtyFolderWarning),
+                { modal: true },
+                canRecycle
+                  ? Messages.MoveToRecycleBinLabel
+                  : Messages.DeleteButtonLabel,
+              ));
+            } else {
+              confirmed = true;
```
scottdover, line 169: Since we default to `confirmed = true` here and down on line 189, I'm wondering if it makes more sense to get rid of this else case and the one below and instead start w/ `let confirmed = true;`

client/src/components/ContentNavigator/index.ts
```diff
@@ -102,54 +102,114 @@ class ContentNavigator implements SubscriptionProvider {
...
+            !(await this.contentDataProvider.closeResourceFiles(deletableItems))
+          ) {
+            return;
+          }
+
+          const recyclableItems = deletableItems.filter(
+            (resource: ContentItem) =>
+              this.contentDataProvider.canRecycleResource(resource),
           );
+          const permanentDelete =
```
scottdover, line 127: I was trying to understand this a bit more. From my general understanding, I would suspect `permanentDelete` means we have _something_ that cannot be recycled. Could this be simplified to something like...
```
const permanentDelete = deletableItems.some(
            (resource: ContentItem) =>
              !this.contentDataProvider.canRecycleResource(resource),
);
```

scottdover:
Hey @Wizzzzzzard . Our `CHANGELOG` process has changed since this PR was introduced. Do you mind rebasing and removing _any_ changes you've made in `CHANGELOG.md` (we've an automated process to handle this moving forward)?

PR #1910, opened 2026-06-16
fix: support library downloads with virtual fs uris

smorrisj:
**Summary:**
Library downloads failed in virtual file systems (e.g., vscode-local:// scheme from code-server's "Show Local" folder) fail because the code only supported local file system paths. Additionally, large table downloads could cause memory exhaustion due to missing backpressure handling in the streaming logic.

Add browser download support for non-file schemes, and fix the underlying stream backpressure issue to prevent memory problems on large libraries.

[index.ts]

- Added [streamTableToBrowserDownload()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) method with single-use token authentication, HTTP server, and browser download pipeline for virtual file systems
- Refactor download command to detect file scheme and route to appropriate handler

[LibraryModel.ts]

- Replace `forEach` with for...of loop to enable async/await backpressure handling
- Add drain event waiting when `write` returns false to prevent memory exhaustion on large tables
- Remove unused TableRow import

**Testing:**
See added unit tests. Additionally built a vsce package and tested in analytics pro and workbench.

**TODOS**
- [ ] Make the streaming timeout configurable.


scottdover (approved):
I don't understand all of browserDownload, but it looks okay to me. And it looks like we either use browser download or what we had before, so I'm good w/ it :)

client/src/components/LibraryNavigator/browserDownload.ts
```diff
@@ -0,0 +1,225 @@
...
+import { LibraryItem } from "./types";
+
+const DOWNLOAD_TOKEN_BYTES = 24; // 192-bit token
+const MAX_DOWNLOAD_FILENAME_LENGTH = 250;
+const BROWSER_CONNECTION_TIMEOUT_MS = 60_000;
+const DOWNLOAD_ENDPOINT_PATH = "/sas-library-download";
+const DOWNLOAD_TOKEN_PARAM = "token";
+const DEFAULT_DOWNLOAD_FILENAME = "table.csv";
+
+export function isValidDownloadToken(
```
scottdover, line 18: I don't love that we're only exposing these functions so that we can test them. I'd prefer just running all tests through what's available in the public interface, but I'm guessing it's likely easier this way and requires less setup

PR #1934, opened 2026-07-09
fix: syntax errors causes no libraries to display

### LINKED ISSUE #1931: Simple bugs (e.g. Variable a has been defined as both character and numeric.) make libraries disappear
### Describe the bug

Some rather simple bugs temporarily crash the library pane, for example the code below expectedly return the error message

> ERROR: Variable a has been defined as both character and numeric.

However, in addition to the error message, the library pane becomes empty and will only reappear after executing code that is working. This also appears with other bugs.

### Steps to reproduce

1. Launch VS Code with the SAS extension.
2. Sign In to local SAS session.
3. Execute the code example and see the libraries disappear

### SAS code (if applicable)

```SAS
data test;
  a = 3;
run;
data test2;
  a = "a";
run;
data combined;
  set test test2;
run;
```

### Expected behavior

Library pane should remain despite error message.

### Screenshots

_No response_

### Client OS

Windows 11

### Extension version

v1.20.0

### SAS version

9.4 TS Level 1M6, X64_10PRO platform (local instance)

### Additional context

_No response_

### Sensitive information

danichiavegatto:
**Summary:**
When SAS code is submitted with syntax issues, the LanguageService throws errors (expected) but this also causes the request to getLibraries to return no values (which is not valid). So, when fetching libraries, clear the cache of errors and query the libraries table properly

**Testing:**
Verified on a local install of the extension

**TODOs:**
none

### DIFF
--- a/client/src/connection/itc/script/itc.ps1
+++ b/client/src/connection/itc/script/itc.ps1
@@ -584,6 +584,12 @@ class SASRunner {
   }
 
   [void]GetLibraries() {
+    # Reset LanguageService to clear any error state from previous syntax errors
+    $this.objSAS.LanguageService.Reset()
+    
+    # Flush any pending output/errors
+    $this.objSAS.LanguageService.FlushLog(32768) | Out-Null
+    
     $objRecordSet = New-Object -comobject ADODB.Recordset
     $objRecordSet.ActiveConnection = $this.dataConnection
     $query = @"


mikewporter (approved):

allenfrazelle-vsce:
Everything looks good.  Libraries no longer disappear with bad SAS code.

danichiavegatto:
@allenfrazelle-vsce code merged in. ready for final verification 😊

allenfrazelle-vsce:
Verified at latest release.

PR #1937, opened 2026-07-13
fix: catch invalid query filters and escape single quotes

### LINKED ISSUE #1911: Filter on a character column in dataset viewer causes crash
### Describe the bug

Image

Image

It only happens when you filter a character column. Filtering on the numeric ones won't have issues.
After it crashes the webview (or maybe the extension host?), you can only do a Reload Window or restart VS Code to make it back to work.

### Steps to reproduce

Image

1.Open a SAS dataset like SASHELP.CARS
2.Enter a filter like Make='Acura'. Seems like only single quote (') would trigger this error. If you do a double quote("), it would be fine. 
3.You will get an error at the bottom-right. If you check the log, it would have a few errors but I did not find anything useful

### SAS code (if applicable)

```SAS

```

### Expected behavior

_No response_

### Screenshots

_No response_

### Client OS

Windows 11

### Extension version

 1.19.1

### SAS version

9.4

### Additional context

_No response_

### Sensitive information

danichiavegatto:
Signed-off-by: Dani Chiavegatto <danichiavegatto@gmail.com>

**Summary:**
If a filter contains single quotes, we need to escape them for SAS 9.4 connections as it was conflicting with the powershell query being run. 

In addition, a malformed query was creating a malformed work library that was corrupt, meaning you couldn't clear it out. 

**Testing:**
added validation tests to ItcLibraryAdapter.test.ts

**TODOs:**
NONE

### DIFF
--- a/client/src/components/LibraryNavigator/LibraryModel.ts
+++ b/client/src/components/LibraryNavigator/LibraryModel.ts
@@ -48,7 +48,11 @@ class LibraryModel {
             ),
           };
         } catch (e) {
-          return { error: e, data: { rows: [], count: 0 } };
+          const data: { data: TableData; error: Error } = {
+            error: e instanceof Error ? e : new Error(String(e)),
+            data: { rows: [], count: 0 },
+          };
+          return data;
         }
       },
     );
--- a/client/src/connection/itc/ItcLibraryAdapter.ts
+++ b/client/src/connection/itc/ItcLibraryAdapter.ts
@@ -13,19 +13,20 @@ import {
   TableQuery,
   TableRow,
 } from "../../components/LibraryNavigator/types";
-import { ColumnCollection, TableInfo } from "../rest/api/compute";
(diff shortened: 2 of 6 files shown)


mikewporter (approved):

anitap-11628:
Testing Updates:
-	Validated the filter query with single quote or double quote,  its working fine in SAS 9.4 connection.
-	There are no errors or corrupt observed in the VS code
-	VS Code works and allows users to clear the filter, without need of reload window or restart VS Code.
-	No error observed at the bottom-right or console log.
-	Tested with different combinations, the filter query is working fine:
Use case:
o	1. Make="Acura"
o	2. Make='Acura'
o	3. Make in ('Acura', 'BMW')
o	4. Make='Acura' AND Type='SUV'
o	5. Make="Acura" AND Type='SUV' AND Origin="Asia"
o	6. Make in ("BMW", "Acura") AND Type='SUV'
o	7. Make in ('BMW', 'Acura') AND Type='SUV'
o	8. Model contains ('RSX')

PR #1951, opened 2026-07-31
fix: don't auto-indent run; on same line as proc statement

### LINKED ISSUE #1200: Unexpected auto-indent at run;
**Describe the bug**
Open VSC SAS Extension base on Main branch. Type some code, indentation does not work properly in a specific scenario.  

**Steps to reproduce**
1.Type code below in the editor
```
proc print data=test;
    title "title";
    format pf70 best8.;
```
2.Type Enter to line 4
3.Type Backspace to the start point of line 4
4.Type code below
```
proc print data=sashelp.cars; run;
```

**Expected behavior**
No auto-indent of the second code block

Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.10.2]

JacinthaDev:
Fixes #1200

**Summary:**
Fixed an issue where the format-on-type provider incorrectly auto-indented a line when typing `run;` on the same line as a proc statement ex: `proc print data=sashelp.cars; run;`. 

Before this change, the extension could incorrectly apply an indentation edit even though the procblock starts and ends on the same line. To fix it I:

- Added logic to skip indentation handling for single-line proc blocks.
- Updated the multiple run; handling logic to determine whether run; is the first token on its line, which more accurately identifies when indentation behavior should be applied.

**Testing:**
Added two unit tests in `FormatOnTypeProvider.test.ts`:

1. Verifies that run; on the same line as a `proc `statement does not trigger an indentation.
2. Is a regression test to verify that `run;` on its own line within a multi-line `proc `block does trigger the expected indentation

`npm run test-server -- ./server/test/sas/FormatOnTypeProvider.test.ts` to verify all tests pass


server/src/sas/FormatOnTypeProvider.ts
```diff
@@ -109,6 +109,10 @@ export class FormatOnTypeProvider {
     if (!foldingBlock) {
       referLine = lastNotEmptyLine;
     } else if (foldingBlock?.startLine === line) {
+      if (foldingBlock?.endLine === line) {
+        // single-line block w/ run statement — don't adjust current line's indentation
+        return [];
+      }
```
michael-yilma-personal, line 118: nit (optional):
const isSingleLineBlock =
  foldingBlock?.startLine === line && foldingBlock?.endLine === line;
if (isSingleLineBlock) {
  // e.g. `proc print data=cars; run;` — leave the line's indentation alone
  return [];
}

Nurquhart (approved):

ksurajofficial (approved):
Looks good to me.

alishakale:
I'm still seeing the auto-indent behavior on this branch. It seems the bug only reproduces when there's an unclosed proc block above the `proc print data=sashelp.cars; run;` line as seen in the description for the bug.

alishakale:
Seems to be working as expected today.

PR #1967, opened 2026-08-06
fix/no-syntax-help-for-the-available-sub-option

### LINKED ISSUE #1591: [syntax] Sub-options are shown unexpectedly and there is no syntax help for the available sub-option in some Procedure
### Describe the bug

Open VSC SAS Extension base on Main branch. Type code, it has sub-options for some statement option unexpectedly. There is no syntax for the available sub-option.

### Steps to reproduce

1.Open VSC
2.Create a new .sas file and type some code
proc deepprice;
    dnn nodes=();
    dnn train=(optimize=)
3.Type Ctrl+Space in the parentheses of nodes=() to show sub-options
4.Move mouse on optimize to show Syntax help 

### SAS code (if applicable)

```SAS
proc deepprice;
    dnn nodes=();
    dnn train=(optimize=)
```

### Expected behavior

1.There is no sub-option if trying to type Ctrl+Space in the parentheses of nodes=()
2.Show syntax help for 'optimize=’

### Screenshots

Image

### Client OS

_No response_

### Extension version

1.16.0

### SAS version

_No response_

### Additional context

_No response_

ofeterovskaya:
**Summary:**
Resolves syntax help inconsistencies in PROC DEEPPRICE by removing invalid sub-option suggestions for nodes=() and providing syntax help for the optimize= sub-option.

**Changes**
Updated completion and hover context handling for parenthesized procedure statement options in CompletionProvider.ts 
Added parenthesized option context normalization so nested sub-option scenarios are resolved in the correct zone, including sub-option value context.
Added missing completion path for sub-option value context to avoid incorrect empty suggestions in nested option flows.
Updated sub-option metadata key normalization and help/syntax mapping in SyntaxDataProvider.ts , so hover lookup for sub-options like optimize works with consistent keys.
Kept expected behavior for nodes: no sub-options should be shown inside nodes parentheses.

Testing part
Added new server test coverage for this bug in deepprice_suboptions.test.ts 
Added fixture used by the test in deepprice_suboptions.sas
New test cases validate:
No sub-option completion inside nodes parentheses.
Hover help is returned for optimize in train parenthesized context.

**Testing:**
Manual:
Open a .sas file and use:
```
proc deepprice;
  dnn nodes=();
  dnn train=(optimize=)
```
Place cursor inside nodes=() and press Ctrl+Space:
Expected: no sub-options shown.
Hover on optimize in train=(optimize=):
Expected: syntax/help tooltip is shown for optimize=.

Regression checks:
Hover on train still shows its own help as before.
No unexpected completion popups outside parenthesized sub-option contexts.

Before: 
Hover mouse over optimize- no syntax help
Type Ctrl+Space in the parentheses of nodes=() -- there are shown sub-options

https://github.com/user-attachments/assets/ee0bfc61-aabd-4273-bee3-9266dd272a78

After fix: working as expected 
Expected behavior
1.There is no sub-option if trying to type Ctrl+Space in the parentheses of nodes=()
2.Show syntax help for 'optimize=’ when mouse hover

https://github.com/user-attachments/assets/d2b100e2-176b-4dae-8503-f5bea4ebc5a9

**TODOs:**


Nurquhart (approved):

server/src/sas/CompletionProvider.ts
```diff
@@ -1766,10 +1781,63 @@ export class CompletionProvider {
...
+    if (
+      ![ZONE_TYPE.PROC_SUB_OPT_NAME, ZONE_TYPE.PROC_STMT_SUB_OPT].includes(zone)
+    ) {
+      return optName;
+    }
+
+    const lineText = this.model
+      .getLine(position.line)
+      .substring(0, position.character);
+    const pattern = /([A-Za-z_][\w.]*)\s*=\s*\(/g;
```
michael-yilma-personal: nit: this looks like it only reads the current line (getLine(position.line)) when figuring out which option's parentheses the cursor is inside. I'm not sure how the completion is invoked in practice, so this may well be fine — but could you confirm it still works when the option and its sub-options span multiple lines? For example:
dnn train=(
  optimize=
);

where the cursor is on the optimize= line but the train=( is on the line above. If that case already works (or isn't a real scenario here), feel free to disregard.

server/src/sas/CompletionProvider.ts
```diff
@@ -1766,10 +1781,63 @@ export class CompletionProvider {
...
+    while ((match = pattern.exec(lineText)) !== null) {
+      parentOpt = match[1];
+      parentOptStart = match.index;
+    }
+
+    if (!parentOpt || parentOptStart < 0) {
+      return optName;
+    }
+    const tail = lineText.substring(parentOptStart);
+    const openCount = (tail.match(/\(/g) || []).length;
```
michael-yilma-personal: nit: the logic here looks like it decides whether the cursor is inside an option's parentheses by counting ( vs ) characters on the line. I'm not fully sure what values show up in practice, so this may be a non-issue — but could you check whether a literal parenthesis inside a quoted value throws the count off? For example something like:
dnn train=(where=")(")

server/src/sas/SyntaxDataProvider.ts
```diff
@@ -2103,9 +2116,11 @@ export class SyntaxDataProvider {
     stmtName = stmtName.toUpperCase();
     optName = optName.toUpperCase();
     return _tryToLoadProcedure(procName, cb, () => {
-      let data = _procStmtObj(procName, stmtName, optName, ID_SUB_OPTS);
-      if (data) {
-        data = data[ID_KEYWORDS];
+      const stmtOpt = _procStmtObj(procName, stmtName, optName);
```
michael-yilma-personal, line 2119: nit: this looks like a slight behavior change from before — previously, when a statement option had no sub-options defined, it fell back to getStatementSubOptions("global", …), whereas now if the option object exists but just has no ID_SUB_OPTS, it returns [] and skips that fallback. I'm not sure how often that path is hit in practice, so this may be intended (or never come up) — but could you confirm that options which used to pick up global sub-option completions still do? If that fallback isn't relevant here, feel free to disregard.

server/test/deepprice/deepprice_suboptions.test.ts
```diff
@@ -0,0 +1,67 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import { TextDocument } from "vscode-languageserver-textdocument";
+
+import { assert } from "chai";
+import fs from "fs";
+import { describe, it } from "mocha";
+
+import { LanguageServiceProvider } from "../../src/sas/LanguageServiceProvider";
+import { HelpData } from "../../src/sas/SyntaxDataProvider";
+
```
michael-yilma-personal, line 11: Can you add test for the multi-line parenthesized case?

michael-yilma-personal (commented):
See comments

server/src/sas/CompletionProvider.ts
ofeterovskaya: Thanks for pointing that out. I checked the multiline case  and the previous logic inspected the current line, so it could fail when the option and its sub options span multiple lines. I've updated the resolver to inspect the text up to the cursor across lines and I added a regression test to cover that scenario.

server/src/sas/CompletionProvider.ts
ofeterovskaya: Thanks for calling that out. I checked that scenario as well and the concern was valid. The previous counting logic could be thrown off by parentheses inside quoted literals. I've replaced it with a quoteaware scan that ignores parentheses inside strings and I added regression tests to cover that case.

server/src/sas/SyntaxDataProvider.ts
ofeterovskaya: I checked that path and you were right about the behavior change. I restored the previous behavior, so options that used to inherit global sub-option completions still do.

Thank you for your review

michael-yilma-personal (approved):

noahhowell117 (approved):

PR #1994, opened 2026-08-20
fix: block invalid characters in filename validation

### LINKED ISSUE #1531: File name error validation against SAS Content and File Sever does not work separately
### Describe the bug

Open VSC SAS Extension base on Main branch. Try to create a file on Content Sever or Filer Server,  the file name error validation does not work separately.

### Steps to reproduce

1.Open VSC
2.Sign in with a valid Viya connection
3.Create a new .sas file 
4.Type file name having some special character which cause error validation

### SAS code (if applicable)

```SAS

```

### Expected behavior

The file name error validation works separately 

### Screenshots

Image

### Client OS

_No response_

### Extension version

1.15.0

### SAS version

_No response_

### Additional context

_No response_

JacinthaDev:
Fixes #1531 

### Summary

Fix filename validation for Content Navigator and move validation logic into a shared validation file.

### Changes
- Refactored validation logic into `validation.ts`.
- Applied shared validation to create and rename flows.
- Updated filename validation to reject the following characters: `?`, `:`, `;`, `{`, `}`, `/`, `\`, `<`, `>`, `*`, `"`, and `|`.
- Added unit tests covering invalid filename scenarios.
- Preserved the existing requirement that filenames must include an extension (for example, `.sas`).

### Testing
- `npm run compile`
- Manually verified that filenames containing prohibited characters are rejected.

### Reviewer Note
- This change only affects client-side validation.
- Filename validation currently uses the same character rules across source types.
- **I expanded the blocked character list based on common filename restrictions. Please confirm that these are the intended validation rules for both Content Server and Sas Server**


Nurquhart:
Hi @JacinthaDev - The code looks good to me, but the list of invalid characters doesn't seem to match up with what Studio has (i.e. I was able to create files in Studio with characters that this PR prohibits).
I found these in the studio code:
Screenshot 2026-09-02 at 8 40 54 AM

I wonder if for validateFileName, we actually do want to use the sourceType and have two different Regex, one for compute and one for content

ksurajofficial:
I agree with @Nurquhart. To match with what studio does, the sourceType should be used to differentiate the regex.

JacinthaDev:
@Nurquhart Do you mind sharing the code snippet you included? and do you mean that characters invalid for compute should not be the same characters that are invalid for content. Also I'm not exactly sure what is meant by compute

JacinthaDev:
@Nurquhart @ksurajofficial Changes have been made.

ksurajofficial:
@JacinthaDev the build has failed, probably during the test phase. Could you please check?

JacinthaDev:
The latest PR fixes a few test failures and makes the client tests more reliable.

For the notebook export tests, I updated the test setup to make sure the SAS extension is fully activated before opening notebook files. I also fixed some test cleanup issues so provider registrations don't carry over into other tests. I updated the export test as well to handle Windows and Unix line ending differences.

For the profile validation test, I updated the client ID/secret test to use the expected test profile directly instead of depending on workspace settings that could change during the test run.

With these changes, the client test suite is running consistently again and the build errors are resolved.

client/src/components/ContentNavigator/index.ts
```diff
@@ -84,19 +93,16 @@ class ContentNavigator implements SubscriptionProvider {
...
-    );
-    workspace.registerTextDocumentContentProvider(
-      `${config.sourceType}ReadOnly`,
-      this.contentDataProvider,
-    );
-
-    // Remove any image preview cache files left behind by a previous session
-    // (e.g. VS Code was force-closed or the extension host crashed/timed-out
-    // before the tab-close/dispose cleanup could run).
-    void this.purgeImagePreviewCache();
```
Nurquhart, line 99: did you mean to remove this? Or was it a merge issue?

PR #1998, opened 2026-08-24
feat: restrict downloading tables

### LINKED ISSUE #1818: Provide the ability to restrict the ability to download/export data from the libraries view using Viya connection

The extension allows users to easily and quickly download data by right-clicking a table within the libraries view.  However for some this is seen as a security concern, since the extension is open source and so there is often little way of preventing users from installing and using it.

A configurable option that will disable the download option

I tested manually removing the appropriate snippet in package.json and then reloading the extension.  This worked and removed the download option entirely, but is not really a feasible solution for many.

### Additional context

_No response_

### SAS version

Viya 4

### Sensitive information

- [x] I have confirmed that this feature request contains no sensitive information

noahhowell117:
**Summary:**
When a VIYA profile's compute context has defined an attribute called allowDownload, we'll use that to determine whether the user can download tables in the Libraries view. Otherwise, this should still function the same as default where tables can be downloaded for any profile that does not have it defined.

I've also updated the viya.md for the website to explain this and provide an example config.

**Testing:**
On a Viya profile you can follow these steps:

1. Open SAS Environment Manager via the endpoint used for your active profile.
2. Select the "Contexts" option from the left-hand sidebar.
3. From the "View:" dropdown, select "Compute contexts" and then the context used for your profile.
4. With your context selected, click the edit icon in the top right of the Basic Properties view.
5. Click the plus (+) icon in the attribute section to add a new attribute.
6. In this "New Attribute" dialog, set "Attribute:" to "allowDownload" and "Value" to "false".
7. Click OK and then Save.
8. You'll likely need to refresh your connection to pick up the new setting in the profile.

You should then see that the download option has been removed for tables in the Libraries view. Switching to a profile that does not have that setting, you should then see the download option again. You can also edit the attribute to true and refresh the connection to see the download option once again on the edited Viya profile.

My testing included:
- [x] making sure the setting is honored when defined on the compute context's attributes.
- [x] making sure a VIYA profile where the attribute is not defined allows downloading tables as well as other profile connection types (Local install, IOM)
- [x] switching back and forth between a profile that has it set to false and to one that does not have it defined or support it, making sure the option appears when expected based on the active profile.
- [x] ran the website locally to make sure the additions in viya.md look as expected.

**TODOs:**

- [ ] update [CHANGELOG.md](CHANGELOG.md)
- [ ] once automation tests are in place, this would be a good candidate to cover there


Nurquhart (approved):
Code looks good to me

Tested to make sure download works normally, then added the allowDownload attribute, verified the 'download' option disappeared. Then switched to a profile that did not use that compute context, verified the 'download' option appeared and worked. Then removed the allowDownload attribute from the original context, switched back to the first profile, and verified download was back.

Only verified on Viya 4 since my understanding is that is the only environment this will be supported

grantfountain-projs (approved):

PR #2006, opened 2026-08-26
Fix: encoding corruption

### LINKED ISSUE #1965: Encoding switch corrupts non-UTF-8 source files (WLATIN1/Windows-1252) for the SAS extension to Visual Studio Code
### Describe the bug

Reported by: Paul Van Mol

More information from the user below:
- I'm working on a customer supporting clinical trials that were stored in WLATIN1 encoding.
- The .sas programs for legacy clinical studies stored on the SAS Viya server are not in UTF-8 encoding, but in WLATIN1 - Windows 1252.
- We are coaching the customer in using SAS Viya for developing clinical studies with SAS Studio.
- However, one of the participants in the webinar (Viktor Barteau) is also using SAS Extension for VS Code to connect to SAS Viya and mentioned that there is a bug in the SAS Extension for VS Code. 
- For local files in VS Code (so not accessed via SAS Extension) I do get the expected results.

### Steps to reproduce

**More information on the unexpected behaviour when using the SAS Extension to VS Code** 
- Initially the test_ANSI.sas is opened with: 

Image

- When i click on the UTF-8 encoding, I get a dialog.

Image

- I can select the correct encoding:
 
Image

- (screenshot below of the BUG)

Image

### SAS code (if applicable)

```SAS
M�ller
�123
Fran�ois
d�j� vu
```

### Expected behavior

**More information below on the expected behaviour** 
- In SAS Studio or with local files in VS Code, the Open with Encoding works.
- When I open the file "test_ansi.sas". (Note that this file contains the following added into the "SAS code section" of the ticket).

Image

- Change encoding to Western Windows (wlatin1) - Windows 1252.

Image

- For local files in VSCode (so not accessed via SAS Extension) i do get the expected results: 

Image

### Screenshots

_No response_

### Client OS

_No response_

### Extension version

1.20.0

### SAS version

_No response_

### Additional context

_No response_

### Sensitive information

EvgeniiRychkov:
**Summary**
Fixes an issue where reading or writing files with non-UTF-8 encodings (e.g. WLatin1) causes content corruption. The file data was being converted to/from string mid-pipeline, destroying the original byte representation.

**Problem**
When a user opens or saves a file encoded in a non-UTF-8 encoding (e.g. WLatin1), the content becomes corrupted, special characters are mangled or lost entirely.

**Root Cause**
File content was passed as string through the entire read/write pipeline, causing binary data to be silently corrupted for any non-UTF-8 encoding.

**Solution**
Replaced string with Uint8Array throughout the read/write pipeline, from the ContentAdapter interface down through all three adapter implementations, ContentModel, and ContentDataProvider. Places that need a string call .toString() at the boundary. For the ITC adapter, file content is now sent as base64 to the PowerShell script, which decodes it as raw bytes instead of assuming a charset. Updated the test mock accordingly.

**Testing**
npm run lint 
npm run compile 
npm test

To reproduce the bug:
1. Create a .sas file.
2. Change the encoding from UTF-8 to another encoding. I used Windows-1252.
3. Insert a non-ASCII character. I used Ü.
4. Save the file.
5. Reopen the file and change the encoding from UTF-8 back to the previously selected encoding.
6. The bug is that Ü should not appear.

It is necessary to verify that the bug cannot be reproduced in SAS Content, SAS Server, or when using an IOM connection (instructions for configuring it are available here https://rndconfluence.sas.com/spaces/COREUI/pages/967742347/Contributor+Guide+FAQ#ContributorGuide&FAQ-RemoteIOMconnectivity).

The changes also affected file upload and download functionality.

### DIFF
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -362,7 +362,7 @@ class ContentDataProvider
   }
 
   public writeFile(uri: Uri, content: Uint8Array): void | Promise<void> {
-    return this.model.saveContentToUri(uri, new TextDecoder().decode(content));
+    return this.model.saveContentToUri(uri, content);
   }
 
   public async deleteResource(item: ContentItem): Promise<boolean> {
--- a/client/src/components/ContentNavigator/ContentDataProvider.ts
+++ b/client/src/components/ContentNavigator/ContentDataProvider.ts
@@ -250,7 +250,7 @@ class ContentDataProvider
 
   public async provideTextDocumentContent(uri: Uri): Promise<string> {
     // use text document content provider to display the readonly editor for the files in the recycle bin
-    return await this.model.getContentByUri(uri);
+    return (await this.model.getContentByUri(uri)).toString();
   }
 
   public getChildren(item?: ContentItem): ProviderResult<ContentItem[]> {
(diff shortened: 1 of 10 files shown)


Nurquhart (approved):
Tested this to verify that opening images is still working, and everything looks good to me. Thanks Evgenii!

grantfountain-projs (approved):

PR #2013, opened 2026-08-31
fix: fetch category for getting iconset

### LINKED ISSUE #1677: [Table Properties] Add icons for each column name and type on column property page
### Describe the bug

Open VSC SAS Extension base on Main branch. Run some SAS code to create a table. Open column property page, it would be a shortcut to identify a column property to use icons. 

### Steps to reproduce

1.Open VSC
2.New a .sas file
3.Type some code 
4.run code
5.Open column property page

### SAS code (if applicable)

```SAS
data work.test;
    input Date: Date9. Country :$13. Actual: 10. Name: $1000.;
    datalines;
08May2017 USA 1000000 test
;
```

### Expected behavior

It would be a helpful to use icons to identify a column property

### Screenshots

Image

### Client OS

_No response_

### Extension version

v1.17.0

### SAS version

_No response_

### Additional context

_No response_

VisheshMalik04:
**Summary:**
Updated format category resolution in the Compute/Viya data provider to use the Compute Informat endpoint when determining format categories for column icon classification. The corresponding `/informats/{name}` endpoint returned the category information required for icon classification.
 
This change retrieves the category from the Informat API and uses it to populate `formatCategory`, allowing the Data Viewer and Table Properties views to display the same format-based icons as SAS Studio.

**Testing:**
Tested against a Viya Compute connection.
Verified category resolution using the Informat endpoint for:
- DATE
- TIME
- DATETIME
- DOLLAR
- COMMA
- PERCENT
- DDMMYY
- MMDDYY
- HEX
- E

Verified that the correct icons are displayed in Data Viewer and Table Properties for:
- Date columns
- Time columns
- Datetime columns
- Currency columns
- Numeric columns
- Character columns
Confirmed existing behavior for columns without an associated format remains unchanged.

**TODOs:**


client/src/connection/formatCategory.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  /**
+   * Returns a map of base format name (e.g. "DOLLAR") to category. Formats whose category could
+   * not be determined map to "" so callers fall back to type based icons.
+   */
+  public async resolve(
+    formats: Array<{ name?: string } | string | undefined>,
+  ): Promise<Map<string, string>> {
+    const names = [...new Set(formats.map(baseFormatName))].filter(Boolean);
```
Nurquhart, line 34: This line confuses me a little bit. I'm not sure why we have a new set inside of an array. Could we not just do:
**formats.map(baseFormatName).filter(Boolean)**

or just turn the filter and map loops into a single reduce?

client/src/connection/formatCategory.ts
```diff
@@ -0,0 +1,57 @@
...
+          this.categories.set(name, category);
+          resolved.set(name, category);
+        });
+      } catch (error) {
+        // Leave the unresolved names uncached so that a later request can retry them.
+        console.warn("Unable to resolve SAS format categories", error);
+      }
+    }
+
+    return new Map(names.map((name) => [name, resolved.get(name) ?? ""]));
```
Nurquhart, line 61: could we end up just returning 'resolved'? I'm not sure what this new Map... gives us

client/test/webview/TablePropertiesViewer.test.ts
```diff
@@ -15,25 +15,6 @@ function safeRequire(paths: string[]): AnyModule {
   assert.fail(`Unable to load module.\nTried:\n${errors.join("\n")}`);
```
Nurquhart, line 15: There are errors in this file for me locally, is that the same for you?

client/src/connection/rest/RestLibraryAdapter.ts
```diff
@@ -172,7 +183,59 @@ class RestLibraryAdapter implements LibraryAdapter {
...
+
+  private async requestFormatCategories(
+    formatNames: string[],
+  ): Promise<Record<string, string>> {
+    const entries = await Promise.all(
+      formatNames.map(async (formatName) => {
+        let informatResp;
+
+        try {
+          informatResp = await this.InformatsApi.getInformat({
```
Nurquhart: So do we make one network request for each of the format names passed in??

Nurquhart (commented):
So it seems like for non viya/rest connections, we submit sas code to give us the correct formats. Which you mention in the comments is actually the same thing the Compute service does to get them. 

But weirdly, the viya/rest code that actually uses the Compute service to fetch the formats (getInformat...) seems more complicated. I'm wondering what benefits the endpoint gives us? I only ask because I feel like it would be easier to maintain a single method for getting these formats, and if we can do it consistently and efficiently by running SAS code, maybe we just use that for all our connection types?

client/src/connection/rest/RestLibraryAdapter.ts
VisheshMalik04: Yes, currently we make one Informat API request per distinct format encountered during category resolution. The list is deduplicated before lookup, and results are cached for the lifetime of the session, so subsequent lookups for the same format do not generate additional requests.

In practice I would expect tables to use a relatively small set of distinct formats (DATE, TIME, DATETIME, COMMA, DOLLAR, etc.)

client/src/connection/formatCategory.ts
VisheshMalik04: The Set is intentional because I wanted to deduplicate format names before resolution. Multiple columns can share the same base format (for example several DATE columns), and without the deduplication we'd potentially perform repeated lookups for the same format. The filter(Boolean) is only removing empty values after normalization.

client/test/webview/TablePropertiesViewer.test.ts
VisheshMalik04: I am not seeing this locally and the project compiles successfully for me.

client/src/connection/formatCategory.ts
VisheshMalik04: I looked into returning resolved directly, but resolved doesn't always contain an entry for every requested format (for example when a lookup fails or a category can't be determined). The final new Map(...) ensures all requested formats are present in the result and unresolved formats consistently map to "", which is relied upon by the callers and existing tests.

VisheshMalik04:
> So it seems like for non viya/rest connections, we submit sas code to give us the correct formats. Which you mention in the comments is actually the same thing the Compute service does to get them.
> 
> But weirdly, the viya/rest code that actually uses the Compute service to fetch the formats (getInformat...) seems more complicated. I'm wondering what benefits the endpoint gives us? I only ask because I feel like it would be easier to maintain a single method for getting these formats, and if we can do it consistently and efficiently by running SAS code, maybe we just use that for all our connection types?

For Viya connections, we've generally followed the principle of using the Compute APIs wherever possible rather than submitting SAS code ourselves. The SAS 9 implementation uses SAS code because there isn't an equivalent API available there.

client/src/connection/formatCategory.ts
Nurquhart: ah okay i see that makes sense, would you mind adding a quick comment explaining why its needed to avoid duplicates and empty values? That way the next person who works on this is aware of the potential edge cases.

Also, nitpick, but below when you are defining **resolved** and **missing**, I am wondering if we can knock that out in one loop, instead of three different loops (.filter, then .map, then another .filter). Something like this maybe? If not, its probably fine.


```
// a map of the provided formats, and their corresponding categories
    const resovled = new Map<string, string>();
    // missing format categories that need to be fetched
    const missing: string[] = [];
    names.forEach((formatName) => {
      if (this.categories.has(formatName)) {
        resovled.set(formatName, this.categories.get(formatName) ?? "");
      } else {
        missing.push(formatName);
      }
    });
```

client/src/connection/formatCategory.ts
Nurquhart: Instead of the additional looping and Map, can we just add something to the catch block so that we ensure resolved is always in the correct state?

```
    // Leave the unresolved names uncached so that a later request can retry them.
        console.warn("Unable to resolve SAS format categories", error);
        missing.forEach((name) => {
          resovled.set(name, "");
        });
```

Nurquhart: Or another suggestion would be to keep the "new Map(names.map(...)", but
just to ditch **resolved** all together and sub in this.categories wherever it is used? Sorry for another nitpick, it just seems odd to keep track of this local variable if we aren't confident that its completely correct

Nurquhart:
> > So it seems like for non viya/rest connections, we submit sas code to give us the correct formats. Which you mention in the comments is actually the same thing the Compute service does to get them.
> > But weirdly, the viya/rest code that actually uses the Compute service to fetch the formats (getInformat...) seems more complicated. I'm wondering what benefits the endpoint gives us? I only ask because I feel like it would be easier to maintain a single method for getting these formats, and if we can do it consistently and efficiently by running SAS code, maybe we just use that for all our connection types?
> 
> For Viya connections, we've generally followed the principle of using the Compute APIs wherever possible rather than submitting SAS code ourselves. The SAS 9 implementation uses SAS code because there isn't an equivalent API available there.

I'm fine to keep the two different implementations (rest vs code submission), I was just kinda curious if you knew of anything the endpoint specifically gives us, or if they have the same end result as the code submission. Having FormatCategoryCache take in a fetching function gives us some added flexibility moving forward depending on where it is being used, but it does just require more maintenance and additional code everywhere it is consumed (which leaves us open to bugs like "it works with a rest profile, but acts differently in a sas9 profile"). Just a trade off of flexibility vs complexity.

client/src/connection/formatCategory.ts
VisheshMalik04: Added a comment explaining the deduplication/empty value handling and refactored the cache-hit/missing detection into a single pass over the format names. The final new Map() remains because callers and tests expect unresolved formats to map to "" rather than be absent from the result.

client/src/connection/formatCategory.ts
VisheshMalik04: I think we still need the local resolved map because fetchCategories() can trigger a reconnect, which clears this.categories during the lookup. resolved preserves the categories already resolved for the current request even if the underlying cache is cleared. I do like the idea of populating the unresolved entries with "" in the error path though, since that would let us simplify the final return logic while still preserving the retry behavior.

Nurquhart (approved):

client/src/connection/rest/RestLibraryAdapter.ts
```diff
@@ -172,7 +186,67 @@ class RestLibraryAdapter implements LibraryAdapter {
...
+              formatName,
+            }),
+            this.InformatsApi.getInformat({
+              sessionId: this.sessionId,
+              informatName: formatName,
+            }),
+          ]);
+
+          const category =
+            formatResp?.data?.category || informatResp?.data?.category || "";
```
ksurajofficial: It looks like if we have `formatResp?.data?.category`, we don't need the `informatResp?.data?.category` lookup. Could we try `getFormat()` first and only call `getInformat()` when the category isn't available? This would help reduce unnecessary network calls while still keeping the fallback.

client/src/connection/rest/RestLibraryAdapter.ts
```diff
@@ -172,7 +186,67 @@ class RestLibraryAdapter implements LibraryAdapter {
...
+  private async fetchFormatCategories(
+    formatNames: string[],
+  ): Promise<Record<string, string>> {
+    const categories = await this.requestFormatCategories(formatNames);
+
+    // An expired session makes every format look unknown, and those empty results would be
+    // cached for the rest of the session. Reconnect once and retry before accepting them.
+    if (!Object.values(categories).some((category) => category)) {
+      await this.connect();
+      return await this.requestFormatCategories(formatNames);
```
ksurajofficial: If we are concerned about whether the session is expired or not, should retryOnFail() serve the purpose? Something like:

const categories = await this.retryOnFail(
  async () => await this.requestFormatCategories(formatNames),
);

My point is - we are saying if all categories are empty, reconnect even though the request may have succeeded. An empty result does not necessarily mean the session expired. It can also mean the formats are custom formats not resolvable through that endpoint, or the server simply returned no category.

client/src/connection/rest/RestLibraryAdapter.ts
```diff
@@ -25,7 +29,12 @@ const requestOptions = {
 
 class RestLibraryAdapter implements LibraryAdapter {
   protected dataAccessApi: ReturnType<typeof DataAccessApi>;
+  protected InformatsApi: ReturnType<typeof InformatsApi>;
+  protected FormatsApi: ReturnType<typeof FormatsApi>;
```
ksurajofficial: Trivial - these should be in camelCase. But, if you have any further changes, then you can consider these as well.

client/src/connection/rest/RestLibraryAdapter.ts
VisheshMalik04: Good point. I'll update the implementation to avoid the unnecessary extra request.

client/src/connection/rest/RestLibraryAdapter.ts
VisheshMalik04: This reconnect heuristic was added to avoid caching empty results from an invalid session, but an empty category response doesn't necessarily indicate session expiration. With the current /formats → /informats fallback, it's possible for both endpoints to legitimately return an empty category, in which case we should simply fall back to the generic icon. Since actual request failures are handled separately, I've removed the reconnect-on-empty logic.

ksurajofficial (approved):

sarattesting-spec:
Tested the Issue against the build 2013, I found all tested scenarios has been fixed.

Scenarios Tested:
In Properties --> Column Tab:

1. Tested with Numeric and Character data types with and without associating formats to it and found that, the TYPE column is always displayed with Numeric and Character with correct icon. Also the tooltip is displayed correctly while hovering over on the icon.
2. Tested with Character format ($10), Numeric formats (Comma12., Currency[DOLLAR15.2], Percent[PERCENT8.2], , Date[Date9, MMDDYY10.], Datetime[Datetime20.] and Time[Time8.], E15.6, Hex6) and observe that, the Name tab of column properties shows
column name with Character column Icon for $10 format,
column name with Character column Icon for no associated format,
column name with Currency Icon for DOLLAR15.2 format,
column name with Date Icon for Date9, MMDDYY10. format,
column name with Time Icon for Time8. format,
column name with Datetime Icon for Datetime20. format,
column name with Numeric icon for rest other numeric format and
column name with Numeric icon for no associated format.
3. Tested the tooltip for all these icons and found these are displayed correctly.
I.e. For Character Icon -- Tooltip is Character
For Numeric Icon -- Tooltip is Numeric
For Date Icon -- Tooltip is Numeric
For Datetime Icon -- Tooltip is Numeric
For Time Icon -- Tooltip is Numeric
For Currency Icon -- Tooltip is Currency
4. Tested with changing the color theme of VS Code with one light and one dark shade, and found that, the icons are still displayed without any issue.
Click on Table name --> Data Tab:
5. Tested that, column headings are displayed the column name with correct icon as per the column name in properties tab. Like Currency Icon for Dollar format, Date Icon for Date format, Datetime icon for Datetime format, Time icon for Time format, Numeric icon for Comma, percent etc., Character icon for Char format.
6. Tested that, the tooltip of each icon in the column heading is displayed correctly as it is displayed in column properties tab.
7. Tested with changing the color theme of VS Code with one light and one dark shade, and found that, the icons are still displayed without any issue.

Performed smoke testing on this build and found no issues.
https://rndconfluence.sas.com/spaces/VSCSASEXT/pages/976499370/Smoke+Test+Scenarios

Hence, Moving the ticket to "Test Complete".

Platform Tested:
Viya4 (Daily.pgc)
SAS9 (IOM)
SAS9 (Local)

VS Code Version Tested:
Version: 1.135.0
Commit: 08d4889f9ec4a1685d257b9b95de036c8e1ce1e5
Date: 2026-08-25T14:26:52Z
Electron: 42.8.1
ElectronBuildId: 14906494
Chromium: 148.0.7778.280
Node.js: 24.18.1
V8: 14.8.178.38-electron.0
@github/copilot: 1.0.81-0
@github/copilot-sdk: 1.0.11
OS: Windows_NT x64 10.0.26200

SAS Extension VSIX version:
Identifier: sas.sas-lsp
Version: 1.20.0
Source: VSIX (2013) - https://github.com/sassoftware/vscode-sas-extension/actions/runs/33715861641

sarattesting-spec:
Verified the reported issue in the latest build from main branch, I found all tested scenarios has been fixed.

Platform Tested:
Viya4 (Daily.pgc)
SAS9 (IOM)
SAS9 (Local)

VS Code Version Tested:
Version: 1.136.0
Commit: 520fb30b2d3d324b4cb2342f6e88e2cd93751de1
Date: 2026-09-02T09:13:42Z
Electron: 42.10.0
ElectronBuildId: 15109253
Chromium: 148.0.7778.280
Node.js: 24.18.1
V8: 14.8.178.38-electron.0
@github/copilot: 1.0.81-0
@github/copilot-sdk: 1.0.11
OS: Windows_NT x64 10.0.26200

SAS Extension VSIX version:
Identifier: sas.sas-lsp
Version: 1.20.0
Source (Main Branch): VSIX (855) - https://github.com/sassoftware/vscode-sas-extension/actions/runs/33885735863

PR #2014, opened 2026-08-31
fix: honor workspace scope for sas connection profile operations

### LINKED ISSUE #988: SAS Profiles commands cannot handle workspace settings
**Describe the bug**
Following #986, I set up a workspace settings.json file in which I copied the SAS profile settings from the global file
"SAS.connectionProfiles": {
        "activeProfile"; "Profile_One",
        "profiles": {
             "Profile_One" : {. . .},
             "Profile_Two" : {. . .}      
        }
}
With that, I was able to have the 'activeProfile' (Profile_One) of that file stay independent of all other workspaces.
However, the SAS.addProfile and SAS.switchProfile keep targeting the user settings.json (I suspect it applies to SAS.updateProfile and SAS.deleteProfile too). As a consequence, I cannot switch profiles with the UI or commands, I need to directly edit the workspace settings.json, which is unwieldly.

**Steps to reproduce**
1. Have/Create at least 2 sas connection profiles
2. Create a workspace settings.json and paste SAS.connectionProfiles settings from the user file in the settings section, and save
3. Try to switch profiles from the UI (bottom left), or to add profiles
4. Observe it does not affect the workspace settings.json, or the UI

**Expected behavior**
When the workspace settings.json exists (and contains SAS.connectionProfiles), the commands should target it instead of the user settings.json file.

**Screenshots**
NA

Client OS: Windows 10
Extension version: v.1.9.0
SAS version: 9.4 (IOM remote)

**Additional context**
#987

VisheshMalik04:
**Summary:**
This change updates SAS connection profile management to honor the VS Code configuration hierarchy when resolving and updating SAS.connectionProfiles.

Previously, profile operations such as Switch Profile, Add Profile, Update Profile, and Delete Profile always targeted the User settings scope, even when SAS.connectionProfiles was defined in workspace settings.

With this change, the extension resolves profiles from the highest-precedence configuration scope that contains profile definitions. Workspace folder settings take precedence over workspace settings, which take precedence over User settings. Empty profile stores (for example {} or { activeProfile: "", profiles: {} }) are treated as unconfigured and fall back to the next lower-precedence scope. Profile operations are written back to the highest-precedence configuration scope that contains profile definitions, ensuring reads and writes occur in the same configuration scope, aligning profile management behavior with VS Code's settings hierarchy.
Note: A workspace profile store that contains one or more profiles is considered authoritative, even when activeProfile is empty. Fallback to lower-precedence scopes occurs only when the profile store is empty (for example {} or { activeProfile: "", profiles: {} }).
Settings preference honored by VS code- https://code.visualstudio.com/docs/configure/settings#_workspace-settings

**Testing:**
Tested the following scenarios:
 
1. **Single-folder workspace**
- Added `SAS.connectionProfiles` to `.vscode/settings.json`.
- Verified profiles were discovered correctly.
- Verified Switch Profile updated `activeProfile` in the same `.vscode/settings.json`.
- Verified Add, Update, and Delete profile operations updated the workspace settings file.
 
2. **User settings fallback**
- Removed `SAS.connectionProfiles` from workspace settings.
- Added `SAS.connectionProfiles` to `%APPDATA%\Code\User\settings.json`.
- Verified profiles were loaded from User settings.
- Verified profile operations updated User settings.

3. **Workspace overrides User settings**
- Defined `SAS.connectionProfiles` in both workspace and User settings.
- Verified workspace profiles were used in preference to User profiles.
- Verified profile operations updated the workspace configuration rather than User settings.

4. **Multi-root workspace**
- Added `SAS.connectionProfiles` to the workspace configuration (`workspace.json` / `.code-workspace`).
- Verified profiles were discovered correctly.
- Verified profile operations updated the workspace configuration scope.

5. **Empty workspace profile store fallback**

- Defined `SAS.connectionProfiles` in User settings.
- Set workspace `SAS.connectionProfiles` to `{}`.
- Verified profiles were loaded from User settings.
- Set workspace `SAS.connectionProfiles` to `{ activeProfile: "", profiles: {} }`.
- Verified profiles were loaded from User settings.
- Verified User `activeProfile` was not modified.

**TODOs:**


client/test/components/profile/profile.test.ts
```diff
@@ -1,6 +1,7 @@
 import { ConfigurationTarget, workspace } from "vscode";
```
Nurquhart, line 1: I am seeing a typescript error in this file, are you seeing the same?

VisheshMalik04: Fixed. I removed the as never assertion and updated the configuration mock to avoid the type mismatch. ESLint, compile, and profile tests are now passing.

client/src/components/profile.ts
```diff
@@ -224,16 +232,61 @@ export class ProfileConfig {
...
+   * Determines which configuration scope owns the connection profiles.
+   * Workspace folder settings take precedence over workspace settings, which
+   * take precedence over user (global) settings.
+   *
+   * @returns the {@link ConfigurationTarget} that profile updates should be written to
+   */
+  getConfigurationTarget(): ConfigurationTarget {
+    const inspected = this.inspectProfileSetting();
+
+    if (inspected?.workspaceFolderValue) {
```
ksurajofficial, line 262: An empty "SAS.connectionProfiles": {} should not be considered a valid profile configuration. Could you check if this condition evaluates to true in that case?

ksurajofficial (commented):
Thought - if Workspace has "SAS.connectionProfiles": {} and User has valid profiles, should the User setting take precedence, or should the empty Workspace setting take precedence?

Worth having a test case for this scenario if it makes sense?

client/src/components/profile.ts
VisheshMalik04: I tested this scenario. When SAS.connectionProfiles is set to {} in the workspace settings and valid profiles exist in the user settings, the implementation currently falls back to the user settings. Since the workspace configuration contains no profiles, it is treated as empty and the user-scoped profiles become the effective configuration

Nurquhart:
Look good to me, I'll let Suraj follow up to make sure his thoughts were addressed

ksurajofficial:
A few more suggestions/observations:
 
1. Empty connectionProfiles in workspace settings
In User settings.json, if we edit SAS.connectionProfiles to an empty object: "SAS.connectionProfiles": {}, or remove it altogether, on saving settings.json, it automatically falls back to:
 
   "SAS.connectionProfiles": {
     "activeProfile": "",
     "profiles": {}
   }
 
It would be good to have the same behavior for the workspace settings.json as well.
 
2. Empty workspace profiles should not modify User settings
If we have valid profiles in the workspace settings.json and later change it to:
 
   "SAS.connectionProfiles": {
     "activeProfile": "",
     "profiles": {}
   }
 
the activeProfile in the User settings.json is currently set to "" as well. I don't think the User setting should be modified in this case.
 
3. How should the precedence work?
One important observation that may need some attention before this fix goes live: as we know, the workspace settings.json takes precedence over the User settings.json.
 
Suppose I have intentionally segregated my connections between the two settings files - some connections in the workspace and others in User settings. I may want the User settings.json activeProfile to be used for signing in to a server for certain operations.
 
However, with the current precedence understanding, if a workspace settings.json contains profiles, I won't be able to access the User activeProfile unless I remove all the profiles from the workspace settings.
 
Do we really want users to have to delete their workspace connections just to use a profile from User settings?
 
To me, it might make more sense to use the activeProfile value to determine which scope should take precedence. For example, if I want to use a profile from User settings, I could simply set activeProfile to "" in the workspace settings.json. This would allow the User activeProfile to be used without requiring me to delete any of the connections stored in the workspace settings.
 
what do you think @smorrisj @Nurquhart @snlwih @davidbaileync ?

Nurquhart:
> A few more suggestions/observations:
> 
> 1. Empty connectionProfiles in workspace settings
>    In User settings.json, if we edit SAS.connectionProfiles to an empty object: "SAS.connectionProfiles": {}, or remove it altogether, on saving settings.json, it automatically falls back to:
>    "SAS.connectionProfiles": {
>    "activeProfile": "",
>    "profiles": {}
>    }
> 
> It would be good to have the same behavior for the workspace settings.json as well.
> 
> 2. Empty workspace profiles should not modify User settings
>    If we have valid profiles in the workspace settings.json and later change it to:
>    "SAS.connectionProfiles": {
>    "activeProfile": "",
>    "profiles": {}
>    }
> 
> the activeProfile in the User settings.json is currently set to "" as well. I don't think the User setting should be modified in this case.
> 
> 3. How should the precedence work?
>    One important observation that may need some attention before this fix goes live: as we know, the workspace settings.json takes precedence over the User settings.json.
> 
> Suppose I have intentionally segregated my connections between the two settings files - some connections in the workspace and others in User settings. I may want the User settings.json activeProfile to be used for signing in to a server for certain operations.
> 
> However, with the current precedence understanding, if a workspace settings.json contains profiles, I won't be able to access the User activeProfile unless I remove all the profiles from the workspace settings.
> 
> Do we really want users to have to delete their workspace connections just to use a profile from User settings?
> 
> To me, it might make more sense to use the activeProfile value to determine which scope should take precedence. For example, if I want to use a profile from User settings, I could simply set activeProfile to "" in the workspace settings.json. This would allow the User activeProfile to be used without requiring me to delete any of the connections stored in the workspace settings.
> 
> what do you think @smorrisj @Nurquhart @snlwih @davidbaileync ?

I am okay with the distinction that: if you have workspace profiles, we will reference those and never your Users settings.json, but I do get your point @ksurajofficial. It could be nice to default back to the Users settings.json if there is nothing active in the workspace. My only concern would be the added complexity, and if it really warrants that use case. Reading the bug, it seems like they only want to use the workspace settings.json. The idea being that one workspace can be pinned to one profile, and separate workspace can be pinned to a different one. I wonder if anyone would go to the extent of creating a workspace settings.json, and then continue to switch back and forth with Users settings.json.

For example, what do we do in the UI if neither place has an activeProfile selected? Do we show them all of the profiles (Users and workspace) in our UI? Then if they choose a workspace file as their activeProfile, do we then only display the workspace profiles? Or do we always show all the profiles in the UI?

A more specific example being if UserProfile1 is active at the User's level, then they have a separate repository with a workspace settings.json where WSProfile1 is active. What happens if in that workspace directory, they update their active profile to UserProfile2 (clearing the workspace activeProfile, and defaulting to a new Users activeProfile). Would they be surprised that now all of their repositories got switched from UserProfile1 to UserProfile2? If I had a workspace file, I think I would be under the impression that any changes i make only effect that workspace. So I may be confused that an action could potentially effect my User settings.

I would be okay with either approach, but would personally air on the side of caution and less complexity unless someone specifically asks for it. Either way we should make sure whatever we decide is documented

ksurajofficial:
> > A few more suggestions/observations:
> > 
> > 1. Empty connectionProfiles in workspace settings
> >    In User settings.json, if we edit SAS.connectionProfiles to an empty object: "SAS.connectionProfiles": {}, or remove it altogether, on saving settings.json, it automatically falls back to:
> >    "SAS.connectionProfiles": {
> >    "activeProfile": "",
> >    "profiles": {}
> >    }
> > 
> > It would be good to have the same behavior for the workspace settings.json as well.
> > 
> > 2. Empty workspace profiles should not modify User settings
> >    If we have valid profiles in the workspace settings.json and later change it to:
> >    "SAS.connectionProfiles": {
> >    "activeProfile": "",
> >    "profiles": {}
> >    }
> > 
> > the activeProfile in the User settings.json is currently set to "" as well. I don't think the User setting should be modified in this case.
> > 
> > 3. How should the precedence work?
> >    One important observation that may need some attention before this fix goes live: as we know, the workspace settings.json takes precedence over the User settings.json.
> > 
> > Suppose I have intentionally segregated my connections between the two settings files - some connections in the workspace and others in User settings. I may want the User settings.json activeProfile to be used for signing in to a server for certain operations.
> > However, with the current precedence understanding, if a workspace settings.json contains profiles, I won't be able to access the User activeProfile unless I remove all the profiles from the workspace settings.
> > Do we really want users to have to delete their workspace connections just to use a profile from User settings?
> > To me, it might make more sense to use the activeProfile value to determine which scope should take precedence. For example, if I want to use a profile from User settings, I could simply set activeProfile to "" in the workspace settings.json. This would allow the User activeProfile to be used without requiring me to delete any of the connections stored in the workspace settings.
> > what do you think @smorrisj @Nurquhart @snlwih @davidbaileync ?
> 
> I am okay with the distinction that: if you have workspace profiles, we will reference those and never your Users settings.json, but I do get your point @ksurajofficial. It could be nice to default back to the Users settings.json if there is nothing active in the workspace. My only concern would be the added complexity, and if it really warrants that use case. Reading the bug, it seems like they only want to use the workspace settings.json. The idea being that one workspace can be pinned to one profile, and separate workspace can be pinned to a different one. I wonder if anyone would go to the extent of creating a workspace settings.json, and then continue to switch back and forth with Users settings.json.
> 
> For example, what do we do in the UI if neither place has an activeProfile selected? Do we show them all of the profiles (Users and workspace) in our UI? Then if they choose a workspace file as their activeProfile, do we then only display the workspace profiles? Or do we always show all the profiles in the UI?
> 
> A more specific example being if UserProfile1 is active at the User's level, then they have a separate repository with a workspace settings.json where WSProfile1 is active. What happens if in that workspace directory, they update their active profile to UserProfile2 (clearing the workspace activeProfile, and defaulting to a new Users activeProfile). Would they be surprised that now all of their repositories got switched from UserProfile1 to UserProfile2? If I had a workspace file, I think I would be under the impression that any changes i make only effect that workspace. So I may be confused that an action could potentially effect my User settings.
> 
> I would be okay with either approach, but would personally air on the side of caution and less complexity unless someone specifically asks for it. Either way we should make sure whatever we decide is documented

Yeah, I agree. According to the bug, if SAS.connectionProfiles is present in the workspace settings, we should use the workspace profiles and ignore the User settings.

My concern was mainly the case where both exist and I want to use a profile from the User settings. With the current approach, I would need to either remove the workspace profiles or copy that profile into the workspace. But considering the added complexity, if this use case is really needed, we can always revisit it later. For now, I agree that keeping the workspace and User settings isolated is simpler and avoids unexpected cases.

One more point: if we are intentionally keeping the Workspace and User settings separate, and Workspace takes precedence, then whenever profiles are defined in the Workspace, I suggest we should show the Workspace connections in the UI as well, to keep the behavior consistent.

Nurquhart:
I haven't been able to test with this yet, but I think everything looks how I expected. I will let Suraj take a look to make sure his concerns have been addressed

client/test/components/profile/profile.test.ts
```diff
@@ -1000,4 +1018,539 @@ describe("Profiles", async function () {
...
+      stubConfiguration({
+        globalValue: userSetting,
+        workspaceValue: {},
+      });
+
+      expect(profileConfig.getActiveProfile()).to.equal("");
+      expect(Object.keys(profileConfig.getAllProfiles())).to.have.length(0);
+    });
+
+    it("falls back to user profiles when workspace setting is an empty normalized structure", function () {
```
ksurajofficial: If by “workspace setting is empty” we mean `"SAS.connectionProfiles": {}`, then it should not fall back to the user settings. Even if the setting is empty, the presence of `"SAS.connectionProfiles"` in the workspace settings means that the workspace settings takes precedence over the user settings.

client/test/components/profile/profile.test.ts
```diff
@@ -1000,4 +1018,539 @@ describe("Profiles", async function () {
...
+
+      const [, value, target] = updateStub.firstCall.args;
+      expect(target).to.equal(ConfigurationTarget.Global);
+      expect(Object.keys(value.profiles)).to.have.members([
+        "userProfile",
+        "newProfile",
+      ]);
+    });
+
+    it("creates a new profile in workspace settings when the workspace setting is empty", async function () {
```
ksurajofficial, line 1449: By empty workspace, if you mean the settings.json has SAS.connectionProfiles tag but nothing set, then please reflect the same here and in the other cases.

client/test/components/profile/profile.test.ts
```diff
@@ -1000,4 +1018,539 @@ describe("Profiles", async function () {
...
+    beforeEach(async () => {
+      sandbox = createSandbox();
+      await initProfile();
+    });
+
+    afterEach(() => {
+      sandbox.restore();
+    });
+
+    it("creates a new profile in workspace settings when a workspace is open but has no connectionProfiles setting", async function () {
```
ksurajofficial, line 1432: For User vs Workspace precedence, I don't think there has to be anything to do with workspace being opened.  Shouldn't the tag "SAS.connectionProfiles" be enough?

client/test/components/profile/profile.test.ts
```diff
@@ -1000,4 +1018,539 @@ describe("Profiles", async function () {
...
+      const isValid = profileConfig.validateSettings();
+      expect(updateStub.callCount).to.be.greaterThan(0);
+      expect(isValid).to.equal(true);
+      const [key, value, target] = updateStub.firstCall.args;
+      expect(key).to.equal(EXTENSION_DEFINE_PROFILES_CONFIG_KEY);
+      expect(value).to.eql({ activeProfile: "", profiles: {} });
+      expect(target).to.equal(ConfigurationTarget.WorkspaceFolder);
+    });
+
+    it("falls back to user profiles and normalizes a stale workspace folder setting", function () {
```
ksurajofficial, line 1322: It should not fall back to the user settings.json. If the workspace has the stale structure, we normalize it while keeping the "SAS.connectionProfiles" tag. This means the workspace takes precedence over the user settings because it now contains the "SAS.connectionProfiles" tag.

client/test/components/profile/profile.test.ts
```diff
@@ -1000,4 +1018,539 @@ describe("Profiles", async function () {
...
+      await profileConfig.upsertProfile("workspaceProfile3", {
+        connectionType: ConnectionType.Rest,
+        endpoint: "https://workspace-host3.sas.com",
+      });
+
+      const [, , target] = updateStub.firstCall.args;
+      expect(target).to.equal(ConfigurationTarget.Workspace);
+    });
+
+    it("normalizes a missing/empty workspace setting when no scope has profiles", function () {
```
ksurajofficial, line 1267: If there is no "SAS.connectionProfiles" tag in both user and workspace, then only user's setting should get normalized.

client/test/components/profile/profile.test.ts
```diff
@@ -1000,4 +1018,539 @@ describe("Profiles", async function () {
...
+        },
+      });
+
+      expect(profileConfig.getActiveProfile()).to.equal("");
+      expect(Object.keys(profileConfig.getAllProfiles())).to.have.members([
+        "testConn",
+      ]);
+    });
+
+    it("does not modify user settings when workspace configuration is empty", function () {
```
ksurajofficial, line 1227: Please clarify if you mean - "does not modify user settings when workspace has "SAS.connectionProfiles" tag even if no profiles"

client/test/components/profile/profile.test.ts
ksurajofficial, line 1194: By empty, if you mean this - "SAS.connectionProfiles": {}. Then, it should use workspace.

ksurajofficial (commented):
The code looks good to me as well. It’s mainly the test cases that I feel need another look.

Apologies for all the comments, but let’s make the test case descriptions more clear. For example, if a test covers switching between profiles, or adding, deleting, or modifying profiles, please explicitly mention that these actions are performed through the UI. This will make it clear to anyone revisiting these tests later that the changes are being made through the UI, rather than by directly modifying settings.json.

Also, let’s clearly differentiate between an empty/missing workspace or user setting and SAS.connectionProfiles: {}, where the SAS.connectionProfiles tag is present but contains no profiles. These are different scenarios and should be reflected clearly in the test case descriptions, especially since the presence of the tag determines which setting takes precedence.

I’m confident the functionality is working as we discussed; it’s mainly the test case descriptions that need more clarity.

client/test/components/profile/profile.test.ts
VisheshMalik04: Done

ksurajofficial (commented):
Looks good to me now. One question: what about the documentation changes? Aren’t we planning to handle those as part of the same PR?

Nurquhart:
I'm okay handling documentation changes in a separate PR to get this in ASAP to start testing, unless you already have it ready to be added immedietly

ksurajofficial:
Yeah, I think that would take some additional time, especially since we’d need to get it reviewed by the technical writer as well.

ksurajofficial (approved):
Approving, lets get this moved to testing.

Nurquhart (approved):

Nurquhart (changes requested):
holding this until 1.22.0

Nurquhart (approved):

Issue #1834, opened 2026-03-24
View image files with VScode internal viewer


oliverwirtz:
### Is your feature request related to a problem? Please describe.

In our clinical trials output documents we use png files to display figures. These files cannot be accessed directly using the SAS extension since the files cannot be handed over to the VScode image viewer. 

### Describe the solution you'd like

I ve developed a feature to download such files locally and display them afterwards using the internal viewer. Can I open a PR for this? 

### SAS version

Viya 4 

### Sensitive information

- [x] I have confirmed that this feature request contains no sensitive information

snlwih:
Hi @oliverwirtz , sure you can open a PR. Guidelines are described in CONTRIBUTING.md and also look at PR #1865 which will provide an update to that file to talk about new signoff requirements that are already effective. Once you have created your PR then please link your PR back to this issue. See https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

snlwih:
Hi @oliverwirtz , is #1891 related to this? That one talks about image files stored in a SAS Content folder or in a directory on the SAS Server file system.

danichiavegatto:
Hi @oliverwirtz, please take a look at our [process documentation](https://github.com/sassoftware/vscode-sas-extension/blob/main/doc/README.md#procedure-for-bugfixes-and-features) for next steps

Nurquhart:
The PR has been merged and tested, but there is a related PR out that changes some of the code (https://github.com/sassoftware/vscode-sas-extension/pull/2006), so I want to wait until this is merged before doing the final testing

PR #2015, opened 2026-08-31
fix: temp image clean up

Nurquhart:
**Issue:**
https://github.com/sassoftware/vscode-sas-extension/issues/1834

**Summary:**
we can't guarantee a seamless disposal... stuff happens. so add a case to the constructor that will clear the image cache as well. that way even if there were left over artifacts,next time you open the extension they will get cleaned up

**Testing:**
1. In a terminal, navigate /Users/<userid>/Library/Application Support/Code/User/globalStorage/sas.sas-lsp (mac)
2. Then open the SAS VS Code extension
3. On a Viya connection, open an image file from either SAS Content or SAS Server. 
4. "ls" in your terminal, notice there should be a "imagePreviewCache" dir
5. 'cd' into that dir
6. notice the image file you just opened
7. now close vs code/wait for it to time out
8. After VS Code closes, that "imagePreviewCache" dir may or may not still have files
9. If it is empty, great! Everything worked how it should (without this extra code)
10. If it isn't empty, open the SAS Extension in VS Code again, wait for the sign in to complete, and verify "imagePreviewCache" should now be empty

**TODOs:**

### DIFF
--- a/client/src/components/ContentNavigator/index.ts
+++ b/client/src/components/ContentNavigator/index.ts
@@ -92,6 +92,11 @@ class ContentNavigator implements SubscriptionProvider {
       `${config.sourceType}ReadOnly`,
       this.contentDataProvider,
     );
+
+    // Remove any image preview cache files left behind by a previous session
+    // (e.g. VS Code was force-closed or the extension host crashed/timed-out
+    // before the tab-close/dispose cleanup could run).
+    void this.purgeImagePreviewCache();
   }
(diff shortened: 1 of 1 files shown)


grantfountain-projs (approved):

noahhowell117 (approved):
Changes look good to me. Was able to verify the fix and see that the cache was deleted as expected when restarting VS Code.

PR #2016, opened 2026-08-31
feat: add support for displaying hidden files in SAS Server

### LINKED ISSUE #1511: Hidden files are not shown in SAS server tree while shown in SAS Studio
### Describe the bug

In VS Code I'm able to see the entire filesystem from root but in SAS Studio but not "hidden" files that start with a dot.  In SAS Studio I can only see my "home" folder and subfolders.

### Steps to reproduce

1. 
2.

### SAS code (if applicable)

```SAS

```

### Expected behavior

Ideally, I would see the same set of SAS server files in both interfaces. 

### Screenshots

Image

Image

### Client OS

Windows 11

### Extension version

v1.12.0

### SAS version

20250603.1748980010666 (Stable 2025.04)

### Additional context

_No response_

ksurajofficial:
**Summary:**
Adds a new display.showHiddenItems option to REST connection profiles to allow hidden files and folders to be displayed in SAS Server file navigation.

The setting defaults to false and is passed through the content adapter factory to RestServerAdapter, which passes the corresponding showAll value to the Compute API.

Added unit tests covering showHiddenItems behavior, including:
- true passes showAll: true to the Compute API.
- false passes showAll: false.
- The setting defaults to false when not configured.
- Hidden items returned by the server are included in the adapter results when enabled.

**Testing:**
1. Add "display": { "showHiddenItems": true } to your SAS Viya (REST) profile.
2. Sign in to that same profile.
3. Open SAS Server and navigate to a folder that contains hidden files or folders.
4. Verify that the hidden files and folders are displayed.
5. Change "showHiddenItems" to false (or remove the display setting), reconnect, and verify that the hidden files and folders are not displayed.

**TODOs:**


client/src/components/profile.ts
```diff
@@ -133,6 +134,14 @@ interface GlobalShortcuts {
   [key: string]: string;
 }
 
+export interface DisplayOptions {
```
Nurquhart, line 137: Curious why you decided to nest showHiddenItems inside another object here? Are there related "display" settings that you think will also be included in here?

ksurajofficial: Yeah, I initially started with a flat structure, but there were a couple of reasons that led me to nest it this way. First, I do see quite a few display-related options in SAS Studio. I’m not sure we’ll need all of them here, but some could definitely be added in the future. Second, for readability, I wanted the code structure to match the structure defined in the package.json.

Nurquhart:
Did we ever get Wilbram's thoughts on where this setting should go? Want to make sure he is okay with it going on the profile connection object before we merge it in

ksurajofficial:
> Did we ever get Wilbram's thoughts on where this setting should go? Want to make sure he is okay with it going on the profile connection object before we merge it in

Not exactly on where this option should go. Since this is profile-specific, similar to Studio, where hidden items should be shown for a particular server/user when the setting is enabled, that was my understanding as well, so I placed the setting at the profile level.

If there’s an alternative way to provide profile-specific settings, then I think it would make sense to revisit where this setting should live. We can wait for Wilbram to be back and get his thoughts on this before we merge.

Nurquhart (approved):

Issue #1303, opened 2024-11-28
Switch language mode from SAS to Python in VSC got error 'Request textDocument/foldingRange failed.' 


Zhirong2022:
**Describe the bug**
Open VSC SAS Extension base on Main branch. Create a .sas file, change language mode from SAS to Python. Type some python code, it will has error 'Request textDocument/foldingRange failed.'.  

**Steps to reproduce**
There are no any python relative extensions installed 

1. Create a .sas file
3. Switch language mode from SAS to Python
4. Type code below
```
print('test')
```

**Expected behavior**
No error 

**Screenshots**
![image](https://github.com/user-attachments/assets/8c264868-1192-4e3e-8259-5eb1930d0d61)

**Environment (please complete the following information):**
Client OS: [e.g. Windows 11]
Extension version: [e.g. v1.12.0]

PR #2028, opened 2026-09-02
fix: avoid SAS server crashes when switching language modes

JacinthaDev:
Bug fix for #1303 

**Summary:**
Scoped SAS language providers to SAS documents only by adding a DocumentSelector ({ language: "sas" }) to capabilities including hover, folding ranges, formatting, completions, semantic tokens, document symbols, and signature help. This prevents the SAS language server from receiving requests for documents that have been switched to another language mode after being removed from documentPool, avoiding potential server crashes.

Also added an integration test covering the language mode switch workflow: verifying SAS features before the switch, switching a SAS document to Python, confirming the server remains responsive for other SAS documents, and verifying SAS features resume when switching the document back to SAS.

**Testing:**
```
npm run compile
npm test
```
Manual Testing

Open a .sas file and verify SAS features (such as hover) are available.
Change the file's language mode from SAS to Python.
Verify the SAS language server remains responsive and does not crash.
Open another .sas file and confirm SAS features are still available.
Change the original file back to SAS and verify SAS features resume.
Expected Result

Switching a .sas file to Python should not cause the SAS language server to crash. SAS language features should continue to work for other SAS documents, and the original file should regain SAS features when switched back to SAS mode.

**TODOs:**

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -4,13 +4,19 @@ import {
   CancellationToken,
   CodeActionKind,
   CompletionItemKind,
+  CompletionRegistrationOptions,
   CompletionTriggerKind,
   Connection,
   DidChangeConfigurationParams,
   DidChangeWatchedFilesParams,
+  DocumentFormattingRegistrationOptions,
   DocumentHighlightParams,
+  DocumentOnTypeFormattingRegistrationOptions,
+  DocumentSelector,
   DocumentSymbol,
+  DocumentSymbolRegistrationOptions,
   ExecuteCommandParams,
+  HoverRegistrationOptions,
   InitializeResult,
   Location,
   Position,
@@ -22,6 +28,7 @@ import {
   RenameParams,
   ResultProgressReporter,
   SemanticTokensRequest,
+  SignatureHelpRegistrationOptions,
   TextDocumentPositionParams,
   TextDocumentSyncKind,
   Unregistration,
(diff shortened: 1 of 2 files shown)


server/src/server.ts
```diff
@@ -69,37 +76,56 @@ export const runServer = (
...
+    };
+    const hoverOpts: HoverRegistrationOptions = {
+      documentSelector: sasSel,
+      workDoneProgress: true,
+    };
+    const completionOpts: CompletionRegistrationOptions = {
+      documentSelector: sasSel,
+      triggerCharacters: _pyrightLanguageProvider.getClientCapabilities()
+        .hasVisualStudioExtensionsCapability
+        ? [".", "[", "@", '"', "'", " "]
```
ksurajofficial, line 100: Looks like since these options don't seem to be used in many places, could we define them directly inside capabilities like earlier and just keep the shared sasSel as a variable to pass to the respective options? I think this would make the code easier to read and reduce unnecessary indirection. completionOpts could be the one exception, considering the conditional logic involved. Thought?

ksurajofficial:
Could we also cover the typing scenario from the original issue? The reported crash occurs after switching the document from SAS to Python and then typing Python code, which triggers a textDocument/foldingRange request. The current test switches to Python and checks hover, but I believe it doesn't exercise the typing/folding scenario. If possible, it would be good to add tests that reproduce these and verify the server remains responsive.

PR #2030, opened 2026-09-03
fix: library cache refresh

### LINKED ISSUE #981: Temp libs still visible in the Libraries pane on reconnect
**Describe the bug**
When you assign a temporary library in a SAS Session, and it times out, the libraries pane doesn't clear but still displays the libraries, once I connect to a new session the library pane doesn't automatically refresh but rather the temporary library is still displayed there, but of course I can not interact with it any longer.

**Steps to reproduce**

1. Connect a SAS session
2. Assign a temporary lib, e.g. libname test '/sharedata/home/username'; and look at it in the Libraries pane
3. Wait for the session to time out
4. Reconnect to SAS
5. Take a look at the Libraries pnae

**Expected behavior**
If I reconnect to a SAS session, even if it is the same profile, the libraries pane should be fully refreshed.

**Screenshots**
None

Client OS: Windows 11
Extension version: v1.9.0
SAS version: [SAS Viya 2024.04]

**Additional context**
None

Nurquhart:
**Issue:**
https://github.com/sassoftware/vscode-sas-extension/issues/981

**Summary:**
when your session has timed out, and a new session is created, we need to refresh the libraries pane automatically.

The architecture made this a little interesting... I wanted to trigger the refresh after a retryOnFail when a new connection needed to be created. But the refresh function needed to be defined at the LibraryDataProvider level so it could fire off the _onDidChangeTreeData event. So in order to call it in the Adapters, I had to pass it through where the LibraryAdapterFactory gets called and created.

**Testing:**

1. Create a new SAS Program with code that will create a temp library:

libname ATEST '/tmp';

data ATEST.test;
set sashelp.class;
run;

2. Wait until your compute context times out (work on something else for 30 minutes, then come back)
3. Expand SASHELP, or some other library
4. Toast will appear indicating that it needs to reconnect to a compute session
5. After a reconnect happens, the Library Pane should refresh and the Temp Library should disappear

**TODOs:**

### DIFF
--- a/client/src/components/LibraryNavigator/LibraryAdapterFactory.ts
+++ b/client/src/components/LibraryNavigator/LibraryAdapterFactory.ts
@@ -6,14 +6,17 @@ import { ConnectionType } from "../profile";
 import { LibraryAdapter } from "./types";
 
 class LibraryAdapterFactory {
-  public create(connectionType: ConnectionType): LibraryAdapter {
+  public create(
+    connectionType: ConnectionType,
+    onConnect?: () => void,
+  ): LibraryAdapter {
     switch (connectionType) {
       case ConnectionType.IOM:
       case ConnectionType.COM:
-        return new ItcLibraryAdapter();
+        return new ItcLibraryAdapter(onConnect);
       case ConnectionType.Rest:
       default:
-        return new RestLibraryAdapter();
+        return new RestLibraryAdapter(onConnect);
     }
   }
 }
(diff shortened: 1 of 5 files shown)


noahhowell117 (approved):
Code changes look good to me. Tested out locally and the temp table does get removed when reconnecting to a new session.

grantfountain-projs (approved):

Issue #1505, opened 2025-06-02
SAS Notebook: Cleaner Python Output


stu-code:
### Is your feature request related to a problem? Please describe.

When running a Python code block in a SAS Notebook, the notebook will always output about 20 lines of the SAS log even if no print statement is run (except if SAS.show(fig) is used). For example:

![Image](https://github.com/user-attachments/assets/88084f10-21d0-4eeb-979f-d8421f2473b8)

This can create a very crowded notebook with log lines that do not provide any additional information for Python developers. It is hard to read through the notebook when all of these are present.

There is no meaningful output from PROC PYTHON unless there is graphical output, a print statement, an expression, or an error. For example:

**Print statement**

![Image](https://github.com/user-attachments/assets/049a97ab-9b27-4944-b28f-3efe6e86cedf)

**Graphical Output**

![Image](https://github.com/user-attachments/assets/c83c553b-37c0-4aca-9096-73533d01e23b)

**Error**

![Image](https://github.com/user-attachments/assets/cddd9e73-a496-4d39-a629-f2947c43aa2f)

In addition, note that errors still show a green checkmark for the SAS notebook with a Python block. While PROC PYTHON ran successfully, the Python code itself did not.

### Describe the solution you'd like

Cleaner, Jupyter-notebook style output should be used for Python code blocks instead.

1. If there is no error, do not display anything and only display a green checkmark:

![Image](https://github.com/user-attachments/assets/a4e4b34c-a5d1-422e-81e2-b3aeba21226d)

2. If there is a Python code error, display an X, then only display the Python error and do not display the entire SAS log. Instead, the error should show up in the same style as Jupyter notebooks, or at least as close as possible:

![Image](https://github.com/user-attachments/assets/0905679e-38fe-4245-99ff-e5276c036e82)

In other words, the Python error should be extracted to be shown with highlighting rather than only the SAS log, and the cell should indicate an error with an X. If there is an error invoking PROC PYTHON, then the SAS log should appropriately be shown since this is a SAS error rather than a Python error.

3. If the last line of a cell is an expression or a variable name, then the notebook should automatically print it. The user should not need to specify a print statement on a variable or expression to display it. Dataframes (such as pandas or polars) should be printed as tables. Models and pipeline objects should be shown in a graphical format. For example:

**Dataframe**:
![Image](https://github.com/user-attachments/assets/5b5934c4-3c00-4b58-b332-764f94261326)

**List, Dict, Object, Model**:

[...]

PR #2060, opened 2026-09-21
Improve Python notebook cell output

VisheshMalik04:
**Summary:**

- Added `SAS.notebook.outputVisibility` setting.
- Added support for switching between:
  - `clean` (default): removes notebook execution noise from Python output.
  - `raw`: preserves the original notebook execution log output.
- Preserved existing SAS, SQL, and R notebook behavior.
- Refactored notebook output handling to centralize default/raw output generation, making future language-specific output processors easier to introduce.
- Added test coverage for both Python auto-print modes:
  - `(auto)`
  - `SAS.show`

Improvements to Python code blocks based on https://github.com/sassoftware/vscode-sas-extension/issues/1505:

- Suppress SAS log for Python cells; show output only when there is Python stdout, a Python error, or a SAS-level error:

575016514-8710705c-870b-4cb7-b0b7-c439b306e30f
575015981-f364e8af-d38b-4139-b43e-e781368c7c55
575015717-e6462d9f-1334-44cb-8c63-d5a741dd39a0
575015188-cde723de-e03b-4ab7-849d-7980b38726d6

- Render Python exceptions using VS Code's built-in error renderer with coloring and highlighting

575016137-162f7e06-046a-4d7a-9262-958f8aca6398

- Python code that fails to run results in an X on the notebook cell

- New default - Auto-display the last expression in a Python cell: objects with repr_html() (DataFrames, sklearn models, pipelines) are rendered as rich HTML via base64 extraction; all other values fall back to print(). A second optional mode, SAS.show(), renders via ODS HTML5 and is configurable via SAS.notebook.python.autoPrintMethod. Configuration is available in UI.

575016353-b63c04a9-9da4-4316-84de-4063eda59810
575016240-c79d8a0f-72bf-41eb-a650-47d2f3d93419
575015845-f8757839-f929-43d9-86b6-4102b8d4b8ae
575015381-6e805868-b6a0-49b4-b62b-54697fe5e9e1

**Testing:**
Tests were performed manually in a Python code block with the exception of (4) which uses a SAS code block to create a SAS error condition. (See TODOs for updating build tests)

### 1. Tests for if there is no Python stdout or an error:

**Test 1.1:**
```python
import pandas as pd
```

**Expected result**: No output

**Test 1.2:**
```python
foo = 'bar'
```

**Expected result**: No output

**Test 1.3:**
```python
SAS.symput('foo', 'bar')
```

**Expected result**: No output

--------------------------------------------------------------------------------------------------------------------------------

### 2. Tests for Python output:

**Test 2.1:**
```python
40+2
```

**Expected result:**
```
42
```

**Test 2.2:**
```python
{'Hello': 'world'}
```

**Expected result:**
```
{'Hello': 'world'}
```

**Test 2.3:**
```python
asdf
```

**Expected result**:
```
--------------------------------------------------------------------------------------------------------------------------------

NameError                                 Traceback (most recent call last)
  File "<stdin>", line 5, in <module>
  File "<stdin>", line 2, in <module>
  File "<string>", line 1, in <module>
NameError: name 'asdf' is not defined
```

--------------------------------------------------------------------------------------------------------------------------------

### 3. Test DataFrame and model rendering.

**Test 3.1:**
```python
import pandas as pd
df = pd.DataFrame({'foo': [1, 2, 3], 'bar': [4, 5, 6]})
df
```

```python
from sklearn.ensemble import GradientBoostingClassifier
model = GradientBoostingClassifier()
model
```

**Test 3.2:** Switch to SAS.show from (auto) in SAS extension Python autoprint settings:

```python
import pandas as pd
df = pd.DataFrame({'foo': [1, 2, 3], 'bar': [4, 5, 6]})
df
```

```python
from sklearn.ensemble import GradientBoostingClassifier
model = GradientBoostingClassifier()
model
```

**Expected result**: DataFrame renders in ODS HTML5. Model produces the following:

```
This method doesn't know the object passed in. This is an object of type <class 'sklearn.ensemble._gb.GradientBoostingClassifier'>.
GradientBoostingClassifier()
```

--------------------------------------------------------------------------------------------------------------------------------

### 4. Test SAS errors: run the following SAS code, then run a Python code block:

**Test 4.1:**
```sas
options set=PROC_PYPATH="/foo";

proc python restart;
run;
```

```python
'Hello world!'
```

**Expected result**: 

```
ERROR: Failed to create Python subprocess.
ERROR: Launch Failed. Cannot execute program "/foo", error = 2 (No such file or directory).
ERROR: The specified executable module either cannot be located or it is not a valid executable.
ERROR: Failed to launch Python sub-process in tkpy extension.
```

**TODOs:**
- [ ] Update build test cases for new expected Python output behavior
- [ ] Translate new Python autoprint option to other languages


client/src/components/notebook/Controller.ts
```diff
@@ -63,36 +140,100 @@ export class NotebookController {
...
-      execution.end(true, Date.now());
+      const outputItems: vscode.NotebookCellOutputItem[] = [];
+      let success = true;
+
+      if (outputVisibility === "raw") {
+        // Preserve original notebook output behavior for all languages.
+        // Future language-specific output processors (Python, R, SQL, SAS)
+        // should respect this setting and bypass any output cleanup when
+        // raw output is requested.
+        this.appendRawOutput(result, logs, outputItems);
```
Nurquhart, line 167: Looks like this same call to appendRawOutput is called in the final else statement below. It may be a little cleaner if we just do something like:
```
if (PythonCell && outputVisibility !== 'raw') {
    ...
} else {
    this.appendRawOutput(...)
}
```

client/src/components/notebook/Controller.ts
```diff
@@ -63,36 +140,100 @@ export class NotebookController {
...
+          const pyErr = buildPythonError(pythonOutput.errorLines);
+
+          outputItems.push(
+            vscode.NotebookCellOutputItem.error({
+              name: pyErr.name,
+              message: pyErr.message,
+              stack: pyErr.stack,
+            }),
+          );
+        } else if (pythonOutput.isSASError) {
```
Nurquhart, line 195: Any chance of there being both a SAS error and a Python error? Don't know if there is, but we may want to put this isSASError case first just to be safe, that we we default to showing the full output lines and not just the python error (hiding the sas error)

client/src/components/notebook/PythonOutputProcessor.ts
```diff
@@ -0,0 +1,277 @@
...
+ */
+export function processPythonLog(logs: LogLine[]): PythonProcessedOutput {
+  const isPythonError = logs.some(
+    (l) => l.type === "error" && l.line.includes(UNHANDLED_PYTHON_EXCEPTION),
+  );
+
+  // A SAS-level error is any error that is NOT the "Unhandled Python exception"
+  // sentinel – meaning something went wrong before or outside the Python code
+  // itself (e.g. Python not installed, syntax error in PROC PYTHON statement).
+  const isSASError =
```
Nurquhart, line 65: same question here, could there be both sas errors and python errors, and in that case which would we want to show?

also, nitpick, i feel like we can turn this into a single loop instead of doing two different .some loops over the same log ( I know sas logs can get pretty long)

client/src/components/utils/SASCodeDocument.ts
```diff
@@ -194,9 +195,11 @@ ${code}
   }
 
   private wrapPython(code: string) {
+    const method = this.parameters.pythonAutoPrintMethod ?? "(auto)";
+    const processedCode = applyPythonAutoPrint(code, method);
```
Nurquhart, line 199: Two main question here
1. Do we want to do this for just python cells in sas notebooks, or for python programs as well? (I initially thought it was just for sas notebooks)
2. I feel a little weird about wrapping their lines of code with our own code. Is this what python notebooks do? Also, Is this the same thing we were talking about that you asked proc python about? Or is that a different issue? I thought we mentioned to maybe pull this out of the story and turn it into its own standalone issue

client/src/components/notebook/PythonOutputProcessor.ts
VisheshMalik04: Good point. I updated this to use a single pass through the log instead of two separate .some() calls. I also investigated the SAS error vs Python error scenarios. From the cases I tested (NameError, SyntaxError, and PROC PYTHON startup failures using an invalid PROC_PYPATH), the error modes appeared to be mutually exclusive. Python exceptions produce the Unhandled Python exception sentinel and traceback, while PROC PYTHON startup failures produce SAS-level errors such as Failed to create Python subprocess. The current implementation gives precedence to Python exceptions if both conditions were ever encountered.

client/src/components/notebook/Controller.ts
VisheshMalik04: Given that this is a Python notebook cell, I think we should keep Python exceptions taking precedence because the traceback is generally the most actionable output for the user. If we find a scenario where both error types occur in the same execution, we can revisit the precedence logic.

client/src/components/utils/SASCodeDocument.ts
VisheshMalik04: Good questions.
The intent for this is notebook-specific only. The setting is notebook-scoped (`SAS.notebook.python.autoPrintMethod`) and does not affect standalone Python programs.

The user's source code is not modified. Similar to how we already wrap notebook Python code in `PROC PYTHON`, the auto-print logic only transforms the generated execution payload to display the final expression.

Regarding `fig.show()`, I agree that's a separate concern. Only `SAS.show(fig)` is currently supported for rendering graphics. The auto-print changes here are focused on displaying the final expression (`df`, `model`, `40 + 2`, etc.).

client/src/components/utils/SASCodeDocument.ts
```diff
@@ -438,3 +441,151 @@ ${code}`;
...
+  const stripped = line.trim();
+  if (!stripped || stripped.startsWith("#")) {
+    return false;
+  }
+  if (stripped.endsWith(":")) {
+    return false; // block opener
+  }
+
+  for (const prefix of PYTHON_STATEMENT_PREFIXES) {
+    if (stripped === prefix.trim() || stripped.startsWith(prefix)) {
```
ksurajofficial: I see why these are in "PYTHON_STATEMENT_PREFIXES", but could the "startsWith()" check cause issues with variable names that start with these keywords, like "password", "breakdown", "continued", or "elsewhere", and many other names that start with one of the "PYTHON_STATEMENT_PREFIXES"? Since "pass", "break", "continue", "else", etc. are in the list, I was wondering if they could get treated as statements instead of variables. Might be worth checking.

client/src/components/utils/SASCodeDocument.ts
```diff
@@ -438,3 +441,151 @@ ${code}`;
...
+  if (lastIdx < 0) {
+    return code;
+  }
+
+  const lastLine = lines[lastIdx];
+  const indent = lastLine.match(/^(\s*)/)?.[1] ?? "";
+  const trimmedLast = lastLine.trim();
+
+  // Only auto-print top-level (unindented) expressions.
+  if (indent === "" && isPythonExpression(trimmedLast)) {
```
ksurajofficial, line 565: Basic unindented multiline case seems to fail with the fix for me. For example: 

print(
"hello"
)

Can we check? Please ignore if it is just for me.
But if it has started failing, it could affect legitimate cases as well.

Can we also check multiline expressions here? For example:

import pandas as pd

df = pd.DataFrame({
"sas": [1, 2],
"ext": [3, 4]
})

df[
["sas", "ext"]
]

Here "trimmedLast" would just be "]", so I’m wondering if the full expression will still be detected correctly for auto-print.

ksurajofficial: Can you check with other notebooks to be in sync?

client/src/components/utils/SASCodeDocument.ts
```diff
@@ -438,3 +441,151 @@ ${code}`;
...
+    if (method === "(auto)") {
+      // Use _repr_html_() when available (DataFrames, sklearn models, etc.)
+      // for rich HTML output.  Fall back to print() for
+      // plain values that don't implement the protocol.
+      lines[lastIdx] = [
+        `_sas_repr_val_ = (${trimmedLast})`,
```
ksurajofficial, line 571: Very minor issue, and I agree this is a very rare case, but since I noticed it, I thought I’d share. I don’t mind if you decide to leave it as is, but avoiding exposing internal variables like this is generally a good thing.

The auto-print code creates temporary variables like `_sas_repr_val_` (and there are some others as well) in the user's Python session, which could rarely conflict with user-defined variables.

Is it possible to keep these temporary variables local without changing the existing execution behavior, for example by using a separate local namespace for the generated auto-print code?

If you run these 3 lines in 3 different python cells, you will see `_sas_repr_val_` printing 14, even though you initially set it to "user value".

1. `_sas_repr_val_ = "user value"`
2. `12 + 2`
3. `print(_sas_repr_val_)`

Basically, the auto-print logic is overwriting the user-defined `_sas_repr_val_` variable.

client/src/components/utils/SASCodeDocument.ts
ksurajofficial, line 533: I see why these are in "PYTHON_STATEMENT_PREFIXES", but could the "startsWith()" check cause issues with variable names that start with these keywords, like "password", "breakdown", "continued", or "elsewhere", and many other names that start with one of the "PYTHON_STATEMENT_PREFIXES"? Since "pass", "break", "continue", "else", etc. are in the list, I was wondering if they could get treated as statements instead of variables. 

Might be worth checking and instead of using startsWith(), would something like this work? This would check for the whole keyword and avoid matching variable names like "password" for "pass".

function isPythonKeyword(line: string): boolean {
  const stripped = line.trim();

  for (const keyword of PYTHON_KEYWORDS) {
    if (new RegExp(`^${keyword}\\b`).test(stripped)) {
      return true;
    }
  }

  return false;
}

client/src/components/utils/SASCodeDocument.ts
ksurajofficial: Also, looks like semicolon-separated statements aren't handled here. For example, with `x = 10; x` or `import pandas as pd; df = pd.DataFrame({"a": [1, 2]}); df`, the final `x` or `df` doesn't get auto-printed.

client/src/components/utils/SASCodeDocument.ts
```diff
@@ -24,6 +24,7 @@ export interface SASCodeDocumentParameters {
   outputHtml?: boolean;
   htmlStyle?: string;
   uuid?: string;
+  pythonAutoPrintMethod?: string;
```
ksurajofficial, line 27: By looking at the test cases, it seems this can only have - "(auto)" | "SAS.show".

Do you think the following type would make more sense than using a general string? This would also give us compile-time checking if someone accidentally passes another value. If I’m understanding it correctly, these values come from "SAS.notebook.python.autoPrintMethod".

type PythonAutoPrintMethod = "(auto)" | "SAS.show";
pythonAutoPrintMethod?: PythonAutoPrintMethod;

client/src/components/utils/settings.ts
```diff
@@ -37,3 +37,19 @@ export function isShowProblemsFromSASLogEnabled(): boolean {
...
+
+export function getNotebookOutputVisibility(): "clean" | "raw" {
+  return (
+    workspace
+      .getConfiguration("SAS")
+      .get<"clean" | "raw">("notebook.outputVisibility") ?? "clean"
+  );
+}
+
+export function getPythonAutoPrintMethod(): string {
```
ksurajofficial, line 49: Same as above for the return type,

type PythonAutoPrintMethod = "(auto)" | "SAS.show";

client/src/components/notebook/renderers/PythonErrorRenderer.ts
```diff
@@ -0,0 +1,146 @@
+// Copyright © 2026, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import type { ActivationFunction } from "vscode-notebook-renderer";
+
+const COLOR_RED = "#e75c58"; // ansi-red-fg  (separator, error type name)
```
ksurajofficial: Can we use the existing VS Code theme color variables here instead of hard-coded colors? This would keep the Python error renderer consistent with the existing notebook renderers and work better across different VS Code themes.

I see one in LogRenderer.ts

client/src/components/notebook/renderers/PythonErrorRenderer.ts
```diff
@@ -0,0 +1,146 @@
...
+const COLOR_GRAY = "#808080"; // dimmed text  (file refs, "Traceback…" label, ":" + message)
```
ksurajofficial, line 6: Can we use the existing VS Code theme color variables here instead of hard-coded colors? This would keep the Python error renderer consistent with the existing notebook renderers and work better across different VS Code themes.

I see one in LogRenderer.ts

client/src/components/notebook/PythonOutputProcessor.ts
```diff
@@ -0,0 +1,277 @@
+// Copyright © 2026, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { LogLine } from "../../connection";
+
+// VS Code's built-in `application/vnd.code.notebook.error` renderer processes.
+const A_RED = "\u001b[0;31m"; // separator line, error-type name
+const A_GREEN = "\u001b[0;32m"; // file path
+const A_CYAN = "\u001b[0;36m"; // line number, frame/function name
+const A_RESET = "\u001b[0m"; // reset to default
```
ksurajofficial, line 9: TBH, I don’t really like the hardcoded color values here, but I don’t have a specific suggestion for what to use instead. If it’s possible to use colors from the VS Code theme, that would be really good.

client/src/components/utils/SASCodeDocument.ts
Nurquhart: Did a quick test, and the applyPythonAutoPrint seems to run for both notebooks and standalone python files, so we may want to avoid that

client/src/components/notebook/PythonOutputProcessor.ts
VisheshMalik04: These are ANSI escape sequences used to format the traceback passed to NotebookCellOutputItem.error() and consumed by VS Code's built-in error renderer. Unlike PythonErrorRenderer.ts, this path isn't rendering HTML/CSS directly, so theme variables can't be applied here. Unless there's another supported formatting mechanism in the VS Code error renderer, I would prefer to keep the ANSI formatting as-is.

